PR #2, opened 2018-02-22
Fudge openmp from homebrew's GCC 7.2

kornelski:
This should make it noticeably faster. However, it's a bit hacky since it needs to link with GCC's library, which brew puts wherever it wants. Maybe it could be improved with injection of `brew --prefix` somewhere in the config.

### DIFF
--- a/Gifski.xcodeproj/project.pbxproj
+++ b/Gifski.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		5A1FDC6F203F0B050065E0F5 /* libgifski.a in Frameworks */ = {isa = PBXBuildFile; fileRef = E3FD619F201BD29E0087160A /* libgifski.a */; };
+		5A1FDC77203F0C6D0065E0F5 /* libgomp.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 5A1FDC72203F0C6D0065E0F5 /* libgomp.a */; };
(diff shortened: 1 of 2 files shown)


sindresorhus:
@kornelski Can you rebase from master and fix the merge conflict?

sindresorhus:
I tested this PR against release build of master, and it decreases the convert time by ~25%:

```
openmp release build 14.65 seconds
master release build 19.61 seconds
```

sindresorhus:
@kornelski You said we could not use GCD as the frames have to be converted in order, I'm then curious how OpenMP is able to achieve this speed improvement. Do you know?

kornelski:
OpenMP affects internals of libimagequant, e.g. remapping of pixels is done in parallel. It's more fine-grained parallelism than what GCD aims for.

sindresorhus:
Do all frames depend on each other? For example, could we split the a video into two videos, process them in parallel, and then merge them at the end?

sindresorhus:
Could also encode intraframes in parallel.

This is a good read if you haven’t read it already: http://www.drdobbs.com/parallel/optimizing-video-encoding-using-threads/225600370

kornelski:
In GIFs produced by gifski every frame depends on its previous frame, by design. It's equivalent to a video without keyframes. Encoding of frames themselves is already very parallelized.

Splitting input video into parts (roughly equal to number of cores), encoding parts independently and then stitching together would parallelize GIF creation quite well at cost of adding keyframe-like frames in the GIF that may have lower quality (since they won't be able to correctly reuse background from their previous frame).

I've also considered encoding video in fragments, stitching fragments, and then re-encoding frames on the stitches to reduce quality loss. 

However, all of this considerably increases complexity.

PR #6, opened 2018-02-23
Report progress to the system

sindresorhus:
So progress is shown in Finder.

Using the `Progress` type also simplifies the code and paves the way for cancelation.

### DIFF
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -2,30 +2,33 @@ import Foundation
 import AVFoundation
 
 final class Gifski {
-	private var frameCount = 0
-	private var frameIndex = 0
 	private(set) var isRunning = false
-	private(set) var progress: Double = 0
+	private var progress: Progress!
+
+	// `progress.fractionCompleted` is KVO-compliant, but we expose this for convenience
 	var onProgress: ((_ progress: Double) -> Void)?
 
 	/**
 	- parameters:
 		- frameRate: Clamped to 5...30. Uses the frame rate of `inputUrl` if not specified.
 	*/
+	@discardableResult
 	func convertFile(
 		_ inputUrl: URL,
 		outputUrl: URL,
 		quality: Double = 1,
 		dimensions: CGSize? = nil,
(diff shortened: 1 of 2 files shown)


sindresorhus:
@LarsJK Would love a review of this PR and the rest of the code if you have time :)

Gifski/Gifski.swift
```diff
@@ -2,31 +2,34 @@ import Foundation
...
 		guard !isRunning else {
-			return
+			fatalError("Create a new instance if you want to run multiple conversions at once")
 		}
 
-		frameCount = 0
-		frameIndex = 0
 		isRunning = true
 
+		progress = Progress(parent: nil, userInfo: [.fileURLKey: outputUrl])
```
LarsJK: I think it's better to pass `Progress.current()` as parent here, so that it can be observed elsewhere with progress.becomeCurrent...

LarsJK: Any reason you set UserInfo and fileURL?

Gifski/Gifski.swift
```diff
@@ -97,5 +96,7 @@ final class Gifski {
 			gifski_write(g, outputUrl.path)
 			gifski_drop(g)
 		}
+
+		return progress
```
LarsJK: I wouldn't return the progress here. Is it because you might want to use KVO? If so I would use create a new progress, use progress.becomeCurrent.. and observe that instance instead..

Gifski/Gifski.swift
sindresorhus: It's for https://github.com/sindresorhus/gifski-app/pull/9#discussion_r170493129. I'm not sure if both are needed and what the difference is though.

Gifski/Gifski.swift
sindresorhus: > Is it because you might want to use KVO?

Yes

> If so I would use create a new progress, use progress.becomeCurrent.. and observe that instance instead..

I agree. That's better. I just didn't fully realize how NSProgress works.

PR #9, opened 2018-02-25
Add thin class wrapper around the gifski api

LarsJK:
These are the changes I suggested in my review of #6 

I made some changes to the progress handling as well. Let me know if you want me to split up this PR.
Need some more work on the error handling..

Feel like I could improve `setProgressCallback` a bit, but this is what I came up with now.

### DIFF
--- a/Gifski/AppDelegate.swift
+++ b/Gifski/AppDelegate.swift
@@ -251,12 +251,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 		}
 	}
 
-	func updateProgress(_ progress: Double) {
-		circularProgress.progress = CGFloat(progress)
+	func updateProgress(_ progress: Progress) {
+		circularProgress.progress = CGFloat(progress.fractionCompleted)
 
-		if progress == 1 {
-			self.progress?.unpublish()
-
+		if progress.isFinished {
 			circularProgress.percentLabelLayer.string = "✔"
 			circularProgress.fadeOut(delay: 1) {
 				self.isRunning = false
@@ -300,14 +298,16 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 		circularProgress.progress = 0
 		circularProgress.animated = true
 
-		progress = gifski.convertFile(
+        progress = Progress(totalUnitCount: 1)
+        progress?.becomeCurrent(withPendingUnitCount: 1)
+		gifski.convertFile(
 			inputUrl,
 			outputUrl: outputUrl,
 			quality: defaults["outputQuality"] as! Double,
 			dimensions: choosenDimensions,
 			frameRate: choosenFrameRate
 		)
-		progress?.publish()
+        progress?.resignCurrent()
 
 		DockIconProgress.progress = progress
 		DockIconProgress.style = .circle(radius: 55, color: .appTheme)
--- a/Gifski/AppDelegate.swift
+++ b/Gifski/AppDelegate.swift
@@ -298,8 +298,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 		circularProgress.progress = 0
 		circularProgress.animated = true
 
-        progress = Progress(totalUnitCount: 1)
-        progress?.becomeCurrent(withPendingUnitCount: 1)
+		progress = Progress(totalUnitCount: 1)
+		progress?.becomeCurrent(withPendingUnitCount: 1)
 		gifski.convertFile(
 			inputUrl,
 			outputUrl: outputUrl,
@@ -307,7 +307,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 			dimensions: choosenDimensions,
 			frameRate: choosenFrameRate
 		)
-        progress?.resignCurrent()
+		progress?.resignCurrent()
 
 		DockIconProgress.progress = progress
 		DockIconProgress.style = .circle(radius: 55, color: .appTheme)
--- a/Gifski/AppDelegate.swift
+++ b/Gifski/AppDelegate.swift
@@ -125,7 +125,13 @@ extension NSColor {
 @NSApplicationMain
 final class AppDelegate: NSObject, NSApplicationDelegate {
 	@IBOutlet private weak var window: NSWindow!
-	let gifski = Gifski()
+	var gifski: Gifski? {
+		didSet {
+			gifski?.onProgress = { [weak self] progress in
+				self?.updateProgress(progress)
+			}
+		}
+	}
 	var progress: Progress?
 
 	lazy var circularProgress = with(CircularProgressView(frame: CGRect(widthHeight: 160))) {
@@ -186,10 +192,6 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 	func applicationDidFinishLaunching(_ notification: Notification) {
 		hasFinishedLaunching = true
 		NSApplication.shared.isAutomaticCustomizeTouchBarMenuItemEnabled = true
-
-		gifski.onProgress = { progress in
-			self.updateProgress(progress)
-		}
 
 		let view = window.contentView!
 		view.addSubview(circularProgress)
@@ -297,7 +299,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
 		progress = Progress(totalUnitCount: 1)
 		progress?.becomeCurrent(withPendingUnitCount: 1)
-		gifski.convertFile(
+		gifski = Gifski.convertFile(
 			inputUrl,
 			outputUrl: outputUrl,
 			quality: defaults["outputQuality"] as! Double,
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -8,7 +8,7 @@ final class Gifski {
 	// `progress.fractionCompleted` is KVO-compliant, but we expose this for convenience
 	var onProgress: ((_ progress: Progress) -> Void)?
 
-	class func convertFile(
+	static func convertFile(
 		_ inputUrl: URL,
 		outputUrl: URL,
 		quality: Double = 1,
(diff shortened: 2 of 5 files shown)


Gifski/Gifski.swift
```diff
@@ -4,53 +4,54 @@ import AVFoundation
...
 	*/
-	@discardableResult
 	func convertFile(
 		_ inputUrl: URL,
 		outputUrl: URL,
 		quality: Double = 1,
 		dimensions: CGSize? = nil,
 		frameRate: Int? = nil
-	) -> Progress {
+	) {
```
sindresorhus: Per https://github.com/sindresorhus/gifski-app/issues/7#issuecomment-368352717, you can drop the `isRunning` variable if you can think of a better way to check if the conversion is already running.

Gifski/Gifski.swift
```diff
@@ -4,53 +4,54 @@ import AVFoundation
...
 		/// TODO: Find a better way to handle this
 		guard !isRunning else {
 			fatalError("Create a new instance if you want to run multiple conversions at once")
 		}
 
 		isRunning = true
 
-		progress = Progress(parent: nil, userInfo: [.fileURLKey: outputUrl])
-		progress.fileURL = outputUrl
+		progress = Progress(parent: Progress.current(), userInfo: [:])
```
sindresorhus: You don't need to specify the `userInfo: [:]` part.

Gifski/Gifski.swift
```diff
@@ -4,53 +4,54 @@ import AVFoundation
...
 
-		var settings = GifskiSettings(
+        observation = progress.observe(\Progress.fractionCompleted) { (progress, _) in
```
sindresorhus: `\Progress.fractionCompleted` => `\.fractionCompleted`

`(progress, _)` => `progress, _`

Gifski/Gifski.swift
```diff
@@ -4,53 +4,54 @@ import AVFoundation
 final class Gifski {
 	private(set) var isRunning = false
 	private var progress: Progress!
+    private var observation: NSKeyValueObservation?
```
sindresorhus: Incorrect indentation here

Gifski/GifskiWrapper.swift
```diff
@@ -0,0 +1,84 @@
...
+    deinit {
+        gifski_drop(pointer)
+    }
+
+    func setProgressCallback(context: UnsafeMutableRawPointer, cb: @escaping (@convention(c) (UnsafeMutableRawPointer?) -> Int32)) {
+        gifski_set_progress_callback(pointer, cb, context)
+    }
+
+    func addFrameARGB(index: Int, image: CGImage, fps: Double) throws {
+        let buffer = CFDataGetBytePtr(image.dataProvider!.data)
```
sindresorhus: Just a thought, maybe we should not force-unwrap `dataProvider`? I know I'm the one that did it, but in hindsight, I guess we can't be sure it won't be `nil`, right?

Gifski/GifskiWrapper.swift
```diff
@@ -0,0 +1,84 @@
+import Foundation
+
+enum GifskiWrapperError: UInt32, Error {
+    /** one of input arguments was NULL */
+    case nullArg = 1
+    /** a one-time function was called twice, or functions were called in wrong order */
+    case invalidState
+    /** internal error related to palette quantization */
```
sindresorhus: Can we provide the comments here as error messages instead?

Gifski/Gifski.swift
```diff
@@ -73,30 +74,24 @@ final class Gifski {
...
-				)
-				precondition(result == GIFSKI_OK, String(describing: result))
+                    frameIndex += 1
 
-				frameIndex += 1
-
-				if frameIndex == frameForTimes.count {
-					gifski_end_adding_frames(g)
-				}
+                    if frameIndex == frameForTimes.count {
```
sindresorhus: Some incorrect indentation sneaked in here too.

Gifski/AppDelegate.swift
```diff
@@ -300,14 +298,16 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
+        progress = Progress(totalUnitCount: 1)
+        progress?.becomeCurrent(withPendingUnitCount: 1)
+		gifski.convertFile(
 			inputUrl,
 			outputUrl: outputUrl,
 			quality: defaults["outputQuality"] as! Double,
 			dimensions: choosenDimensions,
 			frameRate: choosenFrameRate
 		)
-		progress?.publish()
```
sindresorhus: With these changes, I no longer get the nice progress bar in Finder.

Here, I'm saving to `Downloads` (The progress bar only shows in `master`, not this PR):

screen shot 2018-02-26 at 11 43 20

LarsJK: Oh. I didn't see any effect and the docs don't help at all, but now I know where to look!
Will look into that and fix.. suspect it might have something to do with progress.fileurl

Gifski/Gifski.swift
LarsJK: Looks like these indentation error is caused by my Xcode being set to 4 spaces instead of tabs. Will set to tabs and fix!

Gifski/GifskiWrapper.swift
LarsJK: Yes I guess that would be best. Should probably unwrap at call site and change the method signature to take CFData instead of the image then?

sindresorhus: Yes

Gifski/AppDelegate.swift
LarsJK: I re-added publish/unpublish and the fileURL stuff.
The NSProgress header file comments had some more info regarding this, and the way you did it was correct..

Gifski/Gifski.swift
LarsJK: `[.fileURLKey: outputUrl]` was actually needed for the folder progress bar to work..

LarsJK:
I removed the force unwrapping of CGImage.dataProvider, but it will fatalError if it is nil.
Should probably introduce better error handling in the convert function. Maybe it should take a callback that takes a Result type, but that may be better to discuss in a separate issue/PR?

sindresorhus:
> Maybe it should take a callback that takes a Result type, but that may be better to discuss in a separate issue/PR?

Yeah, I agree. That would be much nicer. I think we should also make a convenience wrapper of `generator.generateCGImagesAsynchronously` in an extension that returns a `Result` when done.

Gifski/Gifski.swift
sindresorhus: @LarsJK Did you think of a way to remove `isRunning`? I guess we could check that `process != nil` and then check `!process.isFinished`.

LarsJK: No I haven't looked into it yet, but that sounds like a good plan. Maybe make isRunning a computed property based on that?

Is there a reason why we can't run 2 conversions at the same time? If it's just a UI thing and not a restriction of this function maybe the isRunning check belongs in AppDelegate instead?

sindresorhus: > Maybe make isRunning a computed property based on that?

👍 

> Is there a reason why we can't run 2 conversions at the same time?

No, it's just that calling the method twice on the same instance will override the current instance variables. I'm open to ideas on how to improve this. I considered moving from `convertFile` to just making it an `init()`, so it's not possible to accidentally call it twice on the same instance, but then it would not be possible to add a method in the future to convert from/to `Data`, for example. Do you know any better way to handle this?

LarsJK: What about making static factory methods that call, a private init and convert. Could that work?

sindresorhus: Yeah. That’s a great idea. Go for it. :)

Gifski/Gifski.swift
```diff
@@ -2,55 +2,61 @@ import Foundation
...
 final class Gifski {
-	private(set) var isRunning = false
 	private var progress: Progress!
+	private var observation: NSKeyValueObservation?
 
 	// `progress.fractionCompleted` is KVO-compliant, but we expose this for convenience
-	var onProgress: ((_ progress: Double) -> Void)?
+	var onProgress: ((_ progress: Progress) -> Void)?
+
+	class func convertFile(
```
sindresorhus: No point in using `class` here as it's not going to be subclassed, so `static` would be better.

Gifski/Gifski.swift
```diff
@@ -2,55 +2,61 @@ import Foundation
...
+		_ inputUrl: URL,
+		outputUrl: URL,
+		quality: Double = 1,
+		dimensions: CGSize? = nil,
+		frameRate: Int? = nil
+		) -> Gifski {
+		let gifski = Gifski()
+		gifski.convertFile(inputUrl, outputUrl: outputUrl, quality: quality, dimensions: dimensions, frameRate: frameRate)
```
sindresorhus: Can you put each parameter on separate lines? For readability.

Gifski/Gifski.swift
LarsJK: Good point!

sindresorhus:
This is just so much better than what we had before.
Thanks for helping out and for your guidance, @LarsJK. 🙌

LarsJK:
Cool! Happy too help 😊

Gifski/AppDelegate.swift
```diff
@@ -125,7 +125,13 @@ extension NSColor {
 @NSApplicationMain
 final class AppDelegate: NSObject, NSApplicationDelegate {
 	@IBOutlet private weak var window: NSWindow!
-	let gifski = Gifski()
+	var gifski: Gifski? {
```
sindresorhus: Does `gifski` really need to be an instance variable now? Couldn't we just initiate it in the `startConversion` function and have the `onProgress` callback there too?

Like:

```swift
…

progress = Progress(totalUnitCount: 1)
progress?.becomeCurrent(withPendingUnitCount: 1)
let gifski = Gifski.convertFile(
	inputUrl,
	outputUrl: outputUrl,
	quality: defaults["outputQuality"] as! Double,
	dimensions: choosenDimensions,
	frameRate: choosenFrameRate
)
gifski.onProgress = { [weak self] progress in
	self?.updateProgress(progress)
}
progress?.resignCurrent()

…
```

LarsJK: True. I dont think progress needs to either..

I kind of want to Gifski.convertFile to just be a function too.. Something like:

`static convert(fileAt:outputTo:withQuality:dimensions:frameRate:completionHandler:)`

I have a proof of concept I can submit to explain better..

PR #16, opened 2018-02-28
Use a static func to convert instead of instance method

LarsJK:
Here is the Proof of concept I mentioned in https://github.com/sindresorhus/gifski-app/pull/9#discussion_r171176503..

I don't like how I have to unpublish in each error case. Could maybe create a wrapper class that publishes in init and unpublished in deinit..

Need to work on the naming, and the error, result type..

### DIFF
--- a/Gifski/AppDelegate.swift
+++ b/Gifski/AppDelegate.swift
@@ -283,16 +283,18 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 		let progress = Progress(totalUnitCount: 1)
 
 		progress.performAsCurrent(withPendingUnitCount: 1) {
-			let conversion = Conversion(
+			let conversion = Gifski.Conversion(
 				input: inputUrl,
 				output: outputUrl,
 				quality: defaults["outputQuality"] as! Double,
 				dimensions: self.choosenDimensions,
 				frameRate: self.choosenFrameRate
 			)
-			Gifski.run(conversion) { result in
+			Gifski.run(conversion) { error in
 				DispatchQueue.main.async {
-					result.forceSuccess()
+					if let error = error {
+						fatalError(error.localizedDescription)
+					}
 					self.circularProgress.percentLabelLayer.string = "✔"
 					self.circularProgress.fadeOut(delay: 1) {
 						self.isRunning = false
--- a/Gifski/AppDelegate.swift
+++ b/Gifski/AppDelegate.swift
@@ -281,30 +281,28 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 		circularProgress.animated = true
 
 		let progress = Progress(totalUnitCount: 1)
-		progress.becomeCurrent(withPendingUnitCount: 1)
-		Gifski.convert(
-			fileAt: inputUrl,
-			outputTo: outputUrl,
-			withQuality: defaults["outputQuality"] as! Double,
-			dimensions: choosenDimensions,
-			frameRate: choosenFrameRate
-		) { [weak self] result in
-			DispatchQueue.main.async {
-				switch result {
-				case .success:
-					self?.circularProgress.percentLabelLayer.string = "✔"
-					self?.circularProgress.fadeOut(delay: 1) {
-						self?.isRunning = false
+
+		progress.performAsCurrent(withPendingUnitCount: 1) {
+			let conversion = Conversion(
+				input: inputUrl,
+				output: outputUrl,
+				quality: defaults["outputQuality"] as! Double,
+				dimensions: self.choosenDimensions,
+				frameRate: self.choosenFrameRate
+			)
+			Gifski.run(conversion) { result in
+				DispatchQueue.main.async {
+					result.forceSuccess()
+					self.circularProgress.percentLabelLayer.string = "✔"
+					self.circularProgress.fadeOut(delay: 1) {
+						self.isRunning = false
 					}
-				case .error(let error):
-					fatalError(error.localizedDescription)
 				}
 			}
 		}
-		progress.resignCurrent()
 
-		progressObserver = progress.observe(\.fractionCompleted) { [weak self] progress, _ in
-			self?.circularProgress.progress = CGFloat(progress.fractionCompleted)
+		progressObserver = progress.observe(\.fractionCompleted) { progress, _ in
+			self.circularProgress.progress = CGFloat(progress.fractionCompleted)
 		}
 
 		DockIconProgress.progress = progress
--- a/Gifski/AppDelegate.swift
+++ b/Gifski/AppDelegate.swift
@@ -125,14 +125,8 @@ extension NSColor {
 @NSApplicationMain
 final class AppDelegate: NSObject, NSApplicationDelegate {
 	@IBOutlet private weak var window: NSWindow!
-	var gifski: Gifski? {
-		didSet {
-			gifski?.onProgress = { [weak self] progress in
-				self?.updateProgress(progress)
-			}
-		}
-	}
-	var progress: Progress?
+
+	private var progressObserver: NSKeyValueObservation?
 
 	lazy var circularProgress = with(CircularProgressView(frame: CGRect(widthHeight: 160))) {
 		$0.foreground = .appTheme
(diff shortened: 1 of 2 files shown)


Gifski/AppDelegate.swift
```diff
@@ -297,16 +280,32 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
+			DispatchQueue.main.async {
+				switch result {
+				case .success:
+					self?.circularProgress.percentLabelLayer.string = "✔"
+					self?.circularProgress.fadeOut(delay: 1) {
+						self?.isRunning = false
+					}
+				case .error(let error):
+					fatalError(error.localizedDescription)
+				}
```
sindresorhus: What do you think about something like the follow:

```swift
extension Result {
  func unwrap() -> T {
    switch self {
    case .success(let value):
      return value
    case .error(let error):
      fatalError(error.localizedDescription)
    }
  }
}
```

I would simplify this code.

Gifski/Gifski.swift
```diff
@@ -1,52 +1,48 @@
...
 		- frameRate: Clamped to 5...30. Uses the frame rate of `inputUrl` if not specified.
 	*/
-	private func convertFile(
-		_ inputUrl: URL,
-		outputUrl: URL,
-		quality: Double = 1,
+	static func convert(
+		fileAt inputUrl: URL,
+		outputTo outputUrl: URL,
+		withQuality quality: Double = 1,
```
sindresorhus: I like the Swift convention of human-like parameters, but isn't it inconsistent that it's `withQuality`, but just `dimensions` and `frameRate`?

LarsJK: Should maybe skip "with"? I think dropping "with" is a better option than adding it to the last two params.

Gifski/AppDelegate.swift
```diff
@@ -297,16 +280,32 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
+		progress.becomeCurrent(withPendingUnitCount: 1)
+		Gifski.convert(
+			fileAt: inputUrl,
+			outputTo: outputUrl,
+			withQuality: defaults["outputQuality"] as! Double,
 			dimensions: choosenDimensions,
 			frameRate: choosenFrameRate
-		)
-		progress?.resignCurrent()
+		) { [weak self] result in
```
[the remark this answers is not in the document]
LarsJK: No I don't think it's needed in this case. I just like to add it just in case I copy paste code into a view controller in the future. I can remove it tho!

kornelski:
Maybe the args could use a builder pattern?

```
let c = Conversion()
c.input =…
c.fps =…
c.quality =…
c.perform(); or gifski.convert(c); 
```

Gifski/AppDelegate.swift
LarsJK: Yeah, look better.
Maybe forceUnwrap() to indicate that it will fatall on error?

This error type has a lot of nice helpers: https://github.com/antitypical/Result

Eventually these errors should probably not fatal. Just show an error message, log and allow the user to try again.

sindresorhus: > Maybe forceUnwrap() to indicate that it will fatall on error?

👍 

> This error type has a lot of nice helpers: antitypical/Result

It's good, but too much for our simple needs, I think.

Gifski/Gifski.swift
sindresorhus: Yes, sounds good.

LarsJK:
What do you think about making the convert function an extension to GifskiWrapper? It could be renamed to just gisfki.

I like what @kornelski suggested to. Maybe something like:

`static Gifski.run(_ conversion: Conversion, completionHandler: (Result) -> Void)`

```swift
let conversion = Conversion(
  input: inputUrl,
  output: outputUrl,
  quality: defaults["outputQuality"] as! Double,
  dimensions: choosenDimensions,
  frameRate: choosenFrameRate
)
Gifski.run(conversion) { result in
  ...
}
```

sindresorhus:
I'm not a big fan of the builder pattern. It feels very legacy Cocoa, but not Swifty. One benefit of the builder pattern though, is that you can conditionally set options, that's harder with arguments, as if they have defaults, it will not be triggered if you set `nil`. An options struct could be nice though.

sindresorhus:
> What do you think about making the convert function an extension to GifskiWrapper? It could be renamed to just gisfki.

What would the benefit of doing that? I like them separate as `Gifski` is what most would use, while `GifskiWrapper` is for advanced usage. We could maybe find a clearer name for `GifskiWrapper`, like `LowlevelGifski`/`BaseGifski`/`GifskiAPI`,etc.

> static Gifski.run(_ conversion: Conversion, completionHandler: (Result) -> Void)

I like this.

Gifski/AppDelegate.swift
```diff
@@ -125,14 +125,8 @@ extension NSColor {
...
-	var gifski: Gifski? {
-		didSet {
-			gifski?.onProgress = { [weak self] progress in
-				self?.updateProgress(progress)
-			}
-		}
-	}
-	var progress: Progress?
+
+	private var progressObserver: NSKeyValueObservation?
```
sindresorhus: Do we really need to keep a reference to it here? I tried having it as just a local variable in `startConversion` and it worked fine.

Gifski/AppDelegate.swift
```diff
@@ -297,16 +280,30 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
-			outputUrl: outputUrl,
-			quality: defaults["outputQuality"] as! Double,
-			dimensions: choosenDimensions,
-			frameRate: choosenFrameRate
-		)
-		progress?.resignCurrent()
+		let progress = Progress(totalUnitCount: 1)
+
+		progress.performAsCurrent(withPendingUnitCount: 1) {
+			let conversion = Conversion(
```
sindresorhus: I think it should be `Gifski.Conversion`, so it's scoped under the `Gifski` name.

sindresorhus:
@LarsJK Would you be interested in joining the project as a maintainer? No worries if not. Just thought I'd ask, as you've done some really great contributions here.

Gifski/Gifski.swift
```diff
@@ -1,78 +1,97 @@
...
+		case .invalidSettings:
+			return "Invalid settings"
+		case .generateFrameFailed:
+			return "Failed to generate frame"
+		case .addFrameFailed(let error):
+			return "Failed to add frame, with underlying error: \(error.localizedDescription)"
+		case .endAddingFramesFailed(let error):
+			return "Failed to end adding frames, with underlying error: \(error.localizedDescription)"
+		case .writeFailed(let error):
+			return "Failed to write to output, with underlying errror: \(error.localizedDescription)"
```
sindresorhus: Typo: `errror`

Gifski/AppDelegate.swift
LarsJK: I just tried it and the main window progress did not animate. Did you try on this branch, or could it be that progress was still an ivar when you tested?

Gifski/AppDelegate.swift
LarsJK: Agree 👍

LarsJK: Should I move GifskiConversionError to Gifski.Error as well?

Gifski/Gifski.swift
```diff
@@ -1,78 +1,97 @@
...
+		self.input = input
+		self.output = output
+		self.quality = quality
+		self.dimensions = dimensions
+		self.frameRate = frameRate
+	}
+}
+
+final class Gifski {
+	static func run(_ conversion: Conversion, completionHandler: ((Result) -> Void)?) {
```
LarsJK: We are not passing anything back in the success case. Do we really need the Result type? Would it be better if the callback took an optional Error?

LarsJK:
Thanks @sindresorhus. I'd love to help maintain this project!

Gifski/AppDelegate.swift
sindresorhus: I tried again and you're right. No idea why it worked earlier...

Gifski/AppDelegate.swift
sindresorhus: > Should I move GifskiConversionError to Gifski.Error as well?

👍

Gifski/Gifski.swift
sindresorhus: Yeah, would probably be better. Although, we should keep cancelation in mind, which might require a Result object with a `cancelled` state.

sindresorhus:
@LarsJK Awesome! Added you to the project :)

Gifski/Gifski.swift
LarsJK: Removed the result type for now. Probably better to add in later when needed..

sindresorhus:
This looks great :)

PR #18, opened 2018-03-05
Animate the drag-n-drop text color

sindresorhus:
`NSTextField#textColor` is not animatable, so I had to be creative. Happy to consider less hacky solutions.

### DIFF
--- a/Gifski/VideoDropView.swift
+++ b/Gifski/VideoDropView.swift
@@ -27,7 +27,7 @@ class DropView: SSView {
 	private var isDraggingHighlighted: Bool = false {
 		didSet {
 			needsDisplay = true
-			dropLabel.animateTextColor(to: isDraggingHighlighted ? highlightColor : .disabledControlTextColor)
+			dropLabel.animateTextColor(to: isDraggingHighlighted ? highlightColor : .disabledControlTextColor, duration: 0.3)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
> An alternative would be to use a CATextLayer instead of Label, and animate the foregroundColor.

I considered that, but according to the internet, `CATextLayer` has worse text handling (kerning and other things) than `NSTextField`. Although, it doesn't matter much here.

LarsJK:
you could probably just create a new label on top instead of generating an image?

sindresorhus:
@LarsJK I would be more than happy to scrap this PR if you wanna try coming up with something better?

LarsJK:
I don't think it makes a difference. I like your approach. Just wanted to mention an alternative:

```swift
extension NSView {

	func createCopy() -> NSView {
		let archivedData = NSKeyedArchiver.archivedData(withRootObject: self)
		return NSKeyedUnarchiver.unarchiveObject(with: archivedData) as! NSView
	}

	/**
	Animate by placing a screenshot of the view above it, changing the properties on the view, and then fading out the screenshot.
	Can be useful for properties that cannot normally be animated.
	*/
	func animateCrossFade(
		duration: TimeInterval = 1,
		delay: TimeInterval = 0,
		animations: @escaping (() -> Void),
		completion: (() -> Void)? = nil
	) {
		let fadeView = createCopy() // <- New
		superview?.addSubview(fadeView, positioned: .above, relativeTo: nil)
		animations()
		fadeView.fadeOut(duration: duration, delay: delay, completion: completion)
	}
}
```

sindresorhus:
> I don't think it makes a difference. I like your approach. Just wanted to mention an alternative:

I actually tried exactly that, but I realize now, I did one major mistake. I added it as a subview to the label instead of its superview. Tried again now and it works. Much better solution.

PR #20, opened 2018-03-07
Move view/model code out of AppDelegate into a WindowController

LarsJK:
I thought the AppDelegate had to many responsibilities, so I moved a lot of the code into a WindowController.

I'm not sure what best practice is with handling menu item actions, but I changed the open menu items action, and enabled state to use the responder chain so that I could keep the logic in the Window controller..

### DIFF
--- a/Gifski/AppDelegate.swift
+++ b/Gifski/AppDelegate.swift
@@ -8,7 +8,7 @@ extension NSColor {
 
 @NSApplicationMain
 final class AppDelegate: NSObject, NSApplicationDelegate {
-	var mainWindowController: MainWindowController!
+	var mainWindowController = MainWindowController()
 
 	var hasFinishedLaunching = false
 	var urlsToConvertOnLaunch: URL!
@@ -22,7 +22,6 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 	}
 
 	func applicationDidFinishLaunching(_ notification: Notification) {
-		mainWindowController = MainWindowController()
 		mainWindowController.showWindow(self)
 
 		hasFinishedLaunching = true
--- a/Gifski/AppDelegate.swift
+++ b/Gifski/AppDelegate.swift
@@ -22,7 +22,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 	}
 
 	func applicationDidFinishLaunching(_ notification: Notification) {
-		mainWindowController = MainWindowController(windowNibName: NSNib.Name(rawValue: "MainWindowController"))
+		mainWindowController = MainWindowController()
 		mainWindowController.showWindow(self)
 
 		hasFinishedLaunching = true
--- a/Gifski/AppDelegate.swift
+++ b/Gifski/AppDelegate.swift
@@ -8,44 +8,10 @@ extension NSColor {
 
 @NSApplicationMain
 final class AppDelegate: NSObject, NSApplicationDelegate {
-	@IBOutlet private weak var window: NSWindow!
-
-	private var progressObserver: NSKeyValueObservation?
-
-	lazy var circularProgress = with(CircularProgressView(frame: CGRect(widthHeight: 160))) {
-		$0.foreground = .appTheme
-		$0.strokeWidth = 2
-		$0.percentLabelLayer.setAutomaticContentsScale()
-		$0.percentLabelLayer.implicitAnimations = false
-		$0.layer?.backgroundColor = .clear
-		$0.isHidden = true
-		$0.centerInWindow(window)
-	}
-
-	lazy var videoDropView = with(VideoDropView()) {
-		$0.frame = window.contentView!.frame
-		$0.dropText = "Drop a Video to Convert to GIF"
-		$0.onComplete = { url in
-			self.convert(url.first!)
-		}
-	}
+	var mainWindowController: MainWindowController!
 
 	var hasFinishedLaunching = false
 	var urlsToConvertOnLaunch: URL!
-	var choosenDimensions: CGSize!
-	var choosenFrameRate: Int!
-
-	@objc dynamic var isRunning: Bool = false {
-		didSet {
-			if isRunning {
-				videoDropView.isHidden = true
-			} else {
-				videoDropView.fadeIn()
-			}
-
-			circularProgress.isHidden = !isRunning
-		}
-	}
 
 	func applicationWillFinishLaunching(_ notification: Notification) {
 		defaults.register(defaults: [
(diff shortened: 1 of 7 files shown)


sindresorhus:
Is there a way to become "first responder" in code instead of having a XIB just for that? I usually try to keep my XIB usage to the minimum.

Gifski/AppDelegate.swift
```diff
@@ -8,83 +8,33 @@ extension NSColor {
...
-			$0.styleMask.remove([.resizable, .fullScreen])
-			$0.styleMask.insert(.fullSizeContentView)
-			$0.isRestorable = false
-			$0.setFrame(CGRect(width: 360, height: 240), display: true)
-			$0.center()
-		}
 	}
 
 	func applicationDidFinishLaunching(_ notification: Notification) {
+		mainWindowController = MainWindowController(windowNibName: NSNib.Name(rawValue: "MainWindowController"))
```
sindresorhus: The recommended practice is to create a `NSNib` extension top-level and define the name there, so you can just use `.mainWindowController` or something here. That's why they introduced strongly typed names for these APIs.

Gifski/Base.lproj/MainMenu.xib
```diff
@@ -68,12 +63,7 @@
                         <items>
                             <menuItem title="Open…" keyEquivalent="o" id="IAo-SY-fd9">
                                 <connections>
-                                    <action selector="openDocument:" target="-1" id="bVn-NM-KNZ"/>
-                                    <binding destination="Voe-Tx-rLC" name="enabled" keyPath="self.isRunning" id="zW7-dU-Zy9">
```
sindresorhus: Here, instead of `self.isRunning`. Is it possible to use `mainWindowController.isRunning`?

Gifski/MainWindowController.swift
```diff
@@ -0,0 +1,167 @@
...
+			} else {
+				videoDropView.fadeIn()
+			}
+
+			circularProgress.isHidden = !isRunning
+		}
+	}
+
+    override func windowDidLoad() {
+        super.windowDidLoad()
```
sindresorhus: Incorrect indentation.

sindresorhus:
Nice cleanup!

Gifski/AppDelegate.swift
LarsJK: 👍 Much better

Gifski/AppDelegate.swift
```diff
@@ -8,83 +8,33 @@ extension NSColor {
...
-	}
-
-	lazy var videoDropView = with(VideoDropView()) {
-		$0.frame = window.contentView!.frame
-		$0.dropText = "Drop a Video to Convert to GIF"
-		$0.onComplete = { url in
-			self.convert(url.first!)
-		}
-	}
+	var mainWindowController: MainWindowController!
```
sindresorhus: Why not just `let mainWindowController = MainWindowController()`?

Gifski/MainWindowController.swift
```diff
@@ -0,0 +1,175 @@
...
+		$0.centerInWindow(window)
+	}
+
+	lazy var videoDropView = with(VideoDropView()) {
+		$0.frame = window!.contentView!.frame
+		$0.dropText = "Drop a Video to Convert to GIF"
+		$0.onComplete = { url in
+			self.convert(url.first!)
+		}
+	}
```
sindresorhus: Can you make these two `private`? I usually prefer to keep everything private unless it needs to be used elsewhere. That way I ensure nothing is used in places where I didn't explicitly intend it to be used.

Gifski/MainWindowController.swift
```diff
@@ -0,0 +1,175 @@
...
+
+	var choosenDimensions: CGSize = .zero
```
sindresorhus: I think this should be an optional instead of `.zero`. You would never want `.zero` dimensions.

Gifski/MainWindowController.swift
sindresorhus: From the docs:

> The default implementation does nothing.

So calling `super` isn't strictly needed. What do you think about it? Should we still do it for consistency?

Gifski/MainWindowController.swift
```diff
@@ -0,0 +1,175 @@
...
+	var choosenFrameRate: Int?
```
sindresorhus: You can also make these `private`.

Gifski/MainWindowController.swift
LarsJK: I think we should remove it 👍

Gifski/MainWindowController.swift
LarsJK: The reason I use .zero is that it is the default value in Gifski.Conversion, otherwise I would have to use ?? .zero there.

Do you prefer if I use `choosenDimensions ?? .zero` in Gifski.Conversion initialiser instead, or use `nil` as the default param in Gifski.Conversion and setup GifskiWrapper like we did before: https://github.com/sindresorhus/gifski-app/blob/v1.0.0/Gifski/Gifski.swift#L33

Gifski/Base.lproj/MainMenu.xib
LarsJK: Not sure. Couldn't find a way to hook this up in IB. Can try some more, maybe editing the XML will work, but that feels a little hacky..

Would you prefer to keep the window in MainMenu.xib, linked to app delegate and do this?
I guess we could do `MainWindowController(window window!)`

We could probably add a custom initialiser that sets up the window programmatically in MainWindowController too.

Gifski/MainWindowController.swift
sindresorhus: > or use nil as the default param in Gifski.Conversion and setup GifskiWrapper like we did before: /Gifski/Gifski.swift@v1.0.0#L33

This. I think the intent is clearer that way.

Gifski/Base.lproj/MainMenu.xib
sindresorhus: I was thinking we just override `init` in the window controller and create the window there. I don't see the point of having a XIB file for this when we're already creating the window programmatically.

Gifski/AppDelegate.swift
```diff
@@ -8,83 +8,32 @@ extension NSColor {
...
-	}
-
-	lazy var videoDropView = with(VideoDropView()) {
-		$0.frame = window.contentView!.frame
-		$0.dropText = "Drop a Video to Convert to GIF"
-		$0.onComplete = { url in
-			self.convert(url.first!)
-		}
-	}
+	var mainWindowController = MainWindowController()
```
sindresorhus: This can be `let`

Gifski/Base.lproj/MainMenu.xib
sindresorhus: I'm fine with using `validateMenuItem` for now. Though I'm surprised there's no good way to bind to values in an `NSWindowController` from `MainMenu.xib`. Seems like an obvious need.

LarsJK: @sindresorhus Will have a look at setting up the window programmatically.

Do you know what style masks I should use?
```swift
let window = NSWindow(
  contentRect: rect,
  styleMask: [.titled, .fullSizeContentView],
  backing: .buffered,
  defer: false
)
```

sindresorhus:
```swift
styleMask: [.titled, .closable, .fullSizeContentView]
```

sindresorhus:
Merging this now so I can get other changes into master without massive merge conflicts. Thanks for your hard work on this, @LarsJK. The code-base is in a much better state now. Can you do a follow-up PR with the change we discussed earlier? No rush though.

LarsJK:
Will do @sindresorhus!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #24, opened 2018-03-13
Use my own progress indicator

sindresorhus:
And some other tweaks.


sindresorhus:
@LarsJK I'm merging this right away as I have other changes I need to do, but I would still appreciate a review if/when you have time :)

PR #30, opened 2018-03-19
Window controller tweaks

sindresorhus:
Just some minor simplifications.

### DIFF
--- a/Gifski/util.swift
+++ b/Gifski/util.swift
@@ -65,6 +65,60 @@ class SSView: NSView {
 
 
 extension NSWindow {
+	// Helper
+	private static func centeredOnScreen(rect: CGRect) -> CGRect {
+		guard let screen = NSScreen.main else {
+			return rect
+		}
+
+		// Looks better than perfectly centered
+		let yOffset = 0.12
+
+		return rect.centered(in: screen.visibleFrame, yOffsetPercent: yOffset)
+	}
+
+	static let defaultContentSize = CGSize(width: 480, height: 300)
+
+	/// TODO: Find a way to stack windows, so additional windows are not placed exactly on top of previous ones: https://github.com/sindresorhus/gifski-app/pull/30#discussion_r175337064
+	static var defaultContentRect: CGRect {
+		return centeredOnScreen(rect: defaultContentSize.cgRect)
+	}
+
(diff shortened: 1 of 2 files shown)


Gifski/util.swift
```diff
@@ -64,6 +64,55 @@ class SSView: NSView {
...
+		let size = defaultContentSize
+
+		guard let screen = NSScreen.main else {
+			return CGRect(origin: .zero, size: size)
+		}
+
+		let extraOffset = screen.visibleFrame.height * 0.19
+
+		return CGRect(
+			x: screen.visibleFrame.midX - (size.width / 2),
```
sindresorhus: @LarsJK Can you see anything wrong with my calculations here? When I remove `window.center()` below the window is not in the exact same `x` position...

Gifski/util.swift
```diff
@@ -64,6 +64,55 @@ class SSView: NSView {
 }
 
 
+extension NSWindow {
+	static let defaultContentSize = CGSize(width: 480, height: 300)
+
+	static var defaultContentRect: CGRect = {
```
sindresorhus: @LarsJK Do you have any idea of how we can programmatically replicate the NIB window behavior of automatically stacking new windows not completely on top of each other?

screen shot 2018-03-19 at 12 57 15

Gifski/util.swift
LarsJK: The calculations look correct to me.. Do you have the dock at the bottom or to either side?

Gifski/util.swift
LarsJK: Not really.. I guess you would have to loop through `NSApplication.shared.windows`, or maybe you just need the keyWindow?
What is the standard cascading behaviour? Will it just add a new window at the same position as the keyWindow + the titles height?

Gifski/util.swift
sindresorhus: No, the Dock is hidden.

sindresorhus: Ok, I solved it. I was stupid. I calculated based on the default window size, not our window size

Gifski/util.swift
sindresorhus: I don't really know. I've added a TODO in the code. I'll look into it at some later time.

Issue #25, opened 2018-03-13
UI ideas


sindresorhus:
Now that the app is almost feature complete, I'd like to explore some UI ideas. The app honestly looks a bit boring.

sindresorhus:
We could give it some retro style, like the icon:

<img width="711" alt="screen shot 2018-03-14 at 01 06 19" src="https://user-images.githubusercontent.com/170270/37360847-f6a217fc-2723-11e8-8797-22189eb417e5.png">

kornelski:
Showing currently encoded frame (effectively playing in slow-mo) is possible.

kornelski:
the ✔️ checkmark at the end is meh. Maybe replace it with a proxy icon for the encoded file, so people can drag it directly from the app, without having to dig it out in Finder?

sindresorhus:
> the ✔️ checkmark at the end is meh. Maybe replace it with a proxy icon for the encoded file, so people can drag it directly from the app, without having to dig it out in Finder?

I agree the checkmark is meh. Not sure whether we should have a proxy icon or just a button to show the file in Finder.

sindresorhus:
> Showing currently encoded frame (effectively playing in slow-mo) is possible.

How do you imagine we would show the actual progress of the conversion with this?

Probably difficult, but would be cool to show the GIF frames in a [flipbook style animation](https://www.youtube.com/watch/?v=OoFUtoyr7i0).

kornelski:
I thought about revealing animation from inside the progress, pie-chart like. OR having faded-out (half bright) animation in the window background, with black circle background for the progress.

flipbook would be cool. A task for https://developer.apple.com/scenekit/ ?

LarsJK:
Could fill the window with the current frame and and have a transparent pie-chart progress animation above. Like iOS app install/updates..

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kornelski:
BTW, Valera has sent an updated icon. How do you like it?
![icon](https://user-images.githubusercontent.com/72159/37418616-884a50d0-27aa-11e8-8c48-ecea3167f2a2.png)

sindresorhus:
It's a good looking logo, but I don't think it's a good fit for a macOS app icon. The white `G` is IMHO too noisy.

briankerr:
First off, cool app! 

[...]

PR #31, opened 2018-03-19
Halve framerate and dimension if they are big

### LINKED ISSUE #28: File size and frame rate defaults
To avoid creating huge/wasteful GIFs, I suggest having smart defaults.

* If resolution is >= 640px wide, halve it. Videos have chroma subsampling and compression artefacts that are wasteful in GIF, so halving resolution for GIF is a good idea anyway.
* If framerate is >= 24fps, halve it. Frames cost a lot in GIF.

LarsJK:
We already clamp the framerate to 5-30. Do wee need to also halve it? If so should the halving be based on the clamped value, or the original value then clamped to 5-30..?

closes #28

### DIFF
--- a/Gifski/SavePanelAccessoryViewController.swift
+++ b/Gifski/SavePanelAccessoryViewController.swift
@@ -1,5 +1,4 @@
 import Cocoa
-import AVFoundation
 
 final class SavePanelAccessoryViewController: NSViewController {
 	@IBOutlet private weak var estimatedSizeLabel: NSTextField!
@@ -12,23 +11,6 @@ final class SavePanelAccessoryViewController: NSViewController {
 	var onDimensionChange: ((CGSize) -> Void)?
 	var onFramerateChange: ((Int) -> Void)?
 
-	var metadata: AVURLAsset.VideoMetadata {
-		return inputUrl.videoMetadata!
-	}
-
-	var frameRate: Int {
-		return Int(metadata.frameRate)
-	}
-
-	var maxFrameRate: Double {
-		return Double(frameRate.clamped(to: 5...30))
-	}
-
-	var defaultFrameRate: Int {
-		let defaultFrameRate = frameRate < 24 ? frameRate : frameRate / 2
-		return defaultFrameRate.clamped(to: 5...30)
-	}
-
 	override func viewDidLoad() {
 		super.viewDidLoad()
 
@@ -39,6 +21,8 @@ final class SavePanelAccessoryViewController: NSViewController {
 		formatter.zeroPadsFractionDigits = true
 
 		/// TODO: Use KVO here
+
+		let metadata = inputUrl.videoMetadata!
 		var currentDimensions = metadata.dimensions
 
 		func estimateFileSize() {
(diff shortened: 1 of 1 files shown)


LarsJK:
Maybe its better if we instead of halving, used 24 fps and 640 width as standard if the video is bigger..?

kornelski:
I'd suggest halve first, then cap. Halving (or dividing by integer in general) gives the best quality -- smoothest animation and nicest scaling.

GIF can only express framerates as 100/n, and most screens can play only 60/n, so 10 and 20fps are the best framerates for GIF. 

Uneven framerate scaling causes judder: https://en.wikipedia.org/wiki/Three-two_pull_down

sindresorhus:
@LarsJK Are you still interested in finishing this? No worries if you don't have time right now.

LarsJK:
@sindresorhus Yeah I can finish this.
Do you have any suggestions for the scale slider? See #28 and @kornelski 's comment on [#25](https://github.com/sindresorhus/gifski-app/issues/25#issuecomment-373951951)

I tried implementing a ticked slider but it it didn't look very good. Too many ticks and probably need to do something about the styling to make it work better with the dark background.

sindresorhus:
My idea is that we use a normal slider, but have a list of predefined sizes, like commented in https://github.com/sindresorhus/gifski-app/issues/25#issuecomment-373951951. So let's say if we had 4 predefined sizes, when the slider is at 0%, it would pick the first, when it's at 20% it would still pick the first, when at 26% it would pick the second, and when at 76% it would pick the last element. So we just spread the elements evenly out. So it's like a tick slider without the ticks.

sindresorhus:
I also think we should revert the open/save dialog to be non-dark. It looked cool at first, but it's clearly not designed to be dark and there are a lot of weird visual bugs.

sindresorhus:
We could also maybe make the text on the right-side of the slider a text field instead, for the perfectionists that want an exact size more easily.

LarsJK:
>I also think we should revert the open/save dialog to be non-dark. It looked cool at first, but it's clearly not designed to be dark and there are a lot of weird visual bugs.

@sindresorhus 🤞Real dark mode support announce next week

sindresorhus:
> 🤞Real dark mode support announce next week

We got it!!!!!!! Note my change in https://github.com/sindresorhus/gifski-app/commit/28948889a8a544c731db9cd29cec98423882da3f

sindresorhus:
> I also think we should revert the open/save dialog to be non-dark. It looked cool at first, but it's clearly not designed to be dark and there are a lot of weird visual bugs.

Done in https://github.com/sindresorhus/gifski-app/commit/28948889a8a544c731db9cd29cec98423882da3f

LarsJK:
@sindresorhus I simplified the code a bit, and changed the minimum scale so that it will allow you to scale the GIF down to 10x10 independent of the input size..

I'm struggling to find a good way to use predefined size and a slider. Do you prefer a slider or could we do a dropdown like preview:

preview

The preset that give the best quality (1/n scaling) can be marked with "best quality" or something like that..

Maybe we should merge this to get better defaults and figure out the way to handle predefines in another PR?

sindresorhus:
> Maybe we should merge this to get better defaults and figure out the way to handle predefines in another PR?

👍 Can you open an issue about the remaining things we need to change? Like slider=>dropdown.

sindresorhus:
This looks great :)

Issue #34, opened 2018-06-12
Build failure


sindresorhus:
I get a build failure when compiling the app now. I think it might be related to installing the latest Xcode 10 beta. Even though I'm compiling from the latest stable Xcode 9.

```
Compiling openmp-sys v0.1.3
error: failed to run custom build command for `openmp-sys v0.1.3`
process didn't exit successfully: `/Users/sindresorhus/Library/Developer/Xcode/DerivedData/Gifski-czyyuqkvdbamwdbhqpunkswgmddo/Build/Products/cargo-target/debug/build/openmp-sys-c94a1c77cd1aa27c/build-script-build` (exit code: 101)
--- stdout
cargo:rerun-if-env-changed=CC
cargo:rerun-if-env-changed=OPENMP_STATIC
cargo:rerun-if-changed=src/ffi.rs
cargo:rerun-if-changed=src/lib.rs
OPT_LEVEL = Some("1")
TARGET = Some("x86_64-apple-darwin")
HOST = Some("x86_64-apple-darwin")
TARGET = Some("x86_64-apple-darwin")
TARGET = Some("x86_64-apple-darwin")
HOST = Some("x86_64-apple-darwin")
CC_x86_64-apple-darwin = None
CC_x86_64_apple_darwin = None
HOST_CC = None
CC = Some("gcc-7")
TARGET = Some("x86_64-apple-darwin")
HOST = Some("x86_64-apple-darwin")
CFLAGS_x86_64-apple-darwin = None
CFLAGS_x86_64_apple_darwin = None
HOST_CFLAGS = None
CFLAGS = None
DEBUG = Some("true")
cargo:flag=-fopenmp

--- stderr
thread 'main' panicked at 'running compiler to get the lib paths: Os { code: 2, kind: NotFound, message: "No such file or directory" }', libcore/result.rs:945:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

```
❯ rustc --version
rustc 1.26.2 (594fb253c 2018-06-01)
```

@kornelski Any ideas?

LarsJK:
Have you tried clearing out derived data and doing a full clean and build?

I dont see the same issues myself.. Are you on 10.12?

sindresorhus:
Yup, tried that. I'm on 10.13.5.

kornelski:
The error happens because Running `cc` or `gcc` can't find such executable.

kornelski:
[...]

PR #37, opened 2018-06-12
Update gifski

kornelski:
#34

### DIFF
--- a/gifski-api/Cargo.lock
+++ b/gifski-api/Cargo.lock
@@ -8,8 +8,11 @@ dependencies = [
 
 [[package]]
 name = "ansi_term"
-version = "0.10.2"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "winapi 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+]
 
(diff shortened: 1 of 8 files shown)


sindresorhus:
It now builds again with this PR, but I'm getting a crash at the end of converting a video to GIF:

```
Thread 1: Fatal error: Failed to write to output, with underlying error: A one-time function was called twice, or functions were called in wrong order
```

kornelski:
I've added better error reporting to the write call

sindresorhus:
That fixed the crash 👌

sindresorhus:
Not important, but I'm getting a warning about:

```
warning: unused manifest key: package.autobins
```

Any way to silence it?

kornelski:
`package.autobins` is needed for forward compatibility with the next version of Rust (one of them will warn in the transition period, ugh).

kornelski:
From what I could find it's because the linker thinks it needs to support C++ exceptions.

Adding "`-Wl,-no_compact_unwind`" flag (somewhere…) should help.

sindresorhus:
> Adding "-Wl,-no_compact_unwind" flag (somewhere…) should help.

Yeah, I found that too, but couldn't figure out where to add it.

PR #39, opened 2018-06-13
Pulsate the drag and drop label

sindresorhus:
gifski-animate

@LarsJK What do you think? Also notice how it scales to the right. Any idea how I can make it scale centered? Maybe setting an anchor point or something.

Oh and I made the window light in https://github.com/sindresorhus/gifski-app/commit/eaaafe2fda6645aec335bfeda0c13ebeda701bec in preparation for macOS 10.14.

### DIFF
--- a/Gifski/VideoDropView.swift
+++ b/Gifski/VideoDropView.swift
@@ -10,8 +10,9 @@ class DropView: SSView {
 	}
 
 	lazy private var dropLabel = with(Label()) {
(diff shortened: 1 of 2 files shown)


sindresorhus:
> Also notice how it scales to the right. Any idea how I can make it scale centered? Maybe setting an anchor point or something.

Solved it in https://github.com/sindresorhus/gifski-app/pull/39/commits/b367536917a6370381a1333cab46007eacd2d898

PR #43, opened 2018-12-26
Add initial `Show in Finder` button

sindresorhus:
This is how it looks:

screen shot 2018-12-27 at 00 31 09

It looks a bit empty, so I thought it might be good to put the filename of the video we generated above the button, and maybe also the final file size. Thoughts?

I created a custom button component. So should be fairly easy to tweak the style. I'm looking for review on the custom button class too.


kornelski:
Yes, a size would be useful. 

An alternative solution could be to show a proxy icon (an icon that and acts like a Finder icon: can be dragged to other apps and double clicked), but I'm not sure how hard that is.

sindresorhus:
I'm merging this now to unblock other work, since I won't be able to finish this with the feedback anytime soon.

PR #45, opened 2019-01-15
Use `Result` type

sindresorhus:
This makes `generateCGImagesAsynchronously` method use a `Result` type. I was not sure whether or not to model the "cancelled" state as success or error.

### DIFF
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -4,7 +4,7 @@ import AVFoundation
 final class Gifski {
 	enum Error: LocalizedError {
 		case invalidSettings
-		case generateFrameFailed
+		case generateFrameFailed(Swift.Error)
 		case addFrameFailed(GifskiWrapperError)
 		case endAddingFramesFailed(GifskiWrapperError)
 		case writeFailed(GifskiWrapperError)
@@ -13,8 +13,8 @@ final class Gifski {
 			switch self {
 			case .invalidSettings:
 				return "Invalid settings"
-			case .generateFrameFailed:
-				return "Failed to generate frame"
+			case .generateFrameFailed(let error):
+				return "Failed to generate frame: \(error.localizedDescription)"
 			case .addFrameFailed(let error):
 				return "Failed to add frame, with underlying error: \(error.localizedDescription)"
 			case .endAddingFramesFailed(let error):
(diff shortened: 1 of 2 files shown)


Gifski/util.swift
```diff
@@ -295,9 +299,63 @@ extension NSAlert {
...
+							image: image,
+							requestedTime: requestedTime,
+							actualTime: actualTime,
+							completedCount: completedCount,
+							totalCount: totalCount,
+							isCancelled: true,
+							isFinished: completedCount == totalCount
+						)
+					)
+				)
```
sindresorhus: I know I could DRY this up, but didn't want to spend too much time on it until I decide on whether to make it a success or failure.

kornelski:
I usually handle cancellations as a type of error, since its handled more like an error path (clean it up) than a success path (keep it).

sindresorhus:
Yeah, I'm starting to question my choice now. I'll try a commit "cancelled" as error and we can compare.

LinusU:
Neat!

I'm personally also leaning towards having cancelation be a failure rather than success. One possible way would be to have the errors conform to a specific protocol that tells that the action was canceled, this is how PromiseKit does it (https://github.com/mxcl/PromiseKit/blob/7f6f86dc8848f153ef3d4f4f037e7a10784bbcaf/Documentation/Cancel.md).

We could potentially make an extension to `Result` to make it easy to detect:

(adapting from https://github.com/mxcl/PromiseKit/blob/0e105ba84217fd53239fedeb3796b08d786de69a/Sources/Error.swift)

```swift
protocol CancellableError: Error {
    /// returns true if this Error represents a cancelled condition
    var isCancelled: Bool { get }
}

extension Result {
    public var isCancelled: Bool {
        do {
            try self.get()
        } catch let error as CancellableError {
            return error.isCancelled
        } catch URLError.cancelled {
            return true
        } catch CocoaError.userCancelled {
            return true
        } catch {
        #if os(macOS) || os(iOS) || os(tvOS)
            let pair = { ($0.domain, $0.code) }(error as NSError)
            return ("SKErrorDomain", 2) == pair
        #else
            return false
        #endif
        }

        return false
    }
}
```

Gifski/util.swift
```diff
@@ -1117,3 +1169,206 @@ extension CGRect {
...
+	}
+	```
+	*/
+    public var isCancelled: Bool {
+        do {
+            _ = try get()
+			return false
+        } catch {
+            return error.isCancelled
+		}
```
LinusU: indentation seems a bit off here (mixed tabs and spaces...)

Gifski/Gifski.swift
```diff
@@ -87,39 +87,45 @@ final class Gifski {
...
+						do {
+							try g.endAddingFrames()
+						} catch {
+							completionHandler?(.endAddingFramesFailed(error as! GifskiWrapperError))
+						}
 					}
-				} catch {
-					completionHandler?(.endAddingFramesFailed(error as! GifskiWrapperError))
-					return
+				case .failure where result.isCancelled:
```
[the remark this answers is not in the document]
LinusU: Potentially something like `case .failure(CancellationError):` but that will only catch specifically `CancellationError`, not any cancelation error 🤔 

Another idea could be to extend `Result` with something that returns a new enum with `.success`, `.cancelation` and `.failure`.

Not sure that would be better though 😁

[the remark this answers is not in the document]
sindresorhus: Hmm, yeah, I don't think that's better than what's already here.

Gifski/util.swift
sindresorhus: Fixed

sindresorhus:
Thanks for the review, @LinusU and @kornelski :)

PR #50, opened 2019-01-17
Add cancel feature

### LINKED ISSUE #5: Add ability to cancel the conversion
The implementation is easy to do. Just return `0` [here](https://github.com/sindresorhus/gifski-app/blob/0cc99c225d3ce20d43582ca6f183f96ea93f2ce3/Gifski/Gifski.swift#L56) when a cancel button is clicked, and change `isRunning` to `false`. We also need to clean up the unfinished file.

I'm not sure yet how to present the button in a nice way in the UI. Suggestions welcome!

boyvanamstel:
Fixes #5

I experimented with a few solutions, before ending up with what I'm submitting now. (Feel free to squash the commit if that's more appropriate.)

The window maintains its minimalist look and feel until you hover over it. The cancel button fades in from the bottom, pushing the circular progress view up a bit. (That should be animating, but it's not. Wrapping the `transform` in a `CATransaction` does make it animate, but resets to its original position after it completes while `autoreverses` etc are not set. Ideas?)

If the mouse exits the window the cancel buttons fades out, restoring the UI.

As an expert feature, the Escape key can also be used to cancel the process.

Upon canceling the conversion the app switches back to its initial state: asking for a file to be dropped on it.

screenshot 2019-01-17 at 23 05 22

(Note: the latest `master` branch introduced an issue with the progress indicator. It seems to start the conversion in the background and takes quite a few moments before it starts indicating progress.)


sindresorhus:
Thanks for working on this! 🙌

sindresorhus:
The dock icon progress is not reset when cancelling.

Gifski/Gifski.swift
```diff
@@ -133,7 +135,12 @@ final class Gifski {
 				try g.write(path: conversion.output.path)
 				completionHandler?(nil)
 			} catch {
-				completionHandler?(.writeFailed(error as! GifskiWrapperError))
+				// TODO: Figure out how to not get a write error when the process was simply cancelled
```
sindresorhus: I'm not seeing any write error.

Gifski/util.swift
```diff
@@ -599,6 +599,27 @@ extension NSView {
 
 		centerInRect(view.bounds)
 	}
+
+	func placeInWindow(_ window: NSWindow? = nil, edge: NSRectEdge, padding: CGFloat = 20) {
```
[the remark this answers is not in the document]
sindresorhus: And make `padding` a `Double`. I prefer using `Double` for my own APIs whenever possible for ease of use.

sindresorhus:
What do you think about this idea for the cancel button? The idea is that it fades over the percentage when you hover the window. It's just a quick mockup. Might look better to have it filled, since the progress circle is already unfilled. The benefit of this is that the progress circle doesn't have to move.

gifski-cancel-button-mockup

sindresorhus:
> (Note: the latest master branch introduced an issue with the progress indicator. It seems to start the conversion in the background and takes quite a few moments before it starts indicating progress.)

I have noticed this too. I'll try to track down when it was introduced.

Gifski/HoverView.swift
```diff
@@ -0,0 +1,37 @@
+import Cocoa
+
+public final class HoverView: NSView {
+	enum Event {
+		case entered, exited
```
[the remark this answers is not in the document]
boyvanamstel: The `HoverView` could be expanded to also trigger `onHover` when the mouse moves inside the window. That's why I went with the `Event` enum straight away.

Gifski/Gifski.swift
```diff
@@ -122,8 +125,7 @@ final class Gifski {
 						}
```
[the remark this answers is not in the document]
boyvanamstel: Added.

Gifski/Gifski.swift
boyvanamstel: screenshot 2019-01-18 at 09 29 42

Even with the cancellation handler on line 83 of `Gifski.swift` implemented, I still trigger this fatal error without the if statement in place.

Gifski/util.swift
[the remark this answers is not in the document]
boyvanamstel: I've added the comment and changed it to take a `Double`.

I prefer floats myself for their memory usage compared to doubles. It also adds a bunch of casts to this method. But I can see that it's more consistent with the existing functions.

boyvanamstel:
> The dock icon progress is not reset when cancelling.

Good catch. See commit message for reasoning behind implementation.

Gifski/Vendor/DockProgress.swift
```diff
@@ -30,6 +30,11 @@ public final class DockProgress {
 		}
 	}
 
+	public static func resetProgress() {
+		progressValue = 0
```
sindresorhus: This should update `previousProgressValue = 0` too.

boyvanamstel:
> What do you think about this idea for the cancel button? The idea is that the `x` button fades over the percentage label when you hover the window. Note that this is just a quick mockup. Might look better to have it filled, since the progress circle is already unfilled. The benefit of this is that the progress circle doesn't have to move. Having the progress circle turn into a cancel button is a common design choice. You can see this in the progress circle button when downloading a song in iTunes, for example.

I think that looks visually quite appealing. Let me see how it translates into code.

boyvanamstel:
circular

I think this looks pretty cool, and it works well. What do you think? :)

boyvanamstel:
screenshot 2019-01-18 at 11 44 21
screenshot 2019-01-18 at 11 44 38

Reduced its size a little.

sindresorhus:
Looks really good :)

The `X` looks a bit weird. It should be totally square. Either using an icon or a different font.

boyvanamstel:
Using a font would make this a lot more portable. I'll have a look.

sindresorhus:
I noticed a problem. After having converted once, I can no longer drag another movie into the window.

boyvanamstel:
> I noticed a problem. After having converted once, I can no longer drag another movie into the window.

🤔 That worked yesterday. I'll look into it.

boyvanamstel:
> And then I thought we could make the progress circle itself the button:

That was my initial thought as well, with the 'growing circle when pressing Esc'. The last mockup looks nice by itself. It's how it transforms between the progress indicator and the cancel state that determines how it finally looks though, I think. It could feel a but too bulky.

Every other element on the screen is outlines only, so it falls out of place pretty quick.

boyvanamstel:
screenshot 2019-01-18 at 12 57 06

I've gone through a few options and this seems to be the only squared x that's positioned properly. 🤔

Reducing the size much makes it rather subtle.

boyvanamstel:
Trying some weird sizes and I came up with this. I rather like it.

screenshot 2019-01-18 at 13 12 18

Especially how pressing cancel basically swallows the progress view.

screenshot 2019-01-18 at 13 12 06

boyvanamstel:
Simply matching the progress view's size, looks cleaner.

screenshot 2019-01-18 at 13 40 21

boyvanamstel:
I can get onboard with this as well. Does not feel as bulky as I expected.

screenshot 2019-01-18 at 13 55 34

boyvanamstel:
dark-filled

sindresorhus:
I love the last one. Just reduce the size of the D slightly. ❤️

boyvanamstel:
@kornelski 💁‍♂️

screenshot 2019-01-18 at 14 32 08

I think this one feels a lot less retro. It's not exactly in the same style as the X, but there's not that many glyphs to choose from unfortunately. Close enough, for now?

boyvanamstel:
Shall we merge this in @sindresorhus? 😊

sindresorhus:
I noticed a problem. After having converted once, I can no longer drag another movie into the window.

Gifski/MainWindowController.swift
```diff
@@ -65,16 +89,23 @@ final class MainWindowController: NSWindowController {
...
+		view?.addSubview(videoDropView, positioned: .above, relativeTo: nil)
 
 		window.makeKeyAndOrderFront(nil)
 		NSApp.activate(ignoringOtherApps: false)
 
 		DockProgress.style = .circle(radius: 55, color: .appTheme)
 	}
 
+	@objc
+	func cancel(_ sender: Any?) {
```
sindresorhus: This could use a comment about where it's coming from. I could also not find this in the docs? Could you?

Gifski/MainWindowController.swift
```diff
@@ -138,23 +173,39 @@ final class MainWindowController: NSWindowController {
...
+		case .entered:
+			circularProgress.isProgressLabelHidden = true
+			cancelButton.fadeIn()
+		case .exited:
+			circularProgress.isProgressLabelHidden = false
+			cancelButton.isHidden = true
+		}
+	}
+
+	private func cancelConversion(_: NSControl?) {
```
sindresorhus: I don't think this should have an argument as it's used in multiple places:

```suggestion
	private func cancelConversion() {
```

Instead, on line 158, use:

```swift
cancelButton.onAction = { _ in
	self.cancelConversion()
}
```

boyvanamstel:
> I noticed a problem. After having converted once, I can no longer drag another movie into the window.

I might've misread this wrong the first time. Do you mean after having cancelled, or should it also accept a new file after successfully completing a conversion?

This commit resolved starting a new conversion after cancelling one first for me: https://github.com/sindresorhus/gifski-app/pull/50/commits/1abb31855a94fb4f8e58a397659c1d6eeed7fac3

Gifski/MainWindowController.swift
```diff
@@ -27,20 +26,45 @@ final class MainWindowController: NSWindowController {
 		$0.centerInWindow(window)
 	}
 
+	private lazy var cancelButton = with(CustomButton.circularButton(title: "╳", size: 130)) {
+		$0.textColor = .appTheme
+		$0.backgroundColor = NSColor.appTheme.with(alpha: 0.1)
```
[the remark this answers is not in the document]
boyvanamstel: That looks absolutely horrendous. The alpha prevents the background color from automatically updating. Should've checked that.

I agree with the `updateLayer` closure being a suitable workaround.

sindresorhus:
> should it also accept a new file after successfully completing a conversion?

Yes

Gifski/MainWindowController.swift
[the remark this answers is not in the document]
sindresorhus: I really wish it was possible to create dynamic colors in user-land... So `NSColor.appTheme.with(alpha: 0.1)` returned a NSColor instance that updated itself, exactly as `NSColor.appTheme` would.

boyvanamstel:
> > should it also accept a new file after successfully completing a conversion?
> 
> Yes

restart

Gifski/MainWindowController.swift
boyvanamstel: I can appreciate the brevity of simply assigning the method to the `onAction` property, but I've committed the change.

Gifski/MainWindowController.swift
boyvanamstel: The only mention of `cancel(_:)` is in `NSSavePanel`, which does not seem immediately related. I don't really like that's it's undocumented.

The alternatives are implementing `cancelOperation(_:)` in a `NSWindow` or `NSView` subclass. The latter will probably require jumping through some extra hoops, accepting first responder etc.

Or falling back to the event listener I implemented first.

Personally, I think there's up and downsides to all of these options. The current implementation seems the most light weight and has my preference.

Gifski/util.swift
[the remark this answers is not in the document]
sindresorhus: The memory usage is negligible in this case.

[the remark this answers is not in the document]
boyvanamstel: Oh sure. Just mentioned it because you said you prefer it.

Gifski/MainWindowController.swift
sindresorhus: It's actually less characters this way. Really depends on how you measure brevity.

Gifski/MainWindowController.swift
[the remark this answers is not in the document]
boyvanamstel: I went with a slightly different solution. Instead of providing an `onUpdateLayer` callback (that could cause some trouble with having to update properties that would in turn call `updateLayer` again), I've added a mechanism that allows you to set a 'color generator' for any of the key paths that return `NSColor`.

```swift
$0.setColorGenerator(for: \.backgroundColor) {
	NSColor.appTheme.with(alpha: 0.1)
}
```

The generator gets called in `animateColor` through a convenience method that returns either the generated color if it exists, or the standard property:

```swift
let backgroundColor = isOn ? color(for: \.activeBackgroundColor) : color(for: \.backgroundColor)
```

Gifski/CustomButton.swift
```diff
@@ -225,6 +225,22 @@ open class CustomButton: NSButton {
 		needsDisplay = true
 	}
 
+	public typealias ColorGenerationHandler = () -> NSColor
+
+	private var colorGenerators = [KeyPath<CustomButton, NSColor>: ColorGenerationHandler]()
+
+	public func setColorGenerator(for keyPath: KeyPath<CustomButton, NSColor>, handler: @escaping ColorGenerationHandler) {
+		colorGenerators[keyPath] = handler
+	}
```
sindresorhus: What do you think about making this setter/getter instead?

Then you could just use it like this:


```swift
// Add
$0.colorGenerator(for: \.backgroundColor) = {
	NSColor.appTheme.with(alpha: 0.1)
}

// Remove
$0.colorGenerator(for: \.backgroundColor) = nil
```

sindresorhus:
Very cool idea to use KeyPaths for this. I would never have thought of that.

sindresorhus:
I also just discovered that it's possible to subclass NSColor: https://pfandrade.me/blog/adopting-dark-mode-and-older-macs/ Although, I have not been able to figure out how to do it in Swift and make it work with `NSColor.white.with(alpha: 0.5)`. Would be super cool to make that work at some point though.

sindresorhus:
And also, please do open issues here if you see anything that could be improved in the app or code, even if you don't intend to fix it yourself.

sindresorhus:
Thank you so much for working on this, @boyvanamstel. It turned out really great! 🙌

boyvanamstel:
> I also just discovered that it's possible to subclass NSColor: https://pfandrade.me/blog/adopting-dark-mode-and-older-macs/ Although, I have not been able to figure out how to do it in Swift and make it work with `NSColor.white.with(alpha: 0.5)`. Would be super cool to make that work at some point though.

Interesting, that might be worth looking into.

sindresorhus:
@boyvanamstel Any thoughts on https://github.com/sindresorhus/gifski-app/pull/50#discussion_r249665639 ?

boyvanamstel:
> @boyvanamstel Any thoughts on [#50 (comment)](https://github.com/sindresorhus/gifski-app/pull/50#discussion_r249665639) ?

Looking into that as we speak.

Gifski/CustomButton.swift
boyvanamstel: I've tried getting close to the syntax you came up with, but I'm not sure it's possible without adding something like `$0.colorGenerator(for: \.backgroundColor).value`. 🤔

I've created a PR that just makes the handler optional, allowing for `$0.setColorGenerator(for: \.backgroundColor, nil)`.

sindresorhus: Yeah, you're right. Too bad that doesn't work though. Would have been nice to have this. I should propose something like that on the Swift Forums.

Could make computed properties support arguments:

```swift
var unicorn(for: Magic): Int {
    get { … }
    set { … }
}
```

Used as:

```swift
.unicorn(for: .rainbow) = 2
```


Thoughts on this?

PR #54, opened 2019-01-22
Tweak custom button color generator setter

boyvanamstel:
See https://github.com/sindresorhus/gifski-app/pull/50#pullrequestreview-194865256

Allows for `$0.setColorGenerator(for: \.backgroundColor, nil)` to be used, making the `removeGenerator(for keyPath:)` method redundant.

### DIFF
--- a/Gifski/CustomButton.swift
+++ b/Gifski/CustomButton.swift
@@ -225,20 +225,20 @@ open class CustomButton: NSButton {
 		needsDisplay = true
 	}
 
-	public typealias ColorGenerationHandler = () -> NSColor
+	public typealias ColorGenerator = () -> NSColor
 
-	private var colorGenerators = [KeyPath<CustomButton, NSColor>: ColorGenerationHandler]()
+	private var colorGenerators = [KeyPath<CustomButton, NSColor>: ColorGenerator]()
 
-	/// Provides a way to re-generate a color on layer update.
-	/// Especially useful for applying alpha values to accent color changes that can be triggered outside of the app.
+	/// Gets or sets the color generation closure for the provided key path.
 	///
-	/// Note: should not be used for returning regular colors.
-	///
-	/// - Parameters:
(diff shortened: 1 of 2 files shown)


sindresorhus:
Alternatively, we could use a subscript:

```swift
$0[colorGenerator: \.backgroundColor] = {
	NSColor.appTheme.with(alpha: 0.1)
}
```

Implemented as:

```swift
subscript (colorGenerator keyPath: KeyPath<CustomButton, NSColor>) -> ColorGenerationHandler? {
	get {
		return colorGenerators[keyPath]
	}
	set {
		colorGenerators[keyPath] = newValue
	}
}
```

I'm not sure about it. Thoughts?

sindresorhus:
What do you think about renaming `ColorGenerationHandler` to `ColorGenerator`?

boyvanamstel:
> What do you think about renaming `ColorGenerationHandler` to `ColorGenerator`?

I usually default to appending `Handler` or `Block` to closures, but I guess `Generator` implies about the same. So I'm for that.

boyvanamstel:
> Alternatively, we could use a subscript:
> 
> ```swift
> $0[colorGenerator: \.backgroundColor] = {
> 	NSColor.appTheme.with(alpha: 0.1)
> }
> ```
> 
> Implemented as:
> 
> ```swift
> subscript (colorGenerator keyPath: KeyPath<CustomButton, NSColor>) -> ColorGenerationHandler? {
> 	get {
> 		return colorGenerators[keyPath]
> 	}
> 	set {
> 		colorGenerators[keyPath] = newValue
> 	}
> }
> ```
> 
> I'm not sure about it. Thoughts?

I considered that, but it seemed a bit unusual and overkill. I do like how it looks.

A downside I see is that if you want to open-source `CustomButton` that 'unusual syntax' might increase the number of questions from users. Nothing documentation can't fix I guess.

I'll add it in and we'll see how it goes.

Issue #46, opened 2019-01-16
Progress sometimes doesn't finish


sindresorhus:
I have encountered this weird problem where sometimes the dock icon progress indicator would stay stuck at 99.9% even when the progress was finished.

<img width="111" alt="screen shot 2019-01-14 at 14 21 01" src="https://user-images.githubusercontent.com/170270/51228589-bc313200-198b-11e9-9440-3fe4c7177c93.png">

I can reproduce it every time with this video file:

[gifski-app-fixture.mov.zip](https://github.com/sindresorhus/gifski-app/files/2762804/gifski-app-fixture.mov.zip)

I added logging here: https://github.com/sindresorhus/gifski-app/blob/4c2e8e4a0d993fe33b3f4a448a44f6e2fc53e05e/Gifski/Gifski.swift#L67

And `print(progress.completedUnitCount, progress.fractionCompleted, progress.isFinished)` correctly reported `176 1.0 true` for the last callback execution.

However, when logging `progress.factionCompletion` in the KVO callback here: https://github.com/sindresorhus/gifski-app/blob/4c2e8e4a0d993fe33b3f4a448a44f6e2fc53e05e/Gifski/Vendor/DockProgress.swift#L18

It outputs `0.9943181818181818` and never resets. So it's clear there's something weird going on with `Progress`...

IssuehuntBot:
@issuehuntfest has funded $200.00 to this issue. [See it on IssueHunt](https://issuehunt.io/repos/119822304/issues/46)

IssuehuntBot:
@boyvanamstel has submitted a pull request. [See it on IssueHunt](https://issuehunt.io/repos/119822304/issues/46)

IssuehuntBot:
@sindresorhus has rewarded $180.00 to @boyvanamstel. [See it on IssueHunt](https://issuehunt.io/repos/119822304/issues/46)

- :moneybag: Total deposit: $200.00
- :tada: Repository reward(0%): $0.00
- :wrench: Service fee(10%): $20.00

boyvanamstel:
Thanks :)

PR #55, opened 2019-01-24
Finish progress based on isFinished instead of fractionCompleted

boyvanamstel:
Addresses the bug outlined in https://github.com/sindresorhus/gifski-app/issues/46.

It sounds like the behavior you described could be caused by some kind of rounding issue.

To rule that out I observe the `.isFinished` property in addition to `.fractionCompleted` to definitively determine when the progress has finished.

Let me know if this resolves the issue in your setup.


sindresorhus:
I created https://github.com/sindresorhus/gifski-app/tree/reproduce-46 to reproduce based on the commit that I discovered the issue, and I'm no longer able to reproduce the incorrect percentage as mentioned in #46. Even though the Dock progress gets stuck, it's still always logged `.fractionCompleted == 1` and `.isFinished == true` in https://github.com/sindresorhus/gifski-app/commit/b196aefc291f248bb7d46253d6d5647c5b5241a9#diff-e88aa751b51e64a91bcf5b68eb33eca6R18.

sindresorhus:
I can also not reproduce it in #56: https://github.com/sindresorhus/gifski-app/pull/56/commits/9e74e02d847afbf41b79188b542f577b11922332

boyvanamstel:
So, the percentage seems fixed, but the dock progress can still get 'stuck'?

boyvanamstel:
Rebased to fix merge conflict.

PR #56, opened 2019-01-25
Fix off-by-one error and ensure Gifski completion handler is only called once

sindresorhus:
I introduced an off-by-one error in 8a764ad695b097cf99356735a9653dd11f0ab1ef. It was only noticeable when trying to convert the video in https://github.com/sindresorhus/gifski-app/issues/46, which is why I didn't discover it when I did the change.

We also had the problem that the `completionHandler` could be called multiple times because we have it in the `generateCGImagesAsynchronously` closure which is called for each finished image or cancellation.

Since there's no good way that I know of to enforce a function to only be called once, I have created my own little clever thing.

### DIFF
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -49,6 +49,8 @@ final class Gifski {
 	}
 
 	static func run(_ conversion: Conversion, completionHandler: ((Error?) -> Void)?) {
+		let completionHandlerOnce = Once().wrap(completionHandler)
+
 		let settings = GifskiSettings(
 			width: UInt32(conversion.dimensions?.width ?? 0),
 			height: UInt32(conversion.dimensions?.height ?? 0),
@@ -58,7 +60,7 @@ final class Gifski {
 		)
 
 		guard let g = GifskiWrapper(settings: settings) else {
-			completionHandler?(.invalidSettings)
+			completionHandlerOnce?(.invalidSettings)
 			return
 		}
 
@@ -94,6 +96,11 @@ final class Gifski {
 			}
 
 			generator.generateCGImagesAsynchronously(forTimePoints: frameForTimes) { result in
+				guard !progress.isCancelled else {
+					completionHandlerOnce?(.cancelled)
+					return
+				}
+
 				switch result {
 				case .success(let result):
 					let image = result.image
@@ -102,13 +109,13 @@ final class Gifski {
 						let data = image.dataProvider?.data,
 						let buffer = CFDataGetBytePtr(data)
 					else {
-						completionHandler?(.generateFrameFailed("Could not get byte pointer of image data provider"))
+						completionHandlerOnce?(.generateFrameFailed("Could not get byte pointer of image data provider"))
 						return
 					}
 
 					do {
 						try g.addFrameARGB(
-							index: UInt32(result.completedCount),
+							index: UInt32(result.completedCount - 1),
 							width: UInt32(image.width),
 							bytesPerRow: UInt32(image.bytesPerRow),
 							height: UInt32(image.height),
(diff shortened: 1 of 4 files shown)


Gifski/Gifski.swift
```diff
@@ -102,48 +109,43 @@ final class Gifski {
...
 				try g.write(path: conversion.output.path)
-				completionHandler?(nil)
+				completionHandlerOnce?(nil)
 			} catch {
-				// TODO: Figure out how to not get a write error when the process was simply cancelled
-				if progress.isCancelled {
-					completionHandler?(.cancelled)
-				} else {
-					completionHandler?(.writeFailed(error as! GifskiWrapperError))
-				}
```
sindresorhus: @boyvanamstel This is fixed now 🎉

Gifski/Gifski.swift
```diff
@@ -102,48 +109,43 @@ final class Gifski {
...
 					else {
-						completionHandler?(.generateFrameFailed("Could not get byte pointer of image data provider"))
+						completionHandlerOnce?(.generateFrameFailed("Could not get byte pointer of image data provider"))
 						return
 					}
 
 					do {
 						try g.addFrameARGB(
-							index: UInt32(result.completedCount),
+							index: UInt32(result.completedCount - 1),
```
sindresorhus: This is the off-by-one error...

Gifski/Gifski.swift
```diff
@@ -94,6 +96,11 @@ final class Gifski {
 			}
 
 			generator.generateCGImagesAsynchronously(forTimePoints: frameForTimes) { result in
+				guard !progress.isCancelled else {
```
sindresorhus: This is needed as Apple is slow at cancelling when we call `generator.cancelAllCGImageGeneration()`, so we end up with `progress.isCancelled == true` and `result.success`, which ends up making it fail.

Gifski/util.swift
```diff
@@ -1373,3 +1373,89 @@ extension Result {
...
+		}
+
+		hasRun = true
+
+		let returnValue = try closure()
+		value = returnValue
+		return returnValue
+	}
+
+	// TODO: Make it support `rethrows` so that if the input `function` is `throws` then the wrapped function becomes throwing too.
```
sindresorhus: @boyvanamstel Any idea how to make this work? Not important, but would definitely be useful to make the `wrap` function work with as many functions as possible.

sindresorhus:
I'm merging this now as I need it for other changes, but I'm still happy to receive review/feedback. Especially on https://github.com/sindresorhus/gifski-app/pull/56#discussion_r251078399.

Gifski/util.swift
LarsJK: @sindresorhus Don't think you can mark a closure as rethrows, but can you just add an overload?

```swift
func wrap<T, U>(_ function: @escaping ((T) throws -> U)) -> ((T) throws -> U) {
  return { parameter in
    try self.run {
      try function(parameter)
    }
  }
}
```

LarsJK:
Sorry for the late review and the lack of contributions/reviews lately.
I've started a new job and am really busy, both at work and home atm.

sindresorhus:
@LarsJK No worries at all, and congrats on the new job 🎉

Gifski/util.swift
sindresorhus: Good idea! I didn't think of that. Still annoying that Swift requires so much duplication. It's now 4 overloads...

LarsJK: Yeah. Haven't tried it yet but with swift 5 and dynamic callable you could maybe make something like:

let onceClosure = Once(closure)
onceClosure("Some Value")

Where the dynamic callable func is rethrowing..

sindresorhus: Hmm, I don’t want to loose type-safety though. I’ll play with it when Swift 5 final is out.

PR #57, opened 2019-01-26
Improve the view after video conversion

### LINKED ISSUE #48: Improve the `Show in Finder` button
An initial version was added in #43, but it could be improved. See the feedback given there.

sentialx:
I've made the draggable proxy icon and also added file name and size info. Here's how it works:

Fixes #48


Gifski/MainWindowController.swift
```diff
@@ -159,7 +159,9 @@ final class MainWindowController: NSWindowController {
 			return
 		}
 
-		showInFinderButton.onAction = { _ in
+		conversionCompletedView.fileUrl = outputUrl
+
+		conversionCompletedView.showInFinderButton.onAction = { _ in
```
sindresorhus: You don't need to set the closure here as `conversionCompletedView` already get's the outputUrl, so you can just handle it inside `conversionCompletedView`.

Gifski/ConversionCompletedView.swift
```diff
@@ -0,0 +1,118 @@
...
+
+	func layoutSubviews() {
+		var height: CGFloat = 0
+
+		for i in (0 ... subviews.count - 1).reversed() {
+			let view = subviews[i]
+			view.frame.origin.y = height + margins[i]
+			height += view.frame.height + margins[i]
+		}
+
```
sindresorhus: Would be better and easier to use auto-layout.

sentialx: I have no idea how to set the view to fit children using auto-layouts :/

sentialx: Done 🎉

sindresorhus:
When I drag out a file in other apps, the temporary image representation turn into the actual file when it's outside the window. Would be nice if it did that here too.

Example (Annotate.app):

link style2

sindresorhus:
The draggable file view should not steal drag state when dragging in a new file after having done a conversion:

link style4

sindresorhus:
We also need the system share button. So people can share their GIF to various services.

sindresorhus:
Don't commit the `.DS_Store` files. You should add it to your global gitignore so they're not accidentally committed.

sindresorhus:
You have not added the `DraggableFile` and `ConversionCompletedView` to the Xcode project.

sindresorhus:
I think you incorrectly rebased https://github.com/sindresorhus/gifski-app/pull/57/commits/9a27726cad2953475b45edb0bb61fa44ca572418 as the latest commit doesn't build.

sindresorhus:
Would be great if quick look worked too. I've seen that in other apps that have the dragging icon like this. So when it's highlighted, and I press <kbd>Space</kbd>, it opens the GIF in quick look.

Gifski/ConversionCompletedView.swift
```diff
@@ -0,0 +1,118 @@
+import Cocoa
+
+class ConversionCompletedView: SSView {
```
sindresorhus: Use `final` for all classes that are not meant to be subclassed.

Gifski/DraggableFile.swift
```diff
@@ -0,0 +1,82 @@
...
+			return
+		}
+
+		guard let image = self.image else {
+			return
+		}
+
+		let size = NSSize(width: log10(image.size.width) * 30, height: log10(image.size.height) * 30)
+
+		if let draggingImage = image.resize(withSize: size) {
```
sindresorhus: Why are you manually resizing the image? I think the dragged image should be the same size as the icon in the conversion view.

sindresorhus:
I only get the correct file icon on the second conversion.

First conversion:

screen shot 2019-01-29 at 22 48 51

Second:

screen shot 2019-01-29 at 22 49 15

Gifski/ConversionCompletedView.swift
```diff
@@ -0,0 +1,118 @@
...
+		do {
+			let attr = try FileManager.default.attributesOfItem(atPath: fileUrl!.path)
+			let dict = attr as NSDictionary
+			size = dict.fileSize()
+		} catch {
+			print("Error: \(error)")
+		}
+
+		fileSize = formatter.string(fromByteCount: Int64(size))
+	}
```
sindresorhus: The logic for getting the pretty file size should be moved into a `URL` extension.

sentialx:
The documentation for the NSFilePromiseProvider is kinda bad. That would be hard to do. To the second comment: maybe I'll just remove the selection as in the Annotation.app there isn't any selection.

sentialx:
Okay, this is how it looks like now:

Gifski/util.swift
```diff
@@ -848,6 +848,23 @@ extension URL {
 	func open() {
 		NSWorkspace.shared.open(self)
 	}
+
+	func getFileSize() -> String {
```
sindresorhus: `func formattedFileSize()`

sindresorhus:
> The documentation for the NSFilePromiseProvider is kinda bad. That would be hard to do.

These are some good resources:
https://buckleyisms.com/blog/how-to-actually-implement-file-dragging-from-your-app-on-mac/
https://www.raywenderlich.com/136272/drag-and-drop-tutorial-for-macos
https://github.com/zhaorui/StickerDrag

sindresorhus:
> Okay, this is how it looks like now:

The icon needs to be larger. Almost double the size. And the file size slightly smaller.

sindresorhus:
> How is the share button supposed to look like?

Not sure. It could be a button like the current "Show in Finder", but only using the system share icon.

.gitignore
```diff
@@ -1,2 +1,3 @@
 xcuserdata
 project.xcworkspace
+.DS_Store
```
sindresorhus: > Don't commit the .DS_Store files. You should add it to your [**global gitignore**](https://www.google.com/search?q=global+gitignore&oq=global+gitignore&aqs=chrome.0.35i39j0l5.2375j0j1&sourceid=chrome&ie=UTF-8) so they're not accidentally committed.

sentialx: I've removed the .DS_Store file and added it to .gitignore

Gifski/ConversionCompletedView.swift
```diff
@@ -0,0 +1,99 @@
+import Cocoa
+
+final class ConversionCompletedView: SSView {
+	var fileUrl: URL? {
+		didSet {
+			draggableFile.fileUrl = fileUrl
+			if let url = fileUrl {
+				draggableFile.image = NSWorkspace.shared.icon(forFile: url.path)
```
sindresorhus: `DraggableFile` should be responsible for getting the icon. Filename too.

sindresorhus: `DraggableFile` should be responsible for getting the icon.

Gifski/DraggableFile.swift
```diff
@@ -0,0 +1,78 @@
...
+
+		if dragDistance < 3 {
+			return
+		}
+
+		guard let image = self.image else {
+			return
+		}
+
+		let draggingItem = NSDraggingItem(pasteboardWriter: NSFilePromiseProvider(fileType: "public.data", delegate: self))
```
sindresorhus: `public.data` => `String(kUTTypeGIF)`

Gifski/util.swift
```diff
@@ -1374,6 +1391,19 @@ extension Result {
 	}
 }
 
+extension NSImage {
+	func resize(withSize targetSize: NSSize) -> NSImage? {
```
sindresorhus: This is no longer used.

sentialx:
Everything works as you wished 🎉 

test

sindresorhus:
I have pushed some cleanup and simplification commits. So make sure you pull before pushing.

sindresorhus:
Can you resolve the merge conflicts?

Gifski/DraggableFile.swift
```diff
@@ -0,0 +1,63 @@
...
+	override func mouseDown(with event: NSEvent) {
+		mouseDownEvent = event
+	}
+
+	override func mouseDragged(with event: NSEvent) {
+		let mouseDownPoint = mouseDownEvent.locationInWindow
+		let dragPoint = event.locationInWindow
+		let dragDistance = hypot(mouseDownPoint.x - dragPoint.x, mouseDownPoint.y - dragPoint.y)
+
+		if dragDistance < 3 {
```
sindresorhus: I don't understand the point of this? I tried removing it and the dragging feels better.

sentialx:
> [#57 (comment)](https://github.com/sindresorhus/gifski-app/pull/57#issuecomment-458592539) is still a problem.

Fixed.

sindresorhus:
Great!

sindresorhus:
What's left:

- [ ] If I start dragging by clicking the top-right part of the move, when I move it, the image suddenly moves to the center of my cursor. It should not do that. It should continue dragging from where I started dragging.
- [ ] Handle longer titles (with auto-layout). https://github.com/sindresorhus/gifski-app/pull/57#issuecomment-458816522
- [ ] Some shadow under the thumbnail. https://github.com/sindresorhus/gifski-app/pull/57#issuecomment-458816345
- [ ] Animate the thumnail dropping into the window when the "conversion completed" view is shown (https://github.com/sindresorhus/gifski-app/issues/48#issuecomment-458642268).
- [ ] Share button. (https://github.com/sindresorhus/gifski-app/issues/48#issuecomment-458554999)

sentialx:
Do I need to wrap longer titles or just show three dots? And I'm not sure if I'd be able to make such an animation.

sentialx:
Everything's done :)

sentialx:
Is anything else needed to be fixed in order to merge this pull request?

Gifski/DraggableFile.swift
```diff
@@ -13,8 +13,17 @@ final class DraggableFile: NSImageView {
...
+		wantsLayer = true
+		
+		let sh = NSShadow()
+		sh.shadowBlurRadius = 5.0
+		sh.shadowOffset = CGSize(width: 0, height: 0)
+		sh.shadowColor = NSColor(red: 0, green: 0, blue: 0, alpha: 0.5)
+		shadow = sh
 	}
-
+
```
sindresorhus: You have introduced a lot of trailing space that should be removed.

Gifski/DraggableFile.swift
```diff
@@ -13,8 +13,17 @@ final class DraggableFile: NSImageView {
 		super.init(frame: frame)
 		isEditable = false
 		unregisterDraggedTypes()
+		
+		superview?.wantsLayer = true
+		wantsLayer = true
+		
+		let sh = NSShadow()
```
sindresorhus: You can use the `with()` function in `util.swift` to simplify the construction of this.

Gifski/DraggableFile.swift
```diff
@@ -13,8 +13,17 @@ final class DraggableFile: NSImageView {
...
+		sh.shadowBlurRadius = 5.0
+		sh.shadowOffset = CGSize(width: 0, height: 0)
+		sh.shadowColor = NSColor(red: 0, green: 0, blue: 0, alpha: 0.5)
```
sindresorhus: The shadow looks good in light mode, but in dark mode it's almost invisible. Make this a named color in asset catalog and make the color in dark mode be `1` instead of `0.5`.

Gifski/DraggableFile.swift
```diff
@@ -34,19 +34,14 @@ final class DraggableFile: NSImageView {
...
 		}
 
 		let size = CGSize(width: 96, height: 96 * (image.size.height / image.size.width))
 
 		let draggingItem = NSDraggingItem(pasteboardWriter: fileUrl as NSURL)
-		let draggingFrameOrigin = convert(mouseDownPoint, from: nil)
-		let draggingImage = image.resizing(to: size)
-		let draggingFrame = CGRect(origin: draggingFrameOrigin, size: draggingImage.size)
-
+		let draggingFrame = CGRect(origin: NSPoint(x: 0, y: (frame.size.height - size.height) / 2), size: size)
```
sindresorhus: `NSPoint` => `CGPoint`

Gifski/DraggableFile.swift
```diff
@@ -13,8 +13,17 @@ final class DraggableFile: NSImageView {
 		super.init(frame: frame)
 		isEditable = false
 		unregisterDraggedTypes()
+		
+		superview?.wantsLayer = true
```
sindresorhus: Subviews should not set `wantsLayer` of its superview. This statement should be in the superview instead.

sindresorhus:
I don't think the auto-layout constraints are correct. This is what I get:

screen shot 2019-02-24 at 13 28 52
screen shot 2019-02-24 at 13 25 32

The buttons are hidden outside the window view.


Here are the test videos I used: [Test videos.zip](https://github.com/sindresorhus/gifski-app/files/2897668/Test.videos.zip)

sindresorhus:
The "Show in Finder" and "Share" buttons should have the same width. Otherwise the flow looks uneven.

screen shot 2019-02-24 at 13 53 59

Gifski.xcodeproj/project.pbxproj
```diff
@@ -488,7 +495,7 @@
 				CODE_SIGN_ENTITLEMENTS = Gifski/Gifski.entitlements;
 				CODE_SIGN_IDENTITY = "Mac Developer";
 				COMBINE_HIDPI_IMAGES = YES;
-				DEVELOPMENT_TEAM = YG56YK5RN5;
+				DEVELOPMENT_TEAM = X3F3ACZNQ7;
```
sindresorhus: You can change this locally, but don't commit this change.

Gifski.xcodeproj/project.pbxproj
```diff
@@ -523,7 +530,7 @@
...
```
sindresorhus: Same as https://github.com/sindresorhus/gifski-app/pull/57/files#r259606220

sindresorhus:
I did some slight tweaks to the animation (https://github.com/sindresorhus/gifski-app/pull/57/commits/bfb3cd41b2a050ecf2b29e8f64be092fc9dcac87). I'm happy with the animation now. Nice work :)

sentialx:
Okay, everything's fixed.

sentialx:
Here's the final look (I think)
image

sindresorhus:
Very good! Almost there. I noticed two tiny issues:

- [ ] When the thumbnail animates in, it now animates from the top-right, instead of from the top, which looks a bit weird.
- [ ] The filename/size labels should be attached (constrained) to the bottom of the thumbnail. It looks good in your above screenshot, but the vertical spacing is too much for non-square GIFs, like this one:

screen shot 2019-03-03 at 18 53 00

I would recommend using `NSStackView` to group some of the views. The filename and size label could be grouped into one `NSStackView`. This filename/size group could then be grouped together with the thumbnail, which is then vertically centered. The button could also be a `NSStackView`.

Gifski/DraggableFile.swift
```diff
@@ -73,10 +73,23 @@ final class DraggableFile: NSImageView {
...
+
+		if width > 96 {
+			width = 96
+			height = 96 * (image.size.height / image.size.width)
+		}
+
+		if height > 96 {
+			height = 96
+			width = 96 * (image.size.width / image.size.height)
+		}
```
sindresorhus: This logic could be simplified with these extensions:

```swift
extension CGSize {
	static func * (lhs: CGSize, rhs: CGFloat) -> CGSize {
		return CGSize(width: lhs.width * rhs, height: lhs.height * rhs)
	}

	func aspectFit(to boundingSize: CGSize) -> CGSize {
		let ratio = min(boundingSize.width / width, boundingSize.height / height)
		return self * ratio
	}
}
```

Gifski/DraggableFile.swift
```diff
@@ -55,17 +55,22 @@ final class DraggableFile: NSImageView {
 	}
 
 	func animateScale() {
-		let springAnimation = CASpringAnimation(keyPath: "transform.scale")
+		let springAnimation = CASpringAnimation(keyPath: #keyPath(CALayer.transform))
+
+		var tr = CATransform3DIdentity
+		tr = CATransform3DTranslate(tr, bounds.size.width / 2, bounds.size.height / 2, 0)
+		tr = CATransform3DScale(tr, 2.0, 2.0, 1)
+		tr = CATransform3DTranslate(tr, -bounds.size.width / 2, -bounds.size.height / 2, 0)
```
sindresorhus: Would be nice to move this logic into a `animateScaleCenter`-kinda method as an extension on `CALayer`, with a `scale` parameter.

sentialx: I've changed it a little bit to animate also the y position. So how could I move this to `animateScaleCenter`?

sindresorhus: Call it `animateScaleMove()` or something and handle both scaling and moving.

sentialx:
Fixed.

sindresorhus:
Did you see this?

> I would recommend using NSStackView to group some of the views. The filename and size label could be grouped into one NSStackView. This filename/size group could then be grouped together with the thumbnail, which is then vertically centered. The button could also be a NSStackView.

sentialx:
Yeah, but how's that different from the actual solution?

sindresorhus:
> Yeah, but how's that different from the actual solution?

It makes the constraint code more readable and more easily maintainable. It can prevent weird constraint issues. And it would let us vertically center the thumbnail/filename/size group in the window.

See: https://www.raywenderlich.com/1206-os-x-stack-views-with-nsstackview

sindresorhus:
When I drag a tall thumbnail, the dragged thumbnail is smaller than before, but still not the same size as the undragged thumbnail.

screen shot 2019-03-03 at 22 04 48

sindresorhus:
The animation looks good now.

sentialx:
> When I drag a tall thumbnail, the dragged thumbnail is smaller than before, but still not the same size as the undragged thumbnail.

What??? It works fine on my Mac. Did you pull the latest commits?

sindresorhus:
Yes, I reproduced it with https://github.com/sindresorhus/gifski-app/pull/57/commits/efa35070e354a7f098c70fd018ce9e688422b21d. You have to first convert a normal rectangle video. Then the tall one, and you'll see it when dragging.

sentialx:
Okay, I've used `NSStackView`.

> Yes, I reproduced it with efa3507. You have to first convert a normal rectangle video. Then the tall one, and you'll see it when dragging.

This doesn't happen to me

sentialx:
kapture 2019-03-03 at 16 38 22

sentialx:
Is everything okay now?

sindresorhus:
> This doesn't happen to me

I can reproduce it every time.

1. Drag [Test - super short.mp4.zip](https://github.com/sindresorhus/gifski-app/files/2923346/Test.-.super.short.mp4.zip) into the app and start the conversion
2. When the completed view is shown, drag [Test - tall - duration short.mov.zip](https://github.com/sindresorhus/gifski-app/files/2923347/Test.-.tall.-.duration.short.mov.zip) into the window and start the conversion.
3. In the completed view, try dragging the thumbnail.

sentialx:
When I convert Test - super short.mp4 it hangs on 0%

sentialx:
Okay, fixed.

sentialx:
Any news?

Gifski/util.swift
```diff
@@ -1030,7 +1030,13 @@ extension CGSize {
 
 	func aspectFit(to boundingSize: CGSize) -> CGSize {
 		let ratio = min(boundingSize.width / width, boundingSize.height / height)
-		return self * ratio
+
+		var size = self * ratio
+
+		size.width = min(self.width, size.width)
+		size.height = min(self.height, size.height)
```
sindresorhus: This is not correct. It doesn't preserve the aspect ratio, which is the main point of this method. It only works correctly since the `boundingSize` has the same width and height.

sentialx: Done

Gifski/util.swift
```diff
@@ -1030,8 +1030,11 @@ extension CGSize {
 
 	func aspectFit(to boundingSize: CGSize) -> CGSize {
 		let ratio = min(boundingSize.width / width, boundingSize.height / height)
+		return self * ratio
+	}
 
-		var size = self * ratio
+	func aspectFit(to scale: CGFloat) -> CGSize {
```
sindresorhus: I don't understand what you're trying to achieve here?

`scale` is not the correct word to use here. It's not a scale you're passing it. It could be called `sides` or `widthHeight`.

Gifski/util.swift
```diff
@@ -1030,8 +1030,11 @@ extension CGSize {
...
+		var size = aspectFit(to: CGSize(width: 96, height: 96))
```
sindresorhus: You're not using the `scale` argument.

Gifski/util.swift
```diff
@@ -1030,8 +1030,11 @@ extension CGSize {
...
 
 		size.width = min(self.width, size.width)
```
sindresorhus: I don't understand what you're trying to do here. By definition, aspect fit fits the image size  inside the provided size, so ensuring the new size is smaller than the existing size is totally moot.

Gifski/util.swift
sindresorhus: If you want to add a new utility method, please add a good description of its purpose to the code too.

sentialx:
Can we merge this finally?

sindresorhus:
Yes, thanks for contributing :)

PR #58, opened 2019-01-28
Update circular progress view

boyvanamstel:
The separate cancel and hover code has been removed and I've copied in the new circular progress view.

There is quite some overlap in the `util.swift` in this project and the one in the circular progress' repo. I've done my best to not add duplicates and uncomment existing commented code, but I think some refactoring might make this an easier process in the future.


sindresorhus:
> There is quite some overlap in the util.swift in this project and the one in the circular progress' repo. I've done my best to not add duplicates and uncomment existing commented code, but I think some refactoring might make this an easier process in the future.

Yeah, it's a mess, but it's only temporarily. We had problems with Swift and frameworks that caused crashes for some users, so I vendored all the code. We can switch back to Carthage again when Swift 5 is out.

Gifski/Vendor/CircularProgress.swift
```diff
@@ -4,7 +4,8 @@ import Cocoa
 @IBDesignable
 public final class CircularProgress: NSView {
 	private var lineWidth: Double = 2
-	private lazy var radius = bounds.width < bounds.height ? bounds.midX * 0.8 : bounds.midY * 0.8
+	// TODO: Remove the closure here when targeting Swift 5
+	private lazy var radius: CGFloat = { bounds.width < bounds.height ? bounds.midX * 0.8 : bounds.midY * 0.8 }()
```
sindresorhus: I got a really weird compile error here so had to make it a closure.

I got this before:

> 1.	While emitting SIL for getter for radius at /Users/sindresorhus/dev/oss/gifski-app/Gifski/Vendor/CircularProgress.swift:8:19
error: Segmentation fault: 11

sindresorhus:
Looks great!

@boyvanamstel Would you be interested in joining the project as a maintainer? No commitment required, just want you to get some credit in the readme. No worries if not though.

Gifski/Vendor/CircularProgress.swift
boyvanamstel: That's odd.

PR #63, opened 2019-03-01
Update CircularProgress

boyvanamstel:
I've moved the utils from CircularProgress into `CircularProgress+Util.swift`, which should make it a little easier to drop in updates.


sindresorhus:
If I start a conversion, cancel it, then start another one, the color is still the disabled style. Something is not being reset correctly.

boyvanamstel:
> If I start a conversion, cancel it, then start another one, the color is still the disabled style. Something is not being reset correctly.

Hmm, I think it’s only resetting the opacity, not the original color. Didn’t catch that in the Example because it sets a new color every time.

PR #65, opened 2019-03-20
Show time remaining estimate during conversion

### LINKED ISSUE #52: Show time left estimate during conversion
It might be useful to show an estimate of the time left during the GIF creation. The conversion speed seems pretty stable, at least from my naive perspective.

@kornelski Thoughts on the best way to calculate it?

---

Note, this is not an easy issue. It's expected that you put a lot of work and perfection into your work.

allewun:
Fixes #52

### DIFF
--- a/Gifski/MainWindowController.swift
+++ b/Gifski/MainWindowController.swift
@@ -142,8 +142,7 @@ final class MainWindowController: NSWindowController {
 	private var progress: Progress?
 
 	private var progressObserver: NSKeyValueObservation?
-	private var indeterminateObserver: NSKeyValueObservation?
-	private var startTime = NSDate()
+	private var startTime = Date()
 
 	func startConversion(inputUrl: URL, outputUrl: URL) {
 		guard !isRunning else {
@@ -153,7 +152,7 @@ final class MainWindowController: NSWindowController {
 		outUrl = outputUrl
 
 		isRunning = true
-		startTime = NSDate()
+		startTime = Date()
 
 		progress = Progress(totalUnitCount: 1)
 		circularProgress.progressInstance = progress
@@ -165,13 +164,13 @@ final class MainWindowController: NSWindowController {
 			let secondsLeft = (secondsElapsed / percent) * (1 - percent)
 
 			DispatchQueue.main.async {
-				self.timeRemainingLabel.text = self.elapsedTimeFormatter.string(from: secondsLeft) ?? "Calculating time remaining…"
-			}
-		}
+				if let remainingText = self.formattedTimeRemaining(seconds: secondsLeft) {
+					if self.timeRemainingLabel.text.isEmpty {
+						self.timeRemainingLabel.fadeIn()
+					}
 
-		indeterminateObserver = progress?.observe(\.isIndeterminate) { _, _ in
-			DispatchQueue.main.async {
-				self.timeRemainingLabel.text = "Calculating time remaining…"
+					self.timeRemainingLabel.text = remainingText
+				}
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
Thanks for working on this 👍

sindresorhus:
Note this from the original issue, which should be handled:

> Yeah, there will be some uneven progress in the beginning (roughly for the first N frames where N = cpu cores), but then it should be smooth.

Gifski/MainWindowController.swift
```diff
@@ -130,6 +141,10 @@ final class MainWindowController: NSWindowController {
 
 	private var progress: Progress?
 
+	private var progressObserver: NSKeyValueObservation?
+	private var indeterminateObserver: NSKeyValueObservation?
+	private var startTime = NSDate()
```
sindresorhus: Why not use `Date()`?

sindresorhus:
I don't think we should show `Calculating time remaining…`. Would be better to just wait to fade in the time remaining label when we have a guesstimate.

sindresorhus:
I think we should only show seconds when less than 1 minutes.

Gifski/MainWindowController.swift
allewun: muscle memory :)

allewun:
> Note this from the original issue, which should be handled:
> 
> > Yeah, there will be some uneven progress in the beginning (roughly for the first N frames where N = cpu cores), but then it should be smooth.

Regarding this, it seems unavoidable that the estimation will be inaccurate at start. Browser file downloads, torrent downloads, video encoding etc.. all tend to behave similarly from what I've seen. We could try waiting a few more seconds before showing the progress so it feels more accurate overall.

sindresorhus:
> We could try waiting a few more seconds before showing the progress so it feels more accurate overall.

Yeah, I think that would be good.

sindresorhus:
And after waiting a few more seconds, if the estimated progress is 10 seconds or less, I don't think we should show the time estimate at all.

sindresorhus:
I think we should fade it out when the time estimate has reached 1 seconds or less, as the below doesn't look very good (It stays there until the progress circle also fades out):

Screen Shot 2019-03-20 at 16 13 12

Gifski/MainWindowController.swift
```diff
@@ -186,6 +217,30 @@ final class MainWindowController: NSWindowController {
...
+		timeRemainingLabel.translatesAutoresizingMaskIntoConstraints = false
+
+		NSLayoutConstraint.activate([
+			timeRemainingLabel.centerXAnchor.constraint(equalTo: view.centerXAnchor),
+			timeRemainingLabel.topAnchor.constraint(equalTo: circularProgress.bottomAnchor)
+		])
+	}
+
+	private func formattedTimeRemaining(seconds: TimeInterval) -> String? {
+		elapsedTimeFormatter.allowedUnits = seconds < 60 ? .second : .minute
```
sindresorhus: It should still show hours.

```suggestion
		elapsedTimeFormatter.allowedUnits = seconds < 60 ? .second : [.hour, .minute]
```

allewun:
@sindresorhus I moved the estimation logic into its own class and implemented:
1. buffer for 3 seconds before showing estimate
2. if initial estimate is less than 10 seconds, don't bother showing it at all
3. fade the estimate 1 second before finishing


**Long processing, show label after estimate stabilizes**:
buffer

**Short processing, don't show estimate**:
skip

sindresorhus:
This is looking great now. Thanks for working on this, @allewun 🙌

allewun:
@sindresorhus thanks for working with me on it!

sindresorhus:
It will be out very soon. It’s waiting for review.

geordiemhall:
Great change! 

Very minor thing - might be nicer to round away from zero for the time shown in the label so that it never hits "0 seconds remaining" (would just remain on "1 second remaining" during fade out).

sindresorhus:
@geordiemhall I agree, and already did that in a follow-up commit: https://github.com/sindresorhus/gifski-app/commit/fc5482ee9c74f26b5281f4c80d848626d8c3dc1e#diff-81f725e4fa6f9c266d8d01bc0c462570R104

geordiemhall:
Ah awesome, didn’t notice sorry!

PR #71, opened 2019-03-28
Ask for App Store review after 5 successful conversions

sindresorhus:
And only ask once! I think this is a good balance of not being intrusive.

### DIFF
--- a/Gifski/Constants.swift
+++ b/Gifski/Constants.swift
@@ -6,4 +6,5 @@ extension NSColor {
 
 extension Defaults.Keys {
 	static let outputQuality = Defaults.Key<Double>("outputQuality", default: 1)
+	static let totalConversionsCount = Defaults.Key<Int>("totalConversionsCount", default: 0)
 }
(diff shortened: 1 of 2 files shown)


Gifski/Constants.swift
```diff
@@ -6,4 +6,5 @@ extension NSColor {
 
 extension Defaults.Keys {
 	static let outputQuality = Defaults.Key<Double>("outputQuality", default: 1)
+	static let totalConversionsCount = Defaults.Key<Int>("totalConversionsCount", default: 0)
```
boyvanamstel: I prefer `successfulConversionsCount` or just `conversionsCount`. I think the key itself implies that it’s the total amount.

sindresorhus:
This worked out well. Already got 2 new reviews:

Screen Shot 2019-04-06 at 05 50 24
Screen Shot 2019-04-06 at 05 50 32

boyvanamstel:
That’s cool. Nice work. :)

sindresorhus:
The reviews keeps pouring in. This was definitely worth doing.

Screen Shot 2019-04-10 at 12 19 25

Nice work, team! 🙌

PR #79, opened 2019-05-07
Add Quick Look shortcut to the conversion completed view

### LINKED ISSUE #66: Press `Space` key to open the GIF in Quick Look
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/gifski-app/issues/66)

Add ability to press `Space` key to open the GIF in Quick Look in the conversion completed view.

There should also be a menu bar item for it called "Quick Look".

When you press Space, the existing thumbnail should animate into the Quick Look window, just like it does in Finder.

---

> There is a $80.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/gifski-app/issues/66).

sunshinejr:
Resolves #66.

Hey @sindresorhus, first of all thanks for this awesome tool. I'm using it on a daily basis and finally I'm able to help out. I'm gonna do more improvements some time in the future as well.

This PR introduces QuickLook addition and fixes one crash (I've added a comment on the line). See some screenshots below and play with the branch yourself!

Also, let me know what documentation would you like me to add in this PR as well.

## QuickLook disabled as there is no GIF yet
Zrzut ekranu 2019-05-7 o 12 33 52

## QuickLook disabled as the GIF is converting still
Zrzut ekranu 2019-05-7 o 12 34 14

## QuickLook enabled as we finished converting
Zrzut ekranu 2019-05-7 o 12 34 37

## QuickLook opened
Zrzut ekranu 2019-05-7 o 12 35 43

### DIFF
--- a/Gifski/Base.lproj/MainMenu.xib
+++ b/Gifski/Base.lproj/MainMenu.xib
@@ -70,7 +70,7 @@
                             <menuItem title="Quick Look" keyEquivalent=" " id="Wfh-bM-KvX">
                                 <modifierMask key="keyEquivalentModifierMask"/>
                                 <connections>
-                                    <action selector="quickLook:" target="-1" id="Zr4-ny-Pim"/>
+                                    <action selector="quickLook:" target="-1" id="IQy-U1-llq"/>
                                 </connections>
                             </menuItem>
                             <menuItem isSeparatorItem="YES" id="m54-Is-iLE"/>
--- a/Gifski/Base.lproj/MainMenu.xib
+++ b/Gifski/Base.lproj/MainMenu.xib
@@ -66,6 +66,12 @@
                                     <action selector="open:" target="-1" id="E6Y-20-4aZ"/>
                                 </connections>
                             </menuItem>
+                            <menuItem title="Quick Look..." keyEquivalent=" " id="Wfh-bM-KvX">
+                                <modifierMask key="keyEquivalentModifierMask"/>
+                                <connections>
+                                    <action selector="quickLook:" target="-1" id="h7S-AO-RZK"/>
+                                </connections>
+                            </menuItem>
                             <menuItem isSeparatorItem="YES" id="m54-Is-iLE"/>
                             <menuItem title="Close" keyEquivalent="w" id="3u8-9s-Y0B">
                                 <connections>
--- a/Gifski/Base.lproj/MainMenu.xib
+++ b/Gifski/Base.lproj/MainMenu.xib
@@ -66,7 +66,8 @@
                                     <action selector="open:" target="-1" id="E6Y-20-4aZ"/>
                                 </connections>
                             </menuItem>
-                            <menuItem title="Quick Look..." keyEquivalent=" " id="Wfh-bM-KvX">
+                            <menuItem isSeparatorItem="YES" id="TB9-2E-bka"/>
+                            <menuItem title="Quick Look" keyEquivalent=" " id="Wfh-bM-KvX">
                                 <modifierMask key="keyEquivalentModifierMask"/>
                                 <connections>
                                     <action selector="quickLook:" target="-1" id="h7S-AO-RZK"/>
--- a/Gifski/Base.lproj/MainMenu.xib
+++ b/Gifski/Base.lproj/MainMenu.xib
@@ -70,7 +70,7 @@
                             <menuItem title="Quick Look" keyEquivalent=" " id="Wfh-bM-KvX">
                                 <modifierMask key="keyEquivalentModifierMask"/>
                                 <connections>
-                                    <action selector="quickLook:" target="-1" id="h7S-AO-RZK"/>
+                                    <action selector="quickLook:" target="-1" id="Zr4-ny-Pim"/>
                                 </connections>
                             </menuItem>
                             <menuItem isSeparatorItem="YES" id="m54-Is-iLE"/>
--- a/Gifski/ConversionCompletedView.swift
+++ b/Gifski/ConversionCompletedView.swift
@@ -29,6 +29,10 @@ final class ConversionCompletedView: SSView {
 		$0.spacing = 20
 	}
 
+	private var isConversionCompleted: Bool {
+		return isHidden == false && fileUrl != nil
+	}
+
 	private func createButton(title: String) -> CustomButton {
 		return with(CustomButton()) {
 			$0.title = title
@@ -68,6 +72,8 @@ final class ConversionCompletedView: SSView {
 	}
 
 	func show() {
+		// We need to manually make self as the first responder, but when the view is hidden it is auto-removed
+		window?.makeFirstResponder(self)
 		fadeIn()
 	}
 
@@ -131,6 +137,10 @@ final class ConversionCompletedView: SSView {
 }
 
 extension ConversionCompletedView: QLPreviewPanelDataSource {
+	@IBAction private func quickLook(_ sender: Any) {
+		quickLookPreviewItems(nil)
+	}
+
 	override func quickLook(with event: NSEvent) {
 		quickLookPreviewItems(nil)
 	}
@@ -173,5 +183,16 @@ extension ConversionCompletedView: QLPreviewPanelDelegate {
 
 	func previewPanel(_ panel: QLPreviewPanel!, transitionImageFor item: QLPreviewItem!, contentRect: UnsafeMutablePointer<CGRect>!) -> Any! {
 		return draggableFile.image
+	}
+}
+
+extension ConversionCompletedView: NSMenuItemValidation {
+	func validateMenuItem(_ menuItem: NSMenuItem) -> Bool {
+		switch menuItem.action {
+		case #selector(quickLook(_:))?:
+			return isConversionCompleted
+		default:
+			return true
+		}
 	}
 }
--- a/Gifski/ConversionCompletedView.swift
+++ b/Gifski/ConversionCompletedView.swift
@@ -74,6 +74,7 @@ final class ConversionCompletedView: SSView {
 	func show() {
 		// We need to manually make self as the first responder, but when the view is hidden it is auto-removed
 		window?.makeFirstResponder(self)
+
 		fadeIn()
 	}
 
--- a/Gifski/ConversionCompletedView.swift
+++ b/Gifski/ConversionCompletedView.swift
@@ -168,14 +168,7 @@ extension ConversionCompletedView: QLPreviewPanelDataSource {
 
 extension ConversionCompletedView: QLPreviewPanelDelegate {
 	func previewPanel(_ panel: QLPreviewPanel!, sourceFrameOnScreenFor item: QLPreviewItem!) -> CGRect {
-		guard let imageRect = draggableFile.subviews.first?.frame, let window = window else {
-			return .zero
-		}
-
-		let windowFrame = draggableFile.convert(imageRect, to: nil)
-		let screenFrame = window.convertToScreen(windowFrame)
-		
-		return screenFrame
+		return draggableFile.imageView?.boundsInScreenCoordinates ?? .zero
 	}
 
 	func previewPanel(_ panel: QLPreviewPanel!, transitionImageFor item: QLPreviewItem!, contentRect: UnsafeMutablePointer<CGRect>!) -> Any! {
--- a/Gifski/ConversionCompletedView.swift
+++ b/Gifski/ConversionCompletedView.swift
@@ -71,18 +71,6 @@ final class ConversionCompletedView: SSView {
 		fadeIn()
 	}
 
-	override func quickLook(with event: NSEvent) {
-		quickLookPreviewItems(nil)
-	}
-
-	override func quickLookPreviewItems(_ sender: Any?) {
-		guard let panel = QLPreviewPanel.shared() else {
-			return
-		}
-
-		panel.toggle()
-	}
-
 	override func didAppear() {
 		translatesAutoresizingMaskIntoConstraints = false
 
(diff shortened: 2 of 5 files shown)


Gifski/MainWindowController.swift
```diff
@@ -320,8 +337,38 @@ extension MainWindowController: NSMenuItemValidation {
 		switch menuItem.action {
 		case #selector(open)?:
 			return !isRunning
+		case #selector(quickLook(_:))?:
+			return conversionCompleted
 		default:
-			return validateMenuItem(menuItem)
```
sunshinejr: this was causing infinity loop as we called the same function we are in, and thus crash

Gifski/MainWindowController.swift
```diff
@@ -61,6 +62,10 @@ final class MainWindowController: NSWindowController {
 		}
 	}
 
+	var conversionCompleted: Bool {
+		return conversionCompletedView.isHidden == false && self.outUrl != nil
```
sunshinejr: ah, just noticed, the `self` is probably not needed here

sunshinejr:
```suggestion
		return conversionCompletedView.isHidden == false && outUrl != nil
```

sunshinejr: Oh nice I can actually make a suggestion to my own PR and commit it without going to the branch and pushing to upstream. So cool!

Gifski/MainWindowController.swift
```diff
@@ -320,8 +337,38 @@ extension MainWindowController: NSMenuItemValidation {
...
 			return !isRunning
+		case #selector(quickLook(_:))?:
+			return conversionCompleted
 		default:
-			return validateMenuItem(menuItem)
+			return true
+		}
+	}
+}
+
```
sindresorhus: Can you also implement `QLPreviewPanelDelegate` so we get the "zoom" effect when pressing Space?

Gifski/MainWindowController.swift
```diff
@@ -61,6 +62,10 @@ final class MainWindowController: NSWindowController {
 		}
 	}
 
+	var conversionCompleted: Bool {
```
sindresorhus:
```suggestion
	var isConversionCompleted: Bool {
```

Gifski/Base.lproj/MainMenu.xib
```diff
@@ -66,6 +66,12 @@
                                     <action selector="open:" target="-1" id="E6Y-20-4aZ"/>
                                 </connections>
                             </menuItem>
+                            <menuItem title="Quick Look..." keyEquivalent=" " id="Wfh-bM-KvX">
```
sindresorhus:
```suggestion
                            <menuItem title="Quick Look" keyEquivalent=" " id="Wfh-bM-KvX">
```

sindresorhus: And I think there should be a separator above this item.

Gifski/MainWindowController.swift
```diff
@@ -301,6 +306,18 @@ final class MainWindowController: NSWindowController {
...
+	@IBAction private func quickLook(_ sender: Any) {
+		guard let panel = QLPreviewPanel.shared() else {
+			return
+		}
+
+		if panel.isVisible {
+			panel.orderOut(nil)
+		} else {
+			panel.makeKeyAndOrderFront(nil)
+		}
```
sindresorhus: The above could be a `QLPreviewPanel#toggle()` extension so the code in here could simply be:

```swift
QLPreviewPanel.shared()?.toggle()
```

Gifski/MainWindowController.swift
```diff
@@ -320,8 +337,38 @@ extension MainWindowController: NSMenuItemValidation {
...
+		case #selector(quickLook(_:))?:
+			return conversionCompleted
 		default:
-			return validateMenuItem(menuItem)
+			return true
+		}
+	}
+}
+
+extension MainWindowController: QLPreviewPanelDataSource {
```
sindresorhus: Wouldn't this logic fit better in ConversionCompletedView.swift?

sindresorhus:
Do you know any apps other than Finder that has a `Quick Look` menu item? I'm curious where they place it.

Gifski/MainWindowController.swift
sindresorhus: Good catch. More info here: https://stackoverflow.com/questions/49388800/falling-through-to-super-validatemenuitem-generates-swift-exception?rq=1

Gifski/Base.lproj/MainMenu.xib
sunshinejr: Yeah, was thinking about that as well. Gonna add it and see.

Gifski/MainWindowController.swift
sunshinejr: Damn, not sure what my brain was doing here, but of course it should be `isConversionCompleted`, good catch.

Gifski/MainWindowController.swift
sunshinejr: Oh nice didn't think of that. Makes sense here. Though, as I learned painfully by experience, we should be prefixing extension methods on Apple-provided classes to avoid unexpected behaviors. Do you have a preference for a prefix, like `gifski_toggle()`? Or do you have a different method for that type of additions?

sunshinejr:
@sindresorhus Hmm, not yet, I was trying to copy what Finder is using (without using CMD + Y for the preview shortcut).

Gifski/MainWindowController.swift
sunshinejr: Alright, will try! Is there somewhere in a code a thumbnail I could use or do I need to create it myself?

Gifski/MainWindowController.swift
sindresorhus: I don't see the need for a prefix. If Apple introduces a `.toggle()` method, although very unlikely, it will need change the semantics of an already compiled binary. It might cause a ambiguity / conflict error at compile-time, but that's ok.

Gifski/MainWindowController.swift
sindresorhus: The thumbnail shown in the conversion completed view. It should zoom animate from that one.

Gifski/MainWindowController.swift
sunshinejr: This is not the only case I was thinking of. I’ve added a property once, `aspectRatio` on PHAsset I believe, and even though there was no method with that name, we’ve got undefined behavior in runtime. App worked most of the time correctly, but there was a case where our method overshadowed an existing, private one and it was doing lots of different, unexpected things with the image.

sunshinejr: Anyhow, I've added an extension to `QLPreviewPanel` without any prefix for now.

Gifski/MainWindowController.swift
sunshinejr: Yeah, I had a problem with that and decided it was fine to leave it in the window controller as it was a general advise I've read. But, I've did some research on why it wasn't working and it turns out that the responder chain got confused and didn't have the conversion view in it. I've managed to hack my way around it (I'm gonna make a comment on the line), but I'm not sure this is the best solution. Was researching other people's implementation on GH and SO but didn't see this type of use-case with main menu + QL preview, which is odd.

Gifski/MainWindowController.swift
sunshinejr: Done 👍

Gifski/MainWindowController.swift
```diff
@@ -301,6 +305,14 @@ final class MainWindowController: NSWindowController {
 		}
 	}
 
+	@IBAction private func quickLook(_ sender: Any) {
+		// Kind of a hack to get our conversion view to be the next responder to answer
```
sunshinejr: @sindresorhus this is the hack I was talking about with moving the logic. Interestingly enough, I had to clear the `nextResponder` as it got into some weird state afterwards.

Gifski/MainWindowController.swift
sindresorhus: I've heard of the problem with Objective-C, but never with Swift.

Gifski/MainWindowController.swift
```diff
@@ -301,6 +305,14 @@ final class MainWindowController: NSWindowController {
 		}
 	}
 
+	@IBAction private func quickLook(_ sender: Any) {
```
sindresorhus: Instead of having this here to merely forward the call. I was thinking we instead only have `func quickLook(_ sender: Any) {` in `ConversionCompletedView` and somehow inject `ConversionCompletedView` into the responder chain so `ConversionCompletedView` can respond directly to the `quickLook` message. That way we wouldn't need https://github.com/sindresorhus/gifski-app/pull/79/files#diff-53ab4064437ff40d094178968ddb4d1fR335 either. (You might have to change the view to be removed/added instead of shown/hidden so it doesn't respond when it's not visible. I'm not sure whether it would respond if it's hidden or if there's anyway to prevent that).

Gifski/ConversionCompletedView.swift
```diff
@@ -128,3 +141,44 @@ final class ConversionCompletedView: SSView {
...
+	func previewPanel(_ panel: QLPreviewPanel!, sourceFrameOnScreenFor item: QLPreviewItem!) -> NSRect {
+		guard let imageRect = draggableFile.subviews.first?.frame, let window = window else {
+			return .zero
+		}
+
+		let windowFrame = draggableFile.convert(imageRect, to: nil)
+		let screenFrame = window.convertToScreen(windowFrame)
+		
+		return screenFrame
+	}
```
sindresorhus: I would use this extension:

```swift
extension NSView {
	/// Get the view frame in screen coordinates.
	var frameInScreenCoordinates: CGRect? {
		return window?.convertToScreen(convert(bounds, to: nil))
	}
}
```

Then the above can simply be:

```swift
func previewPanel(_ panel: QLPreviewPanel!, sourceFrameOnScreenFor item: QLPreviewItem!) -> NSRect {
	return draggableFile.subviews.first?.frameInScreenCoordinates ?? .zero
}
```

Gifski/MainWindowController.swift
sunshinejr: It happened in Swift as well! Makes sense as you can have dynamic/message dispatch in Swift as well.

Gifski/ConversionCompletedView.swift
```diff
@@ -128,3 +141,44 @@ final class ConversionCompletedView: SSView {
...
+
+	func previewPanel(_ panel: QLPreviewPanel!, previewItemAt index: Int) -> QLPreviewItem! {
+		return fileUrl as NSURL
+	}
+}
+
+extension ConversionCompletedView: QLPreviewPanelDelegate {
+	func previewPanel(_ panel: QLPreviewPanel!, sourceFrameOnScreenFor item: QLPreviewItem!) -> NSRect {
+		guard let imageRect = draggableFile.subviews.first?.frame, let window = window else {
```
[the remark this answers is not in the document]
sunshinejr: Good point 👍

Gifski/ConversionCompletedView.swift
sunshinejr: Good point as well 👍

Gifski/MainWindowController.swift
sunshinejr: Yeah, the problem with responder chain is currently beyond me. I think we would need a controller, but I don't have the knowledge needed to make the connection between the window and conversion view in a responder chain. I would use some help here.

sunshinejr: Alright, spent the day on learning how does exactly responder chain work in AppKit and fixed the problem with a proper solution.

sunshinejr:
@sindresorhus fixed the problem with responder chain, ready for another review 👍

Gifski/Base.lproj/MainMenu.xib
```diff
@@ -70,7 +70,7 @@
                             <menuItem title="Quick Look" keyEquivalent=" " id="Wfh-bM-KvX">
                                 <modifierMask key="keyEquivalentModifierMask"/>
                                 <connections>
-                                    <action selector="quickLook:" target="-1" id="Zr4-ny-Pim"/>
+                                    <action selector="quickLook:" target="-1" id="IQy-U1-llq"/>
```
sindresorhus: Just curious, what changed here?

Gifski/MainWindowController.swift
sindresorhus: @sunshinejr Awesome! This is exactly what I conceptually had in mind, but I didn't know how to do it in practice either.

Any recommended reading material?

sindresorhus:
This looks great! @sunshinejr Thanks for taking the time to do this feature. Really nice work 🙌

Gifski/Base.lproj/MainMenu.xib
sunshinejr: Unpinned & pinned the action again and so the id regenerated 🤕

Gifski/MainWindowController.swift
sunshinejr: @sindresorhus I had lots of problems as some of the methods of the responder chain in AppKit are different than the ones in UIKit (e.g. `becomeFirstResponder()` is quite the opposite, lol). But yeah basically reading the docs of responder chain on apple doc archive ([this](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/EventArchitecture/EventArchitecture.html) one for instance). 

I also played with all the properties to make sure I understand the differences (e.g. `initialFirstResponder` vs `firstResponder` etc). 

I have lots of material that I want to convert into a blog post as I didn't found one that goes into what I experienced while implementing this one.

sunshinejr:
thanks @sindresorhus, a pleasure to contribute to such a great project 😉

PR #80, opened 2019-05-07
Add a system service for Gifski

### LINKED ISSUE #47: Add a Gifski system service
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/gifski-app/issues/47)

To quickly convert a movie from anywhere to a GIF.

Docs: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/SysServices/Articles/providing.html

---

---

> There is a $100.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/gifski-app/issues/47).

sunshinejr:
Resolves #47.

How to test it:
1. Archive the app, move it to `/Application` folder
2. In terminal make sure you have the service installed:
  a) `/System/Library/CoreServices/pbs -dump | grep Gifski.app`
  b) If you don’t see `NSBundlePath = "/Applications/Gifski.app”;`, try refreshing using `/System/Library/CoreServices/pbs -update`
3. Go to settings -> keyboard -> shortcuts -> services and enable `Gifski: Convert to GIF`

<details><summary>Sneak peak</summary>

</details>

### DIFF
--- a/Gifski/AppDelegate.swift
+++ b/Gifski/AppDelegate.swift
@@ -24,6 +24,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
 		hasFinishedLaunching = true
 		NSApplication.shared.isAutomaticCustomizeTouchBarMenuItemEnabled = true
+		NSApp.servicesProvider = self
 
 		if urlsToConvertOnLaunch != nil {
 			mainWindowController.convert(urlsToConvertOnLaunch)
@@ -62,5 +63,15 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 	func application(_ application: NSApplication, willPresentError error: Error) -> Error {
 		Crashlytics.recordNonFatalError(error: error)
 		return error
+	}
+
+	/// This is called from NSApp as a service resolver
+	@objc
+	func convertToGif(_ pasteboard: NSPasteboard, userData: String, error: NSErrorPointer) {
+		guard let url = pasteboard.fileURLs().first else {
+			return
+		}
+
+		mainWindowController.convert(url)
 	}
 }
--- a/Gifski/AppDelegate.swift
+++ b/Gifski/AppDelegate.swift
@@ -64,7 +64,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 		Crashlytics.recordNonFatalError(error: error)
 		return error
 	}
+}
 
+extension AppDelegate {
 	/// This is called from NSApp as a service resolver
 	@objc
 	func convertToGif(_ pasteboard: NSPasteboard, userData: String, error: NSErrorPointer) {
--- a/Gifski/Info.plist
+++ b/Gifski/Info.plist
@@ -33,13 +33,15 @@
 			<key>NSMenuItem</key>
 			<dict>
 				<key>default</key>
-				<string>Gifski: Convert to GIF</string>
+				<string>Convert to GIF with Gifski</string>
 			</dict>
 			<key>NSRequiredContext</key>
 			<dict/>
 			<key>NSSendFileTypes</key>
 			<array>
-				<string>public.movie</string>
+				<string>public.mpeg-4</string>
+				<string>com.apple.m4v-video</string>
+				<string>com.apple.quicktime-movie</string> 
 			</array>
 		</dict>
 	</array>
--- a/Gifski/Info.plist
+++ b/Gifski/Info.plist
@@ -23,6 +23,26 @@
 			</array>
 		</dict>
 	</array>
+	<key>NSServices</key>
+	<array>
+		<dict>
+			<key>NSMessage</key>
+			<string>convertToGif</string>
+			<key>NSPortName</key>
+			<string>Gifski</string>
+			<key>NSMenuItem</key>
+			<dict>
+				<key>default</key>
+				<string>Gifski: Convert to GIF</string>
+			</dict>
+			<key>NSRequiredContext</key>
+			<dict/>
+			<key>NSSendFileTypes</key>
+			<array>
+				<string>public.movie</string>
+			</array>
+		</dict>
+	</array>
 	<key>CFBundleExecutable</key>
 	<string>$(EXECUTABLE_NAME)</string>
 	<key>CFBundleIdentifier</key>
--- a/readme.md
+++ b/readme.md
@@ -25,14 +25,7 @@ Requires macOS 10.13 or later.
 
 ## System service
 
-Gifski provides a [system service](https://www.computerworld.com/article/2476298/os-x-a-quick-guide-to-services-on-your-mac.html) that lets you quickly convert a video to GIF from anywhere.
-
-You first have to enable it:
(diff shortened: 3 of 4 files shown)


Gifski/Info.plist
```diff
@@ -23,6 +23,26 @@
...
+	<array>
+		<dict>
+			<key>NSMessage</key>
+			<string>convertToGif</string>
+			<key>NSPortName</key>
+			<string>Gifski</string>
+			<key>NSMenuItem</key>
+			<dict>
+				<key>default</key>
+				<string>Gifski: Convert to GIF</string>
```
sindresorhus: Any thoughts on whether we should use the above text or `Convert to GIF with Gifski`? I have a lot of services, and I see both formats being used.

Gifski/Info.plist
```diff
@@ -23,6 +23,26 @@
...
+			</dict>
+			<key>NSRequiredContext</key>
+			<dict/>
+			<key>NSSendFileTypes</key>
+			<array>
+				<string>public.movie</string>
```
sindresorhus: We don't support all types of movies, so it should be the same as https://github.com/sindresorhus/gifski-app/blob/62877fb387299030d780cbee5df13ebb84eebf36/Gifski/Info.plist#L16-L18

sindresorhus:
Should we also set `NSServiceDescription`? It's not really clear where it would be used. Do you know?

https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/SysServices/Articles/properties.html

sindresorhus:
Can you document the feature in the readme and how to activate and use it? From experience, not everyone knows how Services work. And Apple has not done a very good job of making it obvious how to activate and use them.

sindresorhus:
Could you add a `maintaining.md` file and add the above steps for testing the service?

Gifski/AppDelegate.swift
```diff
@@ -63,4 +64,14 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 		Crashlytics.recordNonFatalError(error: error)
 		return error
 	}
+
+	/// This is called from NSApp as a service resolver
+	@objc
+	func convertToGif(_ pasteboard: NSPasteboard, userData: String, error: NSErrorPointer) {
```
sindresorhus: Can you move this into an extension? Maybe we should also call it `convertToGifService`?

Gifski/Info.plist
sunshinejr: I just used what I've seen the most in the services I use. I don't have a preference as both are used, so yeah just pick whatever you like and can change it accordingly 😄

Gifski/Info.plist
sunshinejr: Good catch 👍

Gifski/AppDelegate.swift
sunshinejr: Sure, will move to extension. The name would be quite confusing for me, as this would suggest we would have some "gif service". But I see where you're coming from, you just want to indicate, in the function name, that this is called from a service right? I think we could leave the name as is with the comment and extension for now, but again up to you.

sindresorhus: Alright, we don't have to change the name. It just bothers me in general how implicit all these old Cocoa APIs are.

Gifski/Info.plist
sindresorhus: I think having the app name as prefix is useful if an app exposes multiple services, but when there's only one service, `Convert to GIF with Gifski` reads better, so let's go with that.

sunshinejr: Good point, gonna update the name as well 👍

sunshinejr:
@sindresorhus updated the PR with code changes and initial drafts of readme/maintaining docs. Let me know how it looks now!

Also, I didn't use `NSServiceDescription` and also don't quite understand where would it be seen. When I used the pbs dump command I didn't see any service using this property as well.

sunshinejr:
Also, for some reason when I used the service today (without any code changes) the app would freeze both itself and my Finder 🤔 Wonder what happened.

sindresorhus:
> the app would freeze both itself and my Finder 🤔 Wonder what happened.

I experienced that too. I thought it was just me doing something wrong.

kornelski:
If it's freezing Finder, it may be because the conversion is blocking? The callback should return ASAP and just schedule conversion for later.

sunshinejr:
@kornelski that's a good idea. Although it was working before and, from what I understand, the system service is opening the app for us and then we can do whatever we want in that function as the app is opened.

I did clean everything, restarted the computer (as I thought that the manual Finder opening with debug flag could cause this) and for some reason when I built and moved the app again to /Applications folder, when tapped on "Convert to GIF with Gifski" i**t opened the version of the app that I didn't even archive before**, lol (I'm pretty sure about that because it opened a version of the app with the alert as I was testing #84 for a few minutes yesterday).

So I've removed the app from /Applications, removed all previous archives, cleaned the project, did flush the pbs (`/System/Library/CoreServices/pbs -flush`) and tried the steps that I did before (basically what is in the maintaining.md) and boom - it worked. Not sure which step helped, though. Really weird.

@sindresorhus did it work for you even once and then stopped, or was it broken all the time?

readme.md
```diff
@@ -23,6 +23,18 @@ Requires macOS 10.13 or later.
 [![](https://linkmaker.itunes.apple.com/assets/shared/badges/en-us/macappstore-lrg.svg)](https://itunes.apple.com/no/app/gifski/id1351639930?mt=12)
 
 
+## System service
+
+Gifski provides a [system service](https://www.computerworld.com/article/2476298/os-x-a-quick-guide-to-services-on-your-mac.html) that lets you quickly convert a video to GIF from anywhere.
+
+You first have to enable it:
```
sindresorhus: Are you sure the user has to enable it manually? From experience, most services are auto-enabled when I install new apps.

This seems to confirm it: https://stackoverflow.com/questions/3859747/how-do-i-automatically-activate-an-item-in-the-os-x-services-menu

sindresorhus:
> So I've removed the app from /Applications, removed all previous archives, cleaned the project, did flush the pbs (/System/Library/CoreServices/pbs -flush) and tried the steps that I did before (basically what is in the maintaining.md) and boom - it worked. Not sure which step helped, though. Really weird.

Same. It's working now for me too. So was definitely something weird going on with different builds.

readme.md
sunshinejr: Not totally sure, I had it unchecked 100% though. We may indicate in docs that we should make sure it's enabled in the settings.

sindresorhus:
> Also, I didn't use NSServiceDescription and also don't quite understand where would it be seen. 

I discovered where it's used now. If you hover over a service in the Services preferences for a couple of settings, the description will be shown in a tooltip.

Screen Shot 2019-05-22 at 12 58 10

readme.md
sindresorhus: I have manually verified that services are enabled by default.

PR #83, opened 2019-05-14
Add dimensions dropdown and width/height inputs to the save panel

### LINKED ISSUE #36: Add predefined size dropdown
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/gifski-app/issues/36)

We should consider changing the slider scale to a dropdown of predefined sizes and textfields for precise control.

Something like what preview.app has:

41196637-6a6871ac-6c45-11e8-87ce-3e2e72986592

The predefined sizes should be a union of (ref [comment](https://github.com/sindresorhus/gifski-app/issues/25#issuecomment-373951951)):

> - 1/nth of the input size: 1/2, 1/3, 1/4, 1/5, …, etc.
> - a predefined list of common sizes: 960, 800, 640, 500, 480, 320, 256, 200, 160, 128, 80, 64

As mentioned in the comment 1/n scaling gives the best quality so we should maybe mark those in the dropdown.

---

**Note:** This issue requires you to have a good understanding of Swift and macOS/iOS development.

> There is a $200.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/gifski-app/issues/36).

sunshinejr:
Fixes #36. This is the feature I want the most currently, so I'm gonna focus on it for the next few days. Supersedes #49. Most of the work was done so I'm gonna try to complete missing pieces.

Currently WIP status, gonna add you as a reviewer, Sindre, once I'm finished.

Todo:
- [x] Fix constraints in the save panel (they are breaking).
- [x] Clean the code a little bit.
- [x] Entering invalid size should shake and go back to min/max values.
- [x] Add modifiers to arrows (option + arrow up/down should move by 10, not by 1)


sunshinejr:
@sindresorhus it's finally ready for review! I have also created a unit test bundle with tests for `ResizableDimensions` as well, but didn't want to clutter the PR as it's a lot to review. Will make a PR once we resolve all the issues found here 😉

sindresorhus:
Awesome!

---

Screen Shot 2019-06-17 at 11 21 49

The "Width" and "Height" textfields should be filled out here initially too. So the user can quickly manually change the size.

---

When I selected 320x320 in the dropdown, I got a different value in the custom size textfields:

Screen Shot 2019-06-17 at 11 24 55

---

What do you think about dropping the `pixels` label? I think it's already clear that it's pixels and it's already shown in the percent/pixels dropdown anyway.

Screen Shot 2019-06-17 at 11 26 41

---

I think we could also drop the "GIF Dimensions" and "Other Settings" header labels. I don't think they add much value.

---

In the "Width" textfield, if I press arrow up/down keys cursor stays on the right side, but when holding Option key down, the cursor moves to the left side of the text. It should stay on the right side.

---

In the "Width" textfield, if it contains the value `23` and I type something incorrect, like `,`, it shakes and then resets to `1`. I don't think this is a good behavior. Instead it should just keep the existing valid value and ignore `,`, while still shake.

---


Because of our accessory view, the save panel doesn't have much space by default. I think we could use the trick to set either `NSNavPanelExpandedSizeForSaveMode` or `NSWindow Frame NSNavPanelAutosaveName` UserDefault key to create a larger panel size by default.

Screen Shot 2019-06-17 at 11 33 49

sindresorhus:
The functionality to use arrow up/down (and Option) to increase/decrease the size in the text fields are super useful, but not very discoverable. Can you document it in the readme? Would also be nice to show it in the UI somehow, but not sure how to present it. A one-time tooltip kinda thing would be nice, but not sure. Thoughts?

Gifski/ResizableDimensions.swift
```diff
@@ -0,0 +1,142 @@
...
+
+	init(dimensions: Dimensions, minimumScale: CGFloat? = nil, maximumScale: CGFloat? = nil) {
+		self.originalDimensions = dimensions
+		self.currentDimensions = dimensions
+		self.minimumScale = minimumScale ?? 0.01
+		self.maximumScale = maximumScale ?? 1.0
+		self.currentScale = 1.0
+	}
+
+	private func copy() -> ResizableDimensions {
```
sindresorhus: Would be good to use a protocol for this. https://stackoverflow.com/a/35915186/64949

sindresorhus:
You accidentally made some of the files executable: `100644 → 100755`. Look at the diff view.

Gifski/SavePanelAccessoryViewController.swift
```diff
@@ -83,4 +186,61 @@ final class SavePanelAccessoryViewController: NSViewController {
...
+
+extension SavePanelAccessoryViewController: NSTextFieldDelegate {
+	func controlTextDidChange(_ obj: Notification) {
+		guard let textField = obj.object as? NSTextField else {
+			return
+		}
+		scalingTextFieldTextDidChange(textField)
+	}
+
+	func control(_ control: NSControl, textView: NSTextView, doCommandBy commandSelector: Selector) -> Bool {
```
sindresorhus: See https://github.com/sindresorhus/gifski-app/pull/49#discussion_r248931694

Gifski/SavePanelAccessoryViewController.swift
```diff
@@ -1,71 +1,174 @@
...
+	}
 
-		func estimateFileSize() {
-			let frameCount = videoMetadata.duration * frameRateSlider.doubleValue
-			var fileSize = (Double(currentDimensions.width) * Double(currentDimensions.height) * frameCount) / 3
-			fileSize = fileSize * (qualitySlider.doubleValue + 1.5) / 2.5
-			estimatedSizeLabel.stringValue = formatter.string(fromByteCount: Int64(fileSize))
+	private func setupDropdowns() {
+		predefinedSizesDropdown.removeAllItems()
+		predefinedSizes.forEach { size in
```
sindresorhus: I prefer using normal for loops.

sindresorhus: I prefer using normal for loops.

sindresorhus:
When setting a custom size in percentage, it's not possible to see what it is in dimensions (pixel size). Would be useful to show the pixel size live in the dropdown picker:

Screen Shot 2019-06-17 at 12 09 02

(But not in the actual dropdown)

Same for the inverse. So if you're manually setting pixel size in width/height, it would show the percentage.

sindresorhus:
For the "Width"/"Height" inputs, I think it should be possible to clear them. A common use-case, is to Command+A to select everything, press Delete, and then write a new number. So I think we should do the validation that ensures it's not empty on blur instead of on input changes. That would make the validation less annoying for users.

sindresorhus:
If I select a video that has the dimensions 480x270. Then try some custom sizes. Then in the Width/Height input fields, I write 480 and then 270, it should select the "480x270 (Original)" item in the Dimensions picker, instead of showing "Custom" there. Nitpick, I know. But perfecting it like this makes for a better user-experience as the user can be sure they went back to the original size, or half the size, etc.

Gifski/MainWindowController.swift
```diff
@@ -214,6 +215,8 @@ final class MainWindowController: NSWindowController {
 			if $0 == .OK {
 				self.startConversion(inputUrl: inputUrl, outputUrl: panel.url!)
 			}
+
+			self.accessoryViewController = nil
```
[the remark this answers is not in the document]
sunshinejr: When I fixed retain cycles in the save panel, it turned out that no-one was keeping the reference to the it, so it was alive just because we had retain cycles. We only pass the view to the `accessoryView` so someone has to hold the controller itself. And because of that we should nil it when not using.

Gifski/SavePanelAccessoryViewController.swift
sunshinejr: Oh sure, I don't have a strong opinion on that one so will replace it 👍

Gifski/util.swift
```diff
@@ -66,6 +66,14 @@ extension NSColor {
 
 
 extension NSView {
+	func shake(duration: TimeInterval = 0.3) {
```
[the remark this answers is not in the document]
sunshinejr: Good idea 👍

Gifski/ResizableDimensions.swift
sunshinejr: Good idea 👍

Gifski/SavePanelAccessoryViewController.swift
sunshinejr: Oh, missed it. Will do for sure.

sunshinejr:
Hey @sindresorhus, first of all thanks for a great feedback and don't worry about nitpicking - if we can make it even better let's do it 😉 Below is a list that I will be updating shortly.

TODO:
- [ ] Initially fill width/height textfields
- [ ] Fix dropdown selection (when selected  320x320 it filled 480x480)
- [ ] Remove pixels label
- [ ] Remove „Gif Dimensions” and „Other Settings” labels
- [ ] When holding option and moving up/down, cursor should stay on the right (currently goes to the left)
- [ ] Create larger panel size by default (NSNavPanelExpandedSizeForSaveMode/NSWindow Frame NSNavPanelAutosaveName)
- [ ] When typing incorrect value, shake but do not correct the value immediately
- [ ] Correct incorrect value on leaving the textfield only (check if converting without leaving the textfield will work correctly as well)
- [ ] Add a tooltip that says an „option + arrow” could speed up resizing the image
- [ ] Add a doc paragraph to readme about „option  + arrows” as well
- [ ] Show pixel size / percent size with  „Custom" in the dropdown label, but not on the  dropdown list
- [ ] When writing custom width/height, try to select the correct predefined size before selecting custom

CR fixes:
- [ ] Shake func should accept direction argument (horizontal/vertical)
- [ ] Subscript func minor updates
- [ ] Create a subclass for NSTextField that handles keys and validation
- [ ] Replace `forEach` wtih `for` loops

Gifski/MainWindowController.swift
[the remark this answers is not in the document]
sindresorhus: Ah, ok. Makes sense.

Gifski/IntTextField.swift
```diff
@@ -0,0 +1,61 @@
...
+		super.init(frame: frameRect)
+		setup()
+	}
+
+	private func setup() {
+		delegate = self
+	}
+
+	override func performKeyEquivalent(with event: NSEvent) -> Bool {
+		let key = Int(event.keyCode)
```
sindresorhus: I think you can use https://developer.apple.com/documentation/appkit/nsevent/3019455-specialkey for this instead.

Gifski/IntTextField.swift
```diff
@@ -0,0 +1,61 @@
+import AppKit
+import Carbon
+
+final class IntTextField: NSTextField, NSTextFieldDelegate {
+	// Delta used for arrow navigation
+	var delta = 1
+
+	// Delta used for option + arrow navigation
+	var biggerDelta = 10
```
sindresorhus: I would go with `alternativeDelta`.

Gifski/SavePanelAccessoryViewController.swift
```diff
@@ -145,12 +146,51 @@ final class SavePanelAccessoryViewController: NSViewController {
...
+				return
+			}
+
+			let validated = self.resizableDimensions.validate(newHeight: CGFloat(height)) == true
+			if validated {
+				self.resizableDimensions.resize(usingHeight: CGFloat(height))
+				self.dimensionsUpdated()
+				self.predefinedSizesDropdown.selectItem(at: 0)
+			} else {
+				self.heightTextField.shake(direction: .horizontal)
```
sindresorhus: I think this should be a method on the `IntTextField`. Like `.indicateValidationFailure()` or something better.

Gifski/IntTextField.swift
```diff
@@ -0,0 +1,61 @@
...
+
```
sindresorhus: I was thinking that you could set the minimum and maximum value here and have the textfield itself value those simple cases. I want to make it as reusable as possible.

Gifski/IntTextField.swift
```diff
@@ -0,0 +1,61 @@
...
+		let currentValue = Int(stringValue) ?? 0
+		let newValue = currentValue + delta
+		stringValue = "\(newValue)"
+		onTextDidChange?(newValue)	
+
+		return true
+	}
+
+	func controlTextDidChange(_ obj: Notification) {
```
sindresorhus: I know the default is `obj`, but for internal labels, we can use any name, so I prefer `object`.

Gifski/IntTextField.swift
```diff
@@ -0,0 +1,61 @@
...
+		let isHoldingOption = event.modifierFlags.contains(.option)
+
+		let delta: Int
+		switch (key, isHoldingOption) {
+		case (kVK_UpArrow, true):
+			delta = self.biggerDelta
+		case (kVK_UpArrow, false):
+			delta = self.delta
+		case (kVK_DownArrow, true):
+			delta = -1 * self.biggerDelta
```
sindresorhus: I would reverse that statement.

Gifski/IntTextField.swift
sunshinejr: Oh nice didn't notice that! Thanks, updated 👍

Gifski/IntTextField.swift
```diff
@@ -0,0 +1,61 @@
...
+
+	private func setup() {
+		delegate = self
+	}
+
+	override func performKeyEquivalent(with event: NSEvent) -> Bool {
+		let key = Int(event.keyCode)
+		let isHoldingOption = event.modifierFlags.contains(.option)
+
+		let delta: Int
```
sindresorhus:
```swift
let delta = isHoldingOption ? biggerDelta : delta
```

Gifski/IntTextField.swift
sunshinejr: Hmm, the problem here is that the minimum/maximum will change with each type change (pixels/percent). Do you think we should go with min/max still and set it each time we change the type or have a "validate" closure that we could pass to the text field?

Gifski/IntTextField.swift
sunshinejr: it would be nice simplification, but we still need a switch to decide on multiplier so I think we should leave it as is to make the intent clearer (one statement decides the delta value vs multiple lines in and out of switch)

sindresorhus: Right, I misread the code. I meant:

```swift
override func performKeyEquivalent(with event: NSEvent) -> Bool {
		let key = event.specialKey
		let isHoldingOption = event.modifierFlags.contains(.option)
		let initialDelta = isHoldingOption ? alternativeDelta : delta

		let delta: Int
		switch key {
		case .upArrow?:
			delta = initialDelta
		case .downArrow?:
			delta = initialDelta * -1
		default:
			return super.performKeyEquivalent(with: event)
		}
```

Gifski/IntTextField.swift
sindresorhus: > Do you think we should go with min/max still and set it each time we change the type

Yeah, I don't see a problem with that. We manually set a lot of properties elsewhere.

sunshinejr:
@sindresorhus do you know what are the steps to reproduce the dropdown selection bug? Couldn’t make it break yet.

sindresorhus:
Do you mean this one?

> Fix dropdown selection (when selected 320x320 it filled 480x480)

sunshinejr:
@sindresorhus yeah

sindresorhus:
I just tried the latest commit here and can still reproduce.

Here's the video I used to reproduce it with: [Test - super short.mp4.zip](https://github.com/sindresorhus/gifski-app/files/3296892/Test.-.super.short.mp4.zip)

Gifski/util.swift
```diff
@@ -2136,3 +2145,20 @@ extension NSView {
...
+	}
+}
+
+protocol Copyable {
+	init(instance: Self)
+}
+
+extension Copyable {
+	func copy() -> Self {
+		return Self.init(instance: self)
```
sindresorhus:
```suggestion
		return Self(instance: self)
```

sunshinejr: oh interesting, it works without specifying `.init` - I remember some time ago you wouldn't be able to do so in protocol extension, wonder when that changed

sunshinejr:
@sindresorhus oh wow this needed to be exactly the same width as one of the predefined sizes. If you figured this case out and prepared this video to let me figure it by myself, you're a genius 😄

Gifski/CustomButton.swift
```diff
@@ -122,6 +122,10 @@ extension CGSize {
 			height: height
 		)
 	}
+
+	func rounded(_ rule: FloatingPointRoundingRule = .toNearestOrAwayFromZero) -> CGSize {
+		return CGSize(width: width.rounded(rule), height: height.rounded(rule))
+	}
```
sindresorhus: Can you put this in `util.swift`? It's not used in CustomButton.

Gifski/SavePanelAccessoryViewController.swift
```diff
@@ -53,7 +53,11 @@ final class SavePanelAccessoryViewController: NSViewController {
 		let dimensions = Dimensions(type: .pixels, value: videoMetadata.dimensions)
 		resizableDimensions = ResizableDimensions(dimensions: dimensions, minimumScale: minimumScale, maximumScale: maximumScale)
 
-		let pixelCommonSizes: [CGFloat] = [dimensions.value.width, 960.0, 800.0, 640.0, 500.0, 480.0, 320.0, 256.0, 200.0, 160.0, 128.0, 80.0, 64.0].sorted(by: >)
+		var pixelCommonSizes: [CGFloat] = [960.0, 800.0, 640.0, 500.0, 480.0, 320.0, 256.0, 200.0, 160.0, 128.0, 80.0, 64.0]
+		if !pixelCommonSizes.contains(dimensions.value.width) {
+			pixelCommonSizes.append(dimensions.value.width)
+			pixelCommonSizes.sort(by: >)
+		}
```
sindresorhus: Wish Swift had a SortedSet data structure built-in. Would have been perfect for this.

Gifski/CustomButton.swift
sunshinejr: Damn didn't notice that the other `rounded` functions were not in utils but in `CustomButton` file 🤦‍♂

Gifski/util.swift
```diff
@@ -2168,3 +2168,9 @@ extension CGSize {
 		return CGSize(width: width.rounded(rule), height: height.rounded(rule))
 	}
 }
+
+extension BinaryFloatingPoint {
+	func isEqual(to otherValue: Self, roundingDecimalPlaces decimalPlaces: Int) -> Bool {
+		return rounded(toDecimalPlaces: decimalPlaces) == otherValue.rounded(toDecimalPlaces: decimalPlaces)
+	}
+}
```
sindresorhus: I would just use a polyfill of this: https://github.com/apple/swift-evolution/blob/master/proposals/0259-approximately-equal.md

sunshinejr: Good idea 👍

sunshinejr: Good idea 👍

sindresorhus:
> oh wow this needed to be exactly the same width as one of the predefined sizes. If you figured this case out and prepared this video to let me figure it by myself, you're a genius 😄

I wish I could say that, but it's just a pure lucky coincidence. That's just one of my existing fixtures to test stuff with Gifski.

sindresorhus:
Feedback based on the latest commit:

---

For a video with dimensions 480x270, if I set the "Width" input to 478, it shows "Custom - 100%" in the dropdown. I don't think it should show 100% until it's exactly 100%, in which it would show "480x270 (Original)" instead. I would rather round down then.

---

For a video with dimensions 480x270, if I type 490 in the "Width" input, it no longer resets directly back to 480. That validation should not be deferred to input blur. For the mimimum size, I think it makes sense to directly reset to the previous valid value if it encounters: higher than maximum number, negative number, 0, and non-integer characters. For example: `600`, `-4`, `0`, `,`, `@`. However, we must allow `4` even though it's less than the minimum value of `5` as it should be possible to write `400`. In that case, we'll validate and correct on blur instead.

---

If something is validated as invalid on blur, it should also shake to make it clear to the user why it changed the value.

---

If I click on the Dimensions dropdown, shouldn't it then get the focus? So I can then continue tabbing to the "Width" input, etc.

---

In the Dimensions dropdown menu, do you think we should vertically align the alternative values (in parens)?

```
480 x 270  (Original)
240 x 135  (50%)
80 x 60    (20%)
24 x 20    (7%)
```

(Only in the actual dropdown menu, not in the dropdown preview)

And should we maybe have two spaces between them to give them some breathing room?

sunshinejr:
@sindresorhus 

- [x] Validation on blur
- [x] Rounding for dimensions type change from pixels to percent goes down rather that to the nearest
- [x] Dimensions + pixels|percent dropdowns get focus on clicking
- [x] Fixed some scaling calculations problems
- [ ] Rewrite validation logic
- [ ] Vertically aligning values in dropdown
---
_leftovers from the previous comment:_
- [ ] Create larger panel size by default (NSNavPanelExpandedSizeForSaveMode/NSWindow Frame NSNavPanelAutosaveName)
- [ ] Add a tooltip that says an „option + arrow” could speed up resizing the image
- [ ] Add a doc paragraph to readme about „option + arrows” as well
- [ ] Bring back 644 file permissions

---
So the new validation logic you mentioned is gonna take some time I think. Also I don't really know how you can align the values in dropdown menu, is there a better way then doing it manually in strings?

sindresorhus:
> Also I don't really know how you can align the values in dropdown menu, is there a better way then doing it manually in strings?

Possible, but not better. Strings are the easiest.

I quickly hacked together this:

```swift
var x = [
	"100x100",
	"10x10",
	"65x121",
	"1x1"
]

extension Collection where Element == String {
	/// Pad the elements to equal length.
	func paddingToEqualLength() -> [String] {
		guard let maxLength = max(by: { $1.count > $0.count })?.count else {
			return []
		}

		return map { $0.padding(toLength: maxLength, withPad: " ", startingAt: 0) }
	}
}

x = x
	.paddingToEqualLength()
	.map { "\($0)  (100%)" }

print(x.joined(separator: "\n"))
```

```
100x100  (100%)
10x10    (100%)
65x121   (100%)
1x1      (100%)
```

sunshinejr:
@sindresorhus was trying to do align vertically those strings but it's harder than expected. Because some of the letters are wider than the others, this seems to not only be a matter of added spaces at indexes. I think an easier way would be to create custom views for that and base it on constraints/pixels, but still I'd skip that for this iteration.

Zrzut ekranu 2019-06-20 o 20 50 54

Gifski/SavePanelAccessoryViewController.swift
```diff
@@ -1,86 +1,257 @@
...
+		setupDropdowns()
+		setupSliders()
+		setupWidthAndHeightTextFields()
+	}
+
+	override func viewWillAppear() {
+		super.viewWillAppear()
+
+		// Hack to enlarge extended save panel frame
+		UserDefaults.standard.set("{841, 681}", forKey: "NSNavPanelExpandedSizeForSaveMode")
```
sunshinejr: added 200pts in height from what we had before

sindresorhus:
> Because some of the letters are wider than the others, this seems to not only be a matter of added spaces at indexes.

Hmm, we could use NSAttributedString with a monospace font, but I'm not sure monospace would look good there. I think what you have there is fine and we can iterate on it later on.

> Add a tooltip that says an „option + arrow” could speed up resizing the image

A tooltip like this (probably a NSPopover) that stays there until you click somewhere would be nice for this:

Screen Shot 2019-06-19 at 7 57 16 AM

Gifski/SavePanelAccessoryViewController.swift
sindresorhus: Can you add the original initial size in a comment, so it's clear how much it changes when reading the code?

sindresorhus:
- If I write something in the "Width" input and then tab, it goes to the "pixels" dropdown, but it should move focus to "Height" first and then "pixels".
- Some of the labels are not horizontally aligned with their input (widget).
- I noticed some IB warnings:
Screen Shot 2019-06-21 at 15 00 09

sunshinejr:
@sindresorhus

- [x] Update focus order to `width -> height -> pixels` from `width -> pixels -> height`
- [x] Add initial frame size for extended save panel in the comment
- [x] Fix misaligned labels
- [x] Fix clipping content warnings
---
_leftovers from the previous comments:_
- [ ] Add a tooltip that says an „option + arrow” could speed up resizing the image
- [ ] Add a doc paragraph to readme about „option + arrows” as well
- [ ] Rewrite validation logic

Gifski/IntTextField.swift
```diff
@@ -0,0 +1,91 @@
+import AppKit
+import Carbon
```
sindresorhus: Not needed anymore.

```suggestion
```

sunshinejr:
@sindresorhus added the tooltip. It dismisses on both click inside and outside. Play with the positioning/size/text as you like. Below a screenshot of a current version:
Zrzut ekranu 2019-06-21 o 18 41 29

sindresorhus:
Very cool. Exactly what I had in mind!

Can we add an option to show once, and then it would manage it's own userdefault setting and checking? Would be nice to not have to think about it. We could require to set a unique id/name for the tooltip. And it would use a userdefault like: `__Tooltip_id` or something.

---

Screen Shot 2019-06-22 at 00 10 09

The text should not be selectable.

---

I will definitely add this tooltip to the conversion completed view later on. Many people don't realize you can drag the thumbnail there.

Gifski/SavePanelAccessoryViewController.swift
```diff
@@ -60,6 +60,12 @@ final class SavePanelAccessoryViewController: NSViewController {
 		dimensionsTypeDropdown.nextKeyView = heightTextField.nextKeyView
 		widthTextField.nextKeyView = heightTextField
 		heightTextField.nextKeyView = dimensionsTypeDropdown
+
+		if !defaults[.tooltipCmdArrowsShown] {
+			defaults[.tooltipCmdArrowsShown] = true
+			let tooltip = Tooltip(text: "You can also navigate using arrows. Hold CMD and you can navigate even faster!", contentInsets: .init(value: 6.0), maxWidth: 250.0)
+			tooltip.show(relativeTo: widthTextField.bounds, of: widthTextField, preferredEdge: .maxX)
```
sindresorhus: Can you add an overload that is just:

```swift
tooltip.show(from: widthTextField, preferredEdge: .maxX)
```

With `preferredEdge` defaulting to top.

Gifski/Tooltip.swift
```diff
@@ -0,0 +1,93 @@
+import AppKit
+
+final class Tooltip: NSPopover {
+	init(text: String, closeOnClick: Bool = true, contentInsets: NSEdgeInsets = .zero, maxWidth: CGFloat? = nil) {
```
sindresorhus: `CGFloat` => `Double`

I prefer using `Double` in my own APIs.

Gifski/Tooltip.swift
```diff
@@ -0,0 +1,93 @@
...
+		self.clickRecognizer = clickRecognizer
+	}
+
+	@objc
+	private func didClick() {
+		onClick?()
+	}
+}
+
+extension NSEdgeInsets {
```
sindresorhus: This should be in `util.swift` too.

Gifski/Tooltip.swift
```diff
@@ -0,0 +1,93 @@
...
+	static var zero: NSEdgeInsets {
+		return NSEdgeInsetsZero
+	}
+
+	init(value: CGFloat) {
```
sindresorhus: I would name it `value` => `all`, and `Double`.

Gifski/Tooltip.swift
sindresorhus: I think `contentInsets` should default to `10` and we should use the default in our tooltip. It's a good-looking default.

I believe in sensible defaults, so I think we should default `maxWidth` to `200`, but still accept `nil` in case the user doesn't want a default. The reason for 200 is that tooltips are meant to be small UI with not much text.

Gifski/Tooltip.swift
```diff
@@ -0,0 +1,93 @@
...
+		setupContent(text: "", closeOnClick: true, contentInsets: .zero, maxWidth: nil)
+	}
+
+	private func setupContent(text: String, closeOnClick: Bool, contentInsets: NSEdgeInsets, maxWidth: CGFloat?) {
+		contentViewController = ToolTipViewController(text: text, contentInsets: contentInsets, maxWidth: maxWidth) { [weak self] in
+			if closeOnClick {
+				self?.close()
+			}
+		}
+		behavior = closeOnClick ? .transient : .applicationDefined
```
sindresorhus: `transient` => `semitransient`

Clicking outside the app should not close the tooltip.

Gifski/Tooltip.swift
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	required init?(coder: NSCoder) {
+		self.text = ""
+		self.contentInsets = .zero
+		super.init(coder: coder)
+	}
+
+	override func loadView() {
+		let wrapperView = NSView()
```
sindresorhus: Why is a wrapper view needed?

Gifski/Tooltip.swift
```diff
@@ -0,0 +1,93 @@
...
+	}
+}
+
+fileprivate final class ToolTipViewController: NSViewController {
+	fileprivate var text: String
+	fileprivate var contentInsets: NSEdgeInsets
+	fileprivate var maxWidth: CGFloat?
+	fileprivate var onClick: (() -> Void)?
+
+	private var clickRecognizer: NSClickGestureRecognizer?
```
sindresorhus:
```swift
private let clickRecognizer = NSClickGestureRecognizer(target: self, action: #selector(didClick))
```

Gifski/SavePanelAccessoryViewController.swift
```diff
@@ -60,6 +60,12 @@ final class SavePanelAccessoryViewController: NSViewController {
 		dimensionsTypeDropdown.nextKeyView = heightTextField.nextKeyView
 		widthTextField.nextKeyView = heightTextField
 		heightTextField.nextKeyView = dimensionsTypeDropdown
+
+		if !defaults[.tooltipCmdArrowsShown] {
+			defaults[.tooltipCmdArrowsShown] = true
+			let tooltip = Tooltip(text: "You can also navigate using arrows. Hold CMD and you can navigate even faster!", contentInsets: .init(value: 6.0), maxWidth: 250.0)
```
sindresorhus: > Press the arrow up/down keys to change the value by 1. Hold the Option key meanwhile to change it by 10.

Gifski/Tooltip.swift
sunshinejr: I use it for insets - do you if I can do it without the wrapper view?

Gifski/SavePanelAccessoryViewController.swift
sunshinejr: ah right good catch

Gifski/util.swift
```diff
@@ -938,6 +938,22 @@ extension NSView {
 		addSubview(view)
 		view.center(inView: superview!)
 	}
+
+	func pinToSuperview(insets: NSEdgeInsets = .zero) {
```
sindresorhus: This is not used anywhere.

Gifski/Tooltip.swift
```diff
@@ -0,0 +1,93 @@
...
+		let wrapperView = NSView()
+		let textField = NSTextField(wrappingLabelWithString: text)
+
+		if let maxWidth = maxWidth {
+			textField.frame.size = textField.sizeThatFits(NSSize(width: maxWidth - contentInsets.horizontal, height: .greatestFiniteMagnitude))
+		}
+		textField.frame.origin = CGPoint(x: contentInsets.left, y: contentInsets.top)
+
+		let contentSize = textField.frame.size
+		let size = CGSize(width: contentSize.width + contentInsets.horizontal, height: contentSize.height + contentInsets.vertical)
```
sindresorhus: I would use auto-layout here instead of manual frame setting.

Gifski/Tooltip.swift
sindresorhus: You can set auto-layout constraints on `textField` against the superview, with inset. I thought that's what the `pinToSuperview` thing was for?

sindresorhus:
By the way, I would like to invite you as a maintainer of Gifski, if you're interested? There's no commitment implied. You do whatever you want or nothing. I just want to show off your name in the readme for the amazing work you've been doing. (And you can still collect bounties as a maintainer).

Gifski/Tooltip.swift
sunshinejr: True, I tried that, but when pinned to the superview wrapper it wasn't looking good - probably it has it's own position/layout (e.g. margin-left to pad it from the arrow).

Gifski/util.swift
sunshinejr: Will be! Replacing the manual frame estimation with auto-layout as we speak.

sunshinejr:
@sindresorhus oh, that would be awesome! Thanks!

Gifski/Tooltip.swift
sindresorhus: Ok. Doesn't matter then. A wrapper view is fine. Just wanted to make sure it was actually needed.

sunshinejr:
@sindresorhus fixed issues + added show behavior.

Gifski/Tooltip.swift
```diff
@@ -1,20 +1,33 @@
...
+
+	init(text: String, showBehavior: ShowBehavior, closeOnClick: Bool = true, contentInsets: NSEdgeInsets = .init(all: 10.0), maxWidth: Double? = nil) {
+		self.showBehavior = showBehavior
 		super.init()
+
 		setupContent(text: text, closeOnClick: closeOnClick, contentInsets: contentInsets, maxWidth: maxWidth)
 	}
 
 	required init?(coder: NSCoder) {
+		self.showBehavior = .always(identifier: NSUUID().uuidString)
```
sindresorhus: `NSUUID` => `UUID`

Gifski/Tooltip.swift
```diff
@@ -1,20 +1,33 @@
...
 		super.init(coder: coder)
+
 		setupContent(text: "", closeOnClick: true, contentInsets: .zero, maxWidth: nil)
 	}
 
 	func show(from positioningView: NSView, preferredEdge: NSRectEdge) {
 		show(relativeTo: positioningView.bounds, of: positioningView, preferredEdge: preferredEdge)
 	}
 
+	override func show(relativeTo positioningRect: NSRect, of positioningView: NSView, preferredEdge: NSRectEdge) {
```
sindresorhus: `NSRect` => `CGRect`

Gifski/ShowBehavior.swift
```diff
@@ -0,0 +1,32 @@
...
+		switch self {
+		case .always:
+			return true
+		case .onlyOnce:
+			return showCount < 1
+		}
+	}
+
+	private var key: Defaults.Key<Int> {
+		return Defaults.Key<Int>(identifier, default: 0)
```
sindresorhus: We should namespace it to prevent possible conflicts.

Gifski/Tooltip.swift
```diff
@@ -1,20 +1,33 @@
 import AppKit
 
 final class Tooltip: NSPopover {
-	init(text: String, closeOnClick: Bool = true, contentInsets: NSEdgeInsets = .init(all: 10.0), maxWidth: Double? = nil) {
+	private let showBehavior: ShowBehavior
+
+	init(text: String, showBehavior: ShowBehavior, closeOnClick: Bool = true, contentInsets: NSEdgeInsets = .init(all: 10.0), maxWidth: Double? = nil) {
```
sindresorhus: Since the `identifier` is required regardless, wouldn't it be better to just have it as a top-level argument here?

`Tooltip(identifier: "sadad", text: "sdas", showBehavior: .once)`

sunshinejr: Well, yeah but then I wouldn't be able to use it as a separate type (I need to store the identifier inside the ShowBehavior). I could try to use a struct/class:
```swift
struct ShowBehavior {
	enum Option {
		case always
		case once
	}

	let identifier: String
	let option: Option

	static func always(identifier: String) -> ShowBehavior {
		return .init(identifier: identifier, option: .always)
	}

	static func once(identifier: String) -> ShowBehavior {
		return .init(identifier: identifier, option: .once)
	}
}
``` 

what do you think?

sindresorhus: 👍

sunshinejr: done 👍

sunshinejr:
@sindresorhus do you think we should still document the arrows + option in the readme with this new tooltip?

sindresorhus:
Is there a reason `.always()` require an identifier? It's not really used.

sindresorhus:
> do you think we should still document the arrows + option in the readme with this new tooltip?

Yes, it's useful as a selling point and also for users to refer to if they missed the tooltip or forgot.

Gifski/ShowBehavior.swift
```diff
@@ -0,0 +1,38 @@
+struct ShowBehavior {
```
sindresorhus: I think this type should be nested in `Tooltip`. It's not really useful on it's own.

Gifski/ShowBehavior.swift
```diff
@@ -0,0 +1,38 @@
...
+
+	static func once(identifier: String) -> ShowBehavior {
+		return .init(identifier: identifier, option: .once)
+	}
+
+	let identifier: String
+	let option: Option
+
+	private var key: Defaults.Key<Int> {
+		return Defaults.Key<Int>("__showBehavior_" + identifier, default: 0)
```
sindresorhus:
```suggestion
		return Defaults.Key<Int>("__Tooltip__\(identifier)", default: 0)
```

sindresorhus: Why is it an Int, can't we just make it a Bool of whether it has been shown or not?

Gifski/util.swift
```diff
@@ -892,6 +904,30 @@ extension NSView {
...
+
+		NSLayoutConstraint.activate([
+			leadingAnchor.constraint(equalTo: superview.leadingAnchor, constant: insets.left),
+			trailingAnchor.constraint(equalTo: superview.trailingAnchor, constant: -insets.right),
+			topAnchor.constraint(equalTo: superview.topAnchor, constant: insets.top),
+			bottomAnchor.constraint(equalTo: superview.bottomAnchor, constant: -insets.bottom)
+		])
+	}
+
+	func constrain(size: CGSize) {
```
sindresorhus:
```suggestion
	func constrain(to size: CGSize) {
```

?

sunshinejr:
The reason I added identifier to `.always` as well, was to cover a case where you would like to switch to other type after a release. This way we can track the number of times it was shown in case we need to use it.

Gifski/ShowBehavior.swift
sunshinejr: I guess I was just being proactive about what the `ShowBehavior` can become in the future and it's not that big of an overhead to use count instead of a flag.

readme.md
```diff
@@ -23,6 +23,10 @@ Requires macOS 10.13 or later.
 [![](https://linkmaker.itunes.apple.com/assets/shared/badges/en-us/macappstore-lrg.svg)](https://itunes.apple.com/no/app/gifski/id1351639930?mt=12)
 
 
+## Resizing GIF using keyboard
+
+Press the arrow up/down keys to change the value by 1. Hold the Option key meanwhile to change it by 10.
```
sunshinejr: This might use a copy change + maybe a screenshot? Or a GIF?

sindresorhus:
It’s only useful one way though. Always to once. But I don’t really buy the use-case that you would want to reuse the usage counter. If you have shown something many times, why would you suddenly switch it to once? At least not without changing the content, and with new content you probably would want to show it once again.

sunshinejr:
@sindresorhus yeah in terms of the `ShowBehavior` for the Tooltip that we currently have, I agree. I think I was thinking of a more generic `ShowBehavior` that I could use in the other apps I'm writing. I was thinking of use cases like e.g. when we show an intro twice in given timespan and our data/product team likes to change these stuff but based on the actions we had previously (so if it popped twice we would only like to pop one more but for the new users it should pop three times). Or the rating pop-up, we wanted to show it once in 2 weeks but then the team decided that we should try making it one a week - also previous count was needed.

I just thought that it's doing no harm while being there but could be useful in the future - we could add additional `static var always()` that would use auto-generated `UUID`.

readme.md
sindresorhus: A GIF would show it off the best, but I don't know any good ways to make a high-quality GIF...

sindresorhus:
Thanks for elaborating. Your choices makes a lot more sense now. It kinda seems like you could use something even more flexible though. How about we require `identifier` as a top-level argument, expose the showCount on the Tooltip instance, and then make the `showBehavior` argument just a boolean called `showOnlyOnce`? By default it would always show, if you set `showOnlyOnce: true` it would show only once, if you need custom logic for showing, for example to show it twice, you can just wrap the `tooltip.show()` in an if-conditional and check `tooltip.showCount`.

readme.md
sunshinejr: Would you be up for making a GIF? For some reason quick time capture doesn't work for me on macOS Catalina 😢

sindresorhus: Sure, but let's wait until you've finished the validation logic, so I can show off the bouncing too when trying to increase to a higher number than allowed.

sunshinejr:
@sindresorhus added improved validation on text edit to have more room to breathe when changing the size. I also changed a little bit the behavior for arrows as I feel like these should have a stricter validation. Try it out and let me know.

sindresorhus:
After https://github.com/sindresorhus/Gifski/pull/83/commits/8b2bc5b034f152e69aa461fa58547e8d59f5cd0a, the sliders are no longer tabbable. If I focus the "Width" input, tab, it goes to "pixels", tab, and it goes to "New Folder". And it never ends up on the sliders.

sindresorhus:
The validation logic is now perfect. 👌

sunshinejr:
@sindresorhus great to hear. I think I fixed the tab focus order as well. With "sliders are no longer tappable" you meant the tab focus? Or you mean that when you change the slider value it should also focus the slider afterwards?

sindresorhus:
There are some tiny alignment and consistency tweaks needed in the UI:

Screen Shot 2019-06-22 at 20 25 58

sindresorhus:
I said ta**bb**le, not ta**pp**able.

sindresorhus:
> I think I fixed the tab focus order as well. With "sliders are no longer tappable" you meant the tab focus?

Yes. Confirmed fixed.

> Or you mean that when you change the slider value it should also focus the slider afterwards?

That would be nice, but not important. I wonder why it doesn't do that by default though.

sunshinejr:
@sindresorhus see updated UI (also added `focus` on sliders after action)

sindresorhus:
If I drag the slider and then press Esc, it no longer exits the save panel.

sunshinejr:
@sindresorhus ah shucks. Maybe this is why there is no focus by default 😅 I'll just remove the focus from sliders for now and we can tinker around it afterwards.

sindresorhus:
Yay!!!! This turned out really really great. Thanks for your perseverance.

1ab7e29c-117e-11e6-9cbb-f0828bacea8e

PR #84, opened 2019-05-17
Improve the error message alert panel

sindresorhus:
Currently, it looks like this:

Screen Shot 2019-05-17 at 18 42 13

But that's not very user-friendly.

I want it to look like this:

Screen Shot 2019-05-17 at 18 46 11

### DIFF
--- a/Gifski/util.swift
+++ b/Gifski/util.swift
@@ -306,16 +306,28 @@ extension NSAlert {
 				scrollView.translatesAutoresizingMaskIntoConstraints = false
 				scrollView.heightAnchor.constraint(equalToConstant: 120).isActive = true
 
+				// Alternative solution
+//				scrollView.frame = CGRect(width: 300, height: 120)
+//				scrollView.onAddedToSuperview {
+//					if let messageTextField = (scrollView.superview?.superview?.subviews.first { $0 is NSTextField }) {
+//						scrollView.frame.width = messageTextField.frame.width
+//					} else {
+//						assertionFailure("Couldn't detect the message textfield view of the NSAlert panel")
+//					}
+//				}
+
 				scrollView.onAddedToSuperview {
-					print("added")
-					scrollView.leftAnchor.constraint(equalTo: scrollView.superview!.leftAnchor).isActive = true
-					scrollView.rightAnchor.constraint(equalTo: scrollView.superview!.rightAnchor).isActive = true
+					if let messageTextField = (scrollView.superview?.superview?.subviews.first { $0 is NSTextField }) {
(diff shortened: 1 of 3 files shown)


Gifski/util.swift
```diff
@@ -266,6 +298,37 @@ extension NSAlert {
...
+			if #available(macOS 10.14, *) {
+				let scrollView = NSTextView.scrollableTextView()
+				//scrollView.frame = CGRect(width: 300, height: 120)
+				scrollView.translatesAutoresizingMaskIntoConstraints = false
+				scrollView.heightAnchor.constraint(equalToConstant: 120).isActive = true
+
+				scrollView.onAddedToSuperview {
+					print("added")
+					scrollView.leftAnchor.constraint(equalTo: scrollView.superview!.leftAnchor).isActive = true
+					scrollView.rightAnchor.constraint(equalTo: scrollView.superview!.rightAnchor).isActive = true
```
sindresorhus: I could use some help with the auto-layout constraints. I want to make the scroll view take up as much width as available without making the alert panel wider. The above just makes it 0 width...

// @sunshinejr

Gifski/util.swift
```diff
@@ -236,7 +236,36 @@ extension NSWindowController: NSWindowDelegate {
...
+			}
+
+			onAdded?()
+			removeFromSuperview()
+		}
+	}
+
+	// Enables you do add contraints and do other initialization without having to subclass the view.
+	// TODO: Please show me a better way to achieve this than using an invisible view 🙏
+	func onAddedToSuperview(_ closure: @escaping () -> Void) {
```
sindresorhus: Does anyone know a better way of setting up constraints to a superview without having to subclass the view and add those contraints in `loadView`?

Jon889: Can’t you add the constraints after you call `addSubview’

Gifski/util.swift
sunshinejr: Yeah _I think_ that the `scrollView.superview` doesn't really have a width and because of that the `scrollView` won't have as well (so the wrapper on top of `accessoryView` is growing based on `accessoryView`, not the other way around). You might have to add a `width` constraint to the `scrollView`, but not sure what should we based that on. There are two `NSTextField`s but this seems really hacky to base it on. Wonder if going with a frame "just-enought" wouldn't be okay for now.

sindresorhus: I tried that (https://github.com/sindresorhus/gifski-app/pull/84/commits/382946db9773d5f402743b968f0737de0d5de99a), but it still doesn't work...

I have a solution for doing it manually (commented out), but I would really prefer using auto-layout for this.

sindresorhus: I'm giving up on auto-layout for this...

PR #86, opened 2019-06-21
Show a notification when the conversion is completed

### LINKED ISSUE #82: Notification when Gifski completed the conversion
There are times that it takes time to convert the GIF, and having hidden dock I have to check few times if the conversion completed. It would be awesome to have a notification that would respect DND (similar to Xcode builds).

sindresorhus:
Fixes #82

The notification will only be shown if the app isn't active or if the app is active but the window is not visible.

Screen Shot 2019-06-22 at 02 07 25

// @sunshinejr

### DIFF
--- a/Gifski/MainWindowController.swift
+++ b/Gifski/MainWindowController.swift
@@ -1,5 +1,6 @@
 import Cocoa
 import AVFoundation
+import UserNotifications
 import StoreKit
 import Crashlytics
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
I noticed one weird thing. If the window is minimized (while Gifski is still the active app), the notification is sent by the code, but it goes directly to the notification center without being shown. That's very weird. @sunshinejr Can you reproduce that?

sindresorhus:
> I think this could be a well-thought design decision (or a bug nobody has noticed? 🤔).

Feels more like a bug than something well thought out. I'll open a Radar, uhm, I mean FB, when I get a chance.

PR #87, opened 2019-06-22
Improve the conversion completed view

sindresorhus:
The main motivation for this is that the existing buttons are not very legible when the window is vibrant. I think it also looks better to just use the native buttons. I also did some cleanup.

#### Before

Screen Shot 2019-06-22 at 14 36 51

#### After

Screen Shot 2019-06-22 at 14 38 27


sindresorhus:
I tried that out too. I kinda like the dim look better in this case though.

PR #88, opened 2019-06-22
Add some colors to the initial view

sindresorhus:
The initial view is a bit boring. This idea might be too much, but at least it could start a discussion.

I want to do something with the text too, but haven't decided what yet. Ideas welcome.

The color scheme is not optimal, ideally, I want to use the colors used in https://github.com/sindresorhus/Gifski/issues/25#issuecomment-377265077, but this will do for now.

And would be super cool if the lines would animate out when the conversion starts.

And if they would move a little bit when you drag in a file.

#### Before

Screen Shot 2019-06-22 at 15 02 26

#### After

Screen Shot 2019-06-22 at 15 00 38


kornelski:
Nice. These elements look like they could be animated (e.g. either open on startup, or close after you drop something).

sindresorhus:
@kornelski Yes, that would be cool.

Gifski/VideoDropView.swift
```diff
@@ -67,6 +67,11 @@ class DropView: SSView {
 	override func draw(_ dirtyRect: CGRect) {
 		super.draw(dirtyRect)
 
+		// We only draw it when the drop view controller is the main view controller.
```
sunshinejr: what do you think about extracting it to a `backgroundImage` property of a `VideoDropView` and then only pass it in the controllers that do not hide the text label as well?

sindresorhus: I agree that would be technically better. I'm just not sure it's worth the extra complexity. VideoDropViewController is pushed in many places. Would mean a lot of duplicate code for the same result. This is only a temporary thing anyway. I'll rewrite it for SwiftUI with animations later this year.

sunshinejr: yeah, agree there would be some extra cost for that - I was also thinking about 2 classes, one that would be invisible and one with text & background, but that would be an excellent one for SwiftUI component composition

Issue #78, opened 2019-05-05
New gifski library API


kornelski:
[<img alt="Issuehunt badges" src="https://issuehunt.io/r/sindresorhus/gifski-app/issues/78/badge.svg" />](https://issuehunt.io/r/sindresorhus/gifski-app/issues/78)

I've rewritten the C interface. It's now async/single-threaded, so you don't need a separate thread for writing and decoding video frames.

```c
gifski *g = gifski_new(&(GifskiSettings){});
gifski_set_file_output(g, "file.gif"); // this is now async

// it's OK to use the same thread
for(int i=0; i < frames; i++) {
     int res = gifski_add_frame_rgba(g, i, width, height, buffer, 5);
     if (res != GIFSKI_OK) break;
}

// this now waits
int res = gifski_finish(g);
if (res != GIFSKI_OK) return;
```

I've also added a new way of writing without a file path, so you can use Swift's I/O.

```c
gifski_set_write_callback(g, callback, user_data);
```

The callback is called many times, every time there's some fragment of file data ready to be written.

and the callback can be something like:

```c
GifskiError callback(size_t buffer_length, const uint8_t *buffer, void *user_data) {
   if (buffer_length > 0) {
      fwrite(buffer, buffer_length, 1, user_data);
   }
   return GIFSKI_OK;
}
```

> There is a $100.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/gifski-app/issues/78).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/gifski-app/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

sindresorhus:
Oh great! We will definitely want to handle I/O with Swift instead. Since nobody would convert a GIF larger than available memory, I'm thinking of just buffering up the whole file in memory and then doing one write to disk at the end. Thoughts? That would also prevent any issues with the destination disk becoming unavailable during conversion. I've seen that happen many times in Crashlytics.

Could you also upgrade the Xcode project in the ImageOptim/gifski repo to simplify pulling it in?

Any chance of changing https://github.com/ImageOptim/gifski/issues/25 while you're at it?

kornelski:
The API remained backwards-compatible, and you've already got quality scaling implemented, so I don't think it's worth changing the settings struct.

kornelski:
[...]

PR #90, opened 2019-06-22
Upgrade the gifski library

### LINKED ISSUE #51: Update gifski library
@kornelski Would you be able to update https://github.com/ImageOptim/gifski/tree/master/gifski.xcodeproj to latest Xcode and then update the gifski library in this repo? I noticed it has had a bunch of improvements since the last time we updated: https://github.com/ImageOptim/gifski/compare/94d6339d917b908a2eacbca6eb29599f71d8f96e...master

sindresorhus:
Fixes #51 
Fixes part of #78

This only upgrades and adapts the new API. It does not switch to handling our own file writing. That can come later.

### DIFF
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -7,8 +7,7 @@ final class Gifski {
 		case invalidSettings
 		case generateFrameFailed(Swift.Error)
 		case addFrameFailed(Swift.Error)
-		case startWritingFailed(Swift.Error)
-		case finishWritingFailed(Swift.Error)
+		case writeFailed(Swift.Error)
 		case cancelled
 
 		var errorDescription: String? {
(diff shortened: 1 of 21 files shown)


Gifski/GifskiWrapper.swift
```diff
@@ -58,10 +58,6 @@ final class GifskiWrapper {
 		self.pointer = pointer
 	}
 
-	deinit {
-		gifski_drop(pointer)
```
sindresorhus: @kornelski This was removed, so I guess there's no cleanup needed anymore? Should it do `pointer = nil` here?

Gifski/Gifski.swift
```diff
@@ -19,10 +19,10 @@ final class Gifski {
...
 			case let .addFrameFailed(error):
 				return "Failed to add frame, with underlying error: \(error.localizedDescription)"
-			case let .endAddingFramesFailed(error):
-				return "Failed to end adding frames, with underlying error: \(error.localizedDescription)"
-			case let .writeFailed(error):
-				return "Failed to write to output, with underlying error: \(error.localizedDescription)"
+			case let .startWritingFailed(error):
+				return "Failed to start writing, with underlying error: \(error.localizedDescription)"
+			case let .finishWritingFailed(error):
+				return "Failed to finish writing, with underlying error: \(error.localizedDescription)"
```
sindresorhus: @kornelski Are these error names correct?

Gifski/GifskiWrapper.swift
kornelski: Fine to remove. The cleanup is done by `gifski_finish()`

sindresorhus: But what if it encounters an error and then `gifski_finish()` is not called.

Gifski/Gifski.swift
kornelski: The user-facing message doesn't need to distinguish between start/finish explicitly. The `gifski_finish()` call will report any error during writing.

sindresorhus: Like this https://github.com/sindresorhus/Gifski/pull/90/commits/ec5905990312aceb079fcf522d0e603d821cf8d6 ?

PR #92, opened 2019-06-22
Handle our own file writing

### LINKED ISSUE #78: New gifski library API

sindresorhus:
Fixes #78

I also finally managed to nail down an old bug in Gifski where if you cancel right before it's finished, it would output an error.

### DIFF
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -27,40 +27,40 @@ final class Gifski {
 	}
 
 	struct Conversion {
-		let input: URL
-		let output: URL
+		let video: URL
 		let quality: Double
 		let dimensions: CGSize?
 		let frameRate: Int?
 
+		// TODO: With Swift 5.1 we can remove the manual `init` and have it synthesized.
 		/**
 		- Parameter frameRate: Clamped to 5...30. Uses the frame rate of `input` if not specified.
 		*/
-		init(input: URL, output: URL, quality: Double = 1, dimensions: CGSize? = nil, frameRate: Int? = nil) {
-			self.input = input
-			self.output = output
+		init(video: URL, quality: Double = 1, dimensions: CGSize? = nil, frameRate: Int? = nil) {
+			self.video = video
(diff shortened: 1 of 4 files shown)


Gifski/Gifski.swift
```diff
@@ -90,20 +83,38 @@ final class Gifski {
...
+				let bufferPointer = bufferPointer
+			else {
+				return 0
+			}
+
+			let data = context!.assumingMemoryBound(to: NSMutableData.self).pointee
+			data.append(bufferPointer, length: bufferLength)
+
+			return 0
+		}
```
sindresorhus: @kornelski I could use an extra set of eyes on this part.

kornelski: LGTM

PR #93, opened 2019-06-23
Add Gifski metadata to the generated GIF

### LINKED ISSUE #61: Add metadata?
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/gifski-app/issues/61)

Would it maybe be useful to add metadata that the GIF was created with Gifski?

Seems like we can use [`kCGImagePropertyIPTCOriginatingProgram`](https://developer.apple.com/documentation/imageio/kcgimagepropertyiptcoriginatingprogram) and [`kCGImagePropertyIPTCProgramVersion`](https://developer.apple.com/documentation/imageio/kcgimagepropertyiptcprogramversion).

Reference code: https://github.com/search?l=Objective-C&q=kCGImagePropertyIPTCOriginatingProgram&type=Code

---

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $60.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($60.00)
### Submitted pull Requests
- [#93 Add Gifski metadata to the generated GIF](https://issuehunt.io/r/sindresorhus/gifski-app/pull/93)
---

### Tips

---
</details>

sunshinejr:
Fixes #61.

So this is not exactly using the IPTC properties mentioned in the issue but one of the common metadata properties from [Apple's docs](https://developer.apple.com/documentation/coreservices/file_metadata/mditem/common_metadata_attribute_keys).

The reason is that GIF doesn't seem to support IPTC properties in the GIF itself, and when I tried to set these as extended attributes they weren't properly recognized as "real" IPTC properties (used `exiftool`/`mdls` for the check). But I found a nice extended attribute that macOS is also displaying on the sidebar in Finder (don't mind the polish language 😄):

Zrzut ekranu 2019-06-23 o 02 09 28

Let me know @sindresorhus @kornelski what do you think about it!

---
**Edit:** Also, there is an API for `MDItem` that could **fetch attributes** but for some reason there is no exposed function to **set attributes**. But, in fact, there is a runtime function for saving these properties and if we expose it in our objc header in Gifski with:
```objective-c
@import CoreServices;

Boolean MDItemSetAttribute(MDItemRef, CFStringRef name, CFTypeRef attr);
```
we have a "nice-looking" API to set these properties instead of using `setxattr`:
```swift
if let item = MDItemCreate(nil, url.path as CFString) {
	MDItemSetAttribute(item, kMDItemCreator, "Gifski 1.8.0 (17)" as CFTypeRef)
}
```
It also seems to be there forever but not sure how safe this is 😄

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#61: Add metadata?](https://issuehunt.io/repos/119822304/issues/61)
---

</details>

### DIFF
--- a/Gifski/MainWindowController.swift
+++ b/Gifski/MainWindowController.swift
@@ -310,9 +310,7 @@ final class MainWindowController: NSWindowController {
 	}
 
 	private func addGifskiMetadata(to url: URL) {
-		let app = "\"\(App.name) \(App.versionWithBuild)\""
-		let itemCreatorKey = kMDItemCreator as String
-		setxattr(url.path, "com.apple.metadata:\(itemCreatorKey)", app, strlen(app), 0, 0)
+		try? url.setMetadata(key: .itemCreator, value: "\(App.name) \(App.versionWithBuild)")
 	}
 }
 
--- a/Gifski/MainWindowController.swift
+++ b/Gifski/MainWindowController.swift
@@ -258,7 +258,7 @@ final class MainWindowController: NSWindowController {
 					self.presentError(error, modalFor: self.window)
 				}
 
-				try? inputUrl.setMetadata(key: .itemCreator, value: "\(App.name) \(App.versionWithBuild)")
+				try? outputUrl.setMetadata(key: .itemCreator, value: "\(App.name) \(App.versionWithBuild)")
 				self.progress?.unpublish()
 				self.isRunning = false
 
--- a/Gifski/MainWindowController.swift
+++ b/Gifski/MainWindowController.swift
@@ -258,7 +258,7 @@ final class MainWindowController: NSWindowController {
 					self.presentError(error, modalFor: self.window)
 				}
 
-				self.addGifskiMetadata(to: outputUrl)
+				try? url.setMetadata(key: .itemCreator, value: "\(App.name) \(App.versionWithBuild)")
 				self.progress?.unpublish()
 				self.isRunning = false
 
@@ -307,10 +307,6 @@ final class MainWindowController: NSWindowController {
 			timeRemainingLabel.centerXAnchor.constraint(equalTo: view.centerXAnchor),
 			timeRemainingLabel.topAnchor.constraint(equalTo: circularProgress.bottomAnchor)
 		])
-	}
-
-	private func addGifskiMetadata(to url: URL) {
-		try? url.setMetadata(key: .itemCreator, value: "\(App.name) \(App.versionWithBuild)")
 	}
 }
 
--- a/Gifski/MainWindowController.swift
+++ b/Gifski/MainWindowController.swift
@@ -258,7 +258,7 @@ final class MainWindowController: NSWindowController {
 					self.presentError(error, modalFor: self.window)
 				}
 
-				try? outputUrl.setMetadata(key: .itemCreator, value: "\(App.name) \(App.versionWithBuild)")
+				try? outputUrl.setMetadata(key: .itemCreator, value: "\(App.name) \(App.version)")
 				self.progress?.unpublish()
 				self.isRunning = false
 
--- a/Gifski/MainWindowController.swift
+++ b/Gifski/MainWindowController.swift
@@ -258,6 +258,7 @@ final class MainWindowController: NSWindowController {
 					self.presentError(error, modalFor: self.window)
 				}
 
+				self.addGifskiMetadata(to: outputUrl)
 				self.progress?.unpublish()
 				self.isRunning = false
 
@@ -306,6 +307,12 @@ final class MainWindowController: NSWindowController {
 			timeRemainingLabel.centerXAnchor.constraint(equalTo: view.centerXAnchor),
 			timeRemainingLabel.topAnchor.constraint(equalTo: circularProgress.bottomAnchor)
 		])
+	}
+
+	private func addGifskiMetadata(to url: URL) {
+		let app = "\"\(App.name) \(App.versionWithBuild)\""
+		let itemCreatorKey = kMDItemCreator as String
+		setxattr(url.path, "com.apple.metadata:\(itemCreatorKey)", app, strlen(app), 0, 0)
 	}
 }
 
--- a/Gifski/MainWindowController.swift
+++ b/Gifski/MainWindowController.swift
@@ -258,7 +258,7 @@ final class MainWindowController: NSWindowController {
 					self.presentError(error, modalFor: self.window)
 				}
 
-				try? url.setMetadata(key: .itemCreator, value: "\(App.name) \(App.versionWithBuild)")
+				try? inputUrl.setMetadata(key: .itemCreator, value: "\(App.name) \(App.versionWithBuild)")
 				self.progress?.unpublish()
 				self.isRunning = false
 
--- a/Gifski/util.swift
+++ b/Gifski/util.swift
@@ -2386,7 +2386,7 @@ extension NSControl {
 
 extension URL {
 	enum MetadataKey {
-		/// The app used to create the file, for example, "Word", "QuickTime", etc.
+		/// The app used to create the file, for example, `Gifski 2.0.0`, `QuickTime Player 10.5`, etc.
 		case itemCreator
 
 		var attributeKey: String {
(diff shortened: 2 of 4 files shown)


sindresorhus:
We cannot use private APIs as Gifski is on the Mac App Store.

---

Some more info about the method: https://github.com/adobe/chromium/blob/cfe5bf0b51b1f6b9fe239c2a3c2f2364da9967d7/content/browser/file_metadata_mac.mm#L40-L51

---

You should submit a feedback on Feedback Assistant to expose `MDItemSetAttribute`. It's weird that it's not exposed.

Gifski/MainWindowController.swift
```diff
@@ -307,6 +308,12 @@ final class MainWindowController: NSWindowController {
 			timeRemainingLabel.topAnchor.constraint(equalTo: circularProgress.bottomAnchor)
 		])
 	}
+
+	private func addGifskiMetadata(to url: URL) {
+		let app = "\"\(App.name) \(App.versionWithBuild)\""
+		let itemCreatorKey = kMDItemCreator as String
+		setxattr(url.path, "com.apple.metadata:\(itemCreatorKey)", app, strlen(app), 0, 0)
```
sindresorhus: I'm surprised this works. The first argument of `setxattr` is typed as `UnsafePointer<Int8>!`. Not sure if it's safe though.

You also need to handle POSIX errors.

sindresorhus:
I actually have a high-level `xattr` class in my private framework. We could use that:

```swift
extension POSIXError {
	/// Create an error from the global C `errno`.
	static var fromErrno: POSIXError {
		return self.init(errno: errno)
	}

	/**
	Create an error from the given C `errno`.

	```
	let length = getxattr(fileSystemPath, name, nil, 0, 0, 0)

	guard length >= 0 else {
		throw POSIXError(errno: errno)
	}
	```
	*/
	init(errno errorCode: Int32) {
		self.init(POSIXErrorCode(rawValue: errorCode) ?? .EPERM)
	}
}



final class ExtendedAttributes {
	let url: URL

	init(url: URL) {
		self.url = url
	}

	private func checkIfFileURL() throws {
		guard url.isFileURL else {
			throw CocoaError(.fileNoSuchFile)
		}
	}

	func has(_ name: String) -> Bool {
		guard url.isFileURL else {
			return false
		}

		return url.withUnsafeFileSystemRepresentation { fileSystemPath in
			getxattr(fileSystemPath, name, nil, 0, 0, 0) > 0
		}
	}

	func get(_ name: String) throws -> Data {
		try checkIfFileURL()

		return try url.withUnsafeFileSystemRepresentation { fileSystemPath in
			let length = getxattr(fileSystemPath, name, nil, 0, 0, 0)

			guard length >= 0 else {
				throw POSIXError.fromErrno
			}

			var data = Data(count: length)

			let result = data.withUnsafeMutableBytes {
				getxattr(fileSystemPath, name, $0.baseAddress, length, 0, 0)
			}

			guard result >= 0 else {
				throw POSIXError.fromErrno
			}

			return data
		}
	}

	/**
	- Note: Ensure you specify a type.

	```
	let isProtected: Bool = try? attributes.get("com.apple.rootless") ?? false
	```
	*/
	func get<T>(_ name: String) throws -> T {
		try checkIfFileURL()

		let data = try get(name)

		let value = try PropertyListSerialization.propertyList(from: data, options: [], format: nil)

		guard let result = value as? T else {
			throw CocoaError(.propertyListReadCorrupt)
		}

		return result
	}

	func set(_ name: String, data: Data) throws {
		try checkIfFileURL()

		try url.withUnsafeFileSystemRepresentation { fileSystemPath in
			let result = data.withUnsafeBytes {
				setxattr(fileSystemPath, name, $0.baseAddress, data.count, 0, 0)
			}

			guard result >= 0 else {
				throw POSIXError.fromErrno
			}
		}
	}

	func set<T>(_ name: String, value: T) throws {
		try checkIfFileURL()

		guard PropertyListSerialization.propertyList(value, isValidFor: .binary) else {
			throw CocoaError(.propertyListWriteInvalid)
		}

		let data = try PropertyListSerialization.data(fromPropertyList: value, format: .binary, options: 0)
		try set(name, data: data)
	}

	func remove(_ name: String) throws {
		try checkIfFileURL()

		try url.withUnsafeFileSystemRepresentation { fileSystemPath in
			guard removexattr(fileSystemPath, name, 0) >= 0 else {
				throw POSIXError.fromErrno
			}
		}
	}

	/// Get all the extended attribute names
	func all() throws -> [String] {
		try checkIfFileURL()

		let list: [String] = try url.withUnsafeFileSystemRepresentation { fileSystemPath in
			let length = listxattr(fileSystemPath, nil, 0, 0)

			guard length >= 0 else {
				throw POSIXError.fromErrno
			}

			var data = Data(count: length)

			let result = data.withUnsafeMutableBytes {
				listxattr(fileSystemPath, UnsafeMutablePointer<Int8>(OpaquePointer($0.baseAddress)), length, 0)
			}

			guard result >= 0 else {
				throw POSIXError.fromErrno
			}

			let list = data.split(separator: 0).compactMap {
				String(data: Data($0), encoding: .utf8)
			}

			return list
		}

		return list
	}

	func debug() {
		print("Extended attributes:\n\(try! all().joined(separator: .newline))")
	}
}
```

Gifski/MainWindowController.swift
```diff
@@ -307,6 +308,12 @@ final class MainWindowController: NSWindowController {
 			timeRemainingLabel.topAnchor.constraint(equalTo: circularProgress.bottomAnchor)
 		])
 	}
+
+	private func addGifskiMetadata(to url: URL) {
```
sindresorhus: Maybe we could make this more general:

```swift
extension URL {
	struct MetadataKey {
		// ...
	}

	func setMetadata<T>(key: MetadataKey, value: T) {

	}
}

url.setMetadata(key: .itemCreator, value: "\(App.name) \(App.versionWithBuild)")
```

sunshinejr:
@sindresorhus yeah I submitted feedback, let's see if that helps. Also updated the PR with your `ExtendedAttributes` class (it's awesome btw) + an extension on `URL`. Currently implemented what we actually use but we can iterate on it further of course.

Gifski/util.swift
```diff
@@ -2383,3 +2383,21 @@ extension NSControl {
...
+		var attributeKey: String {
+			switch self {
+			case .itemCreator:
+				return kMDItemCreator as String
+			}
+		}
+	}
+
+	func setMetadata<T>(key: MetadataKey, value: T) throws {
+		let attributes = ExtendedAttributes(url: self)
```
sindresorhus: This is not needed: https://github.com/sindresorhus/Gifski/pull/93/files/ccf4101d69900875455a2cc47b2a4f3f5d87de1b..b1edcb12d7bae8acaefb105a49dffe2572d04995#diff-bf8c6e8ce59f7cf7b4011ef15b26bf7dR166

Gifski/MainWindowController.swift
```diff
@@ -307,6 +308,10 @@ final class MainWindowController: NSWindowController {
...
```
sindresorhus: I feel like this method is moot now and we can just set the metadata inline.

Gifski/MainWindowController.swift
```diff
@@ -258,7 +258,7 @@ final class MainWindowController: NSWindowController {
 					self.presentError(error, modalFor: self.window)
 				}
 
-				try? url.setMetadata(key: .itemCreator, value: "\(App.name) \(App.versionWithBuild)")
+				try? inputUrl.setMetadata(key: .itemCreator, value: "\(App.name) \(App.versionWithBuild)")
```
sindresorhus: outputUrl

sunshinejr: yeah lol just pushed the update

Gifski/MainWindowController.swift
```diff
@@ -258,6 +258,7 @@ final class MainWindowController: NSWindowController {
 					self.presentError(error, modalFor: self.window)
 				}
 
+				try? outputUrl.setMetadata(key: .itemCreator, value: "\(App.name) \(App.versionWithBuild)")
```
sindresorhus: Been thinking about it. I don't think we should include the build number. It's not useful for users.
```suggestion
				try? outputUrl.setMetadata(key: .itemCreator, value: "\(App.name) \(App.version)")
```

PR #94, opened 2019-06-23
Generate GIF without asking for location

### LINKED ISSUE #72: Generate GIF without asking for location
# Proposed interaction

1. User: drag-drops a video file onto the Gifski app window
2. App: immediately generates gif in a temp (cache) path and displays percent-complete progress
3. App: once GIF is created, a draggable preview thumbnail is displayed alongside 4 action buttons:
    - **Show in Finder**: same as today
    - **Share**: same as today
    - **Move to**: opens dialog for user to choose a destination folder and filename
    - **Copy**: copies GIF data to clipboard to paste in another app
4. User: drags preview thumbnail into another app

## What it does today
Upon drag-dropping a video file on the Gifski app, instead of starting the conversion, it immediately asks where the generated GIF should be saved.

## Preference: Always Save to Folder
For users who always want generated GIFs to be stored in the same folder, this could be done by adding an "Always Save to Folder" preference that defaults to the cache location

# Why?

I use GIFs not as a source-of-truth format, but rather as a ubiquitous, easy-to-share format to capture a few seconds of something I made or something I like. A throw-away format that only exists for sharing.

Examples:
- Take a screen recording of a prototype I made, gif it, and then drag-drop it into a chat window to my team.
- Take a screen recording of an interaction I really like in some app, gif it, and then drag-drop it into a Google Slides presentation.

So in that sense, they're like screenshots: likely not needed after they've been shared. Kinda like how macOS doesn't ask you where to save a screenshot/recording, it either just saves it, or asks you where to save it *after* it's been created.

## More context
[Twitter convo](https://twitter.com/sindresorhus/status/1110945126262534144?s=12)

sunshinejr:
Fixes #72. This needs us to rebuild our current architecture to use controllers as well so it will be another big PR.


sindresorhus:
Keep in mind that all the stages needs to be drop targets. The user should be able to drop a video at the initial view, when showing the settings, and when the conversion is completed.

sindresorhus:
Some quick things I noticed. Feel free to ignore them for now as they're nitpick. Just wanted to write it down while I remember.

- The first time you drag and drop, there's no blue highlight around the window contents.
- If I open a file through the "Open" menu item, there's no animated transition to the larger window.
- If a drag in a file, then press Esc, and drag in the file again, the "Convert" button is no longer highlighted (accent color).

Gifski/DropVideoViewController.swift
```diff
@@ -0,0 +1,113 @@
+import AppKit
+import Crashlytics
+import AVKit
+
+final class DropVideoViewController: NSViewController {
+	private lazy var videoDropView = with(VideoDropView()) {
+		$0.frame.size = CGSize(width: 360, height: 240)
```
sindresorhus: I don't think this should be hard-coded.

Gifski/DropVideoViewController.swift
```diff
@@ -0,0 +1,113 @@
...
+		$0.dropText = "Drop a Video to Convert to GIF"
+		$0.onComplete = { [weak self] url in
+			NSApp.activate(ignoringOtherApps: true)
+			self?.convert(url.first!)
```
sindresorhus: `url.first` reads weird. I think we should change the drop view to only return a single URL.

Gifski/DropVideoViewController.swift
```diff
@@ -0,0 +1,113 @@
...
+			else {
+				NSAlert.showModalAndReportToCrashlytics(
+					for: view.window,
+					message: "The video dimensions must be at least 10×10.",
+					informativeText: "The dimensions of your video are \(asset.dimensions?.formatted ?? "0×0").\n\nIf you think this error is a mistake, please open an issue on https://github.com/sindresorhus/Gifski or email sindresorhus@gmail.com. ZIP the video and attach it.\n\nInclude this info:",
+					debugInfo: asset.debugInfo
+				)
+
+				return
+		}
```
sindresorhus: Can you move all these checks into a separate method? `validateInput()` or something.

Gifski/DropVideoViewController.swift
```diff
@@ -0,0 +1,113 @@
...
+			)
+
+			return
+		}
+
+		guard
+			let dimensions = asset.dimensions,
+			dimensions.width > 10,
+			dimensions.height > 10
+			else {
```
sindresorhus:
```suggestion
		else {
```

Gifski/MainWindowController.swift
```diff
@@ -296,25 +46,23 @@ final class MainWindowController: NSWindowController {
...
-			return
+	func convert(_ inputUrl: URL) {
+		if let dropVideo = window?.contentViewController as? DropVideoViewController {
+			dropVideo.convert(inputUrl)
+		} else if !(window?.contentViewController is ConversionViewController) {
+			let dropVideo = DropVideoViewController()
+			window?.contentViewController?.push(viewController: dropVideo) {
+				dropVideo.convert(inputUrl)
+			}
 		}
```
sindresorhus: Is there really no better way of doing this?

Gifski/util.swift
```diff
@@ -2417,3 +2418,45 @@ extension URL {
...
+extension NSViewController {
+	func push(viewController: NSViewController, completion: (() -> Void)? = nil) {
+		guard let window = view.window else {
+			return
+		}
+
+		let newOrigin = CGPoint(x: window.frame.midX - viewController.view.frame.width / 2.0, y: window.frame.midY - viewController.view.frame.height / 2.0)
+		let newWindowFrame = CGRect(origin: newOrigin, size: viewController.view.frame.size)
+
+		let animation = CABasicAnimation()
```
sindresorhus: Would be cool to use CASpringAnimation to make the transition even more organic.

Gifski/util.swift
```diff
@@ -2417,3 +2418,45 @@ extension URL {
...
+
+	func add(childController: NSViewController) {
+		addChild(childController)
+		view.addSubview(childController.view)
+		childController.view.constrainEdgesToSuperview()
+	}
+}
+
+extension URL {
+	static func generateTempGifUrl(for videoUrl: URL) -> URL {
```
sindresorhus: I don't think this should be an extension on URL. It's too specific to this use-case.

Gifski/util.swift
```diff
@@ -2417,3 +2418,45 @@ extension URL {
...
+		let tempDirectory = FileManager.default.temporaryDirectory
+		let tempName = videoUrl.filenameWithoutExtension + ".\(FileType.gif.fileExtension)"
```
sindresorhus:
```suggestion
		let tempName = "\(videoUrl.filenameWithoutExtension).\(FileType.gif.fileExtension)"
```

sindresorhus:
- There's no longer any animation on the thumbnail when showing the conversion completed view.
- What was the reason for switching to a XIB for the conversion completed view?

Gifski/AppDelegate.swift
```diff
@@ -26,13 +26,13 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
-		if urlsToConvertOnLaunch != nil {
-			mainWindowController.convert(urlsToConvertOnLaunch)
+		if urlToConvertOnLaunch != nil {
+			mainWindowController.convert(urlToConvertOnLaunch)
 		}
 	}
 
 	func application(_ application: NSApplication, open urls: [URL]) {
-		guard !mainWindowController.isRunning else {
+		guard !(mainWindowController.window?.contentViewController is ConversionViewController) else {
```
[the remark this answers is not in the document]
sunshinejr: yes definitely will replace this line, was just a placeholder

Gifski/DropVideoViewController.swift
sunshinejr: Somehow we need to specify a window size for the controller. I cannot use the size of a current window as this might be e.g. Edit controller that has a different size than most of the controllers. Maybe we would need a constant size for all the controllers except the Edit one?

sindresorhus: Yes, I would prefer defining the size in the view controller view instead of here.

Gifski/MainWindowController.swift
sunshinejr: There must be, we don't need the drop video controller here 😄 Will be fixed.

Gifski/Constants.swift
```diff
@@ -8,3 +8,7 @@ extension Defaults.Keys {
 	static let outputQuality = Key<Double>("outputQuality", default: 1)
 	static let successfulConversionsCount = Key<Int>("successfulConversionsCount", default: 0)
 }
+
+extension CGSize {
+	static let defaultWindowSize = CGSize(width: 360, height: 240)
+}
```
sindresorhus: This would make more sense as `NSWindow.defaultWindowSize`.

sindresorhus: Actually, `Constants` would be better, since it's not a generic value.

sunshinejr:
@sindresorhus so I've made it a `.xib` as it was really hard to read the logic in there - lots of views, lots of constraints.

Gifski/VideoUrlValidator.swift
```diff
@@ -0,0 +1,99 @@
+import Crashlytics
+import AVKit
+
+struct VideoUrlValidator {
+	enum Result {
+		case failure
+		case success(AVURLAsset.VideoMetadata)
+	}
```
sindresorhus: Why Result for this? I usually only use Result for async operations. `throws` would be more fitting, but since there are no error message, I would just make the return value optional.

Gifski/VideoUrlValidator.swift
```diff
@@ -0,0 +1,99 @@
+import Crashlytics
+import AVKit
+
+struct VideoUrlValidator {
```
sindresorhus: It's not just validating the URL, so this name is a bit misleading. `VideoValidator`? Does it even need to be wrapped in a struct?

Gifski/VideoUrlValidator.swift
sunshinejr: So I want to validate the video but also return `videoMetadata`. The result of the `validate` operation should still be kind of a `Boolean` and so I'm using `videoMetadata` as an associated value to the success/failure.

Gifski/VideoUrlValidator.swift
sunshinejr: 👍 good call on the name - do you have any other idea than using a struct for that? I wanted to be able to extract the logic for the validation to a separate place

Gifski/VideoUrlValidator.swift
sindresorhus: Ok. I see your point of wanting to make it a boolean. By making it an optional, I meant that I would just rename it to `getVideoMetadata` or something which also validates, but your solution is cleaner, so let's keep it as is.

Gifski/VideoUrlValidator.swift
sindresorhus: Nah, it's fine.

Gifski/VideoDropViewController.swift
```diff
@@ -1,7 +1,7 @@
 import AppKit
 
 final class VideoDropViewController: NSViewController {
-	private let urlValidator = VideoUrlValidator()
+	private let urlValidator = VideoValidator()
```
sindresorhus: `urlValidator` => `videoValidator`

sunshinejr:
I've brought back the thumbnail animation. Also, @sindresorhus, there are the issues you noticed:
- The first time you drag and drop, there's no blue highlight around the window contents.
- If a drag in a file, then press Esc, and drag in the file again, the "Convert" button is no longer highlighted (accent color).

Do you have an idea why would that happen? Seems strange.

Also, I will open the PR for a review now, spent many hours on this one and a fresh pair of eyes would be lovely.

sindresorhus:
> The first time you drag and drop, there's no blue highlight around the window contents.

I figured this one out. It's caused by: https://github.com/sindresorhus/Gifski/blob/8cd24a5184a8bcef9e1555c367ad64ac8dc00799/Gifski/MainWindowController.swift#L84

> If a drag in a file, then press Esc, and drag in the file again, the "Convert" button is no longer highlighted (accent color).

This one is probably about first responder. When you press Esc you resign first responder, so the button is no longer blue. Maybe we should have the view controller be first responder when shown.

sunshinejr:
@sindresorhus when the controller is pushed [I'm making it the first responder by default](https://github.com/sindresorhus/Gifski/pull/94/files#diff-75a4e71e51a82ac8607443b8f202430eR2443). Also both `ESC` and tapping on `Cancel` button go trouch the `cancel` method that's why it's interesting to me what else does `ESC` do that makes the blue button resign its default state.

sunshinejr:
@sindresorhus figured it out. I added a `key equivalent` to the Cancel button and it got in that weird state. Cancel is called anyways on `ESC` click so when I removed it it all works correctly again.

sunshinejr:
With the vibrancy thing, any hack to make it work without deleting the line?

Gifski/ConversionViewController.swift
```diff
@@ -0,0 +1,124 @@
...
+		let conversionCompleted = ConversionCompletedViewController(conversion: conversion, gifUrl: gifUrl)
+		stopConversion {
+			self.push(viewController: conversionCompleted)
+		}
+	}
+
+	private func stopConversion(_ completion: (() -> Void)? = nil) {
+		isRunning = false
+		progress?.cancel()
+		progress?.unpublish()
```
sindresorhus: To fix the crash when canceling, you need to do this in a next tick. The reason is that after `.cancel()`, the libgifski side needs to cancel too. Otherwise they will continue sending their last update, but `progress` will no longer exist and it will crash.
```suggestion

		DispatchQueue.main.async {
			self.progress?.unpublish()
		}

```

sindresorhus:
> With the vibrancy thing, any hack to make it work without deleting the line?

I think you could use `view.insertVibrancyView(material: .underWindowBackground)` on the view controllers instead.

Gifski/ConversionViewController.swift
sunshinejr: Hmm, tried that and still got the crash sometimes 🤔

sindresorhus:
I noticed now that the progress circle get's less opacity/color when it shows the checkmark (before fading out). I don't think it was like this before.

sindresorhus:
For the conversion completed view, the buttons fade in must earlier and faster than the thumbnail. In the current Gifski version, it's better synced.

Gifski/ConversionCompletedViewController.swift
```diff
@@ -0,0 +1,159 @@
...
+			notification.title = "Conversion Completed"
+			notification.subtitle = conversion.video.filename
+			let request = UNNotificationRequest(identifier: "conversionCompleted", content: notification, trigger: nil)
+			UNUserNotificationCenter.current().add(request)
+		}
+
+		draggableFile.layer?.animateScaleMove(fromScale: 3, fromY: view.frame.height + draggableFile.frame.size.height)
+	}
+
+	private func setupUI() {
```
sindresorhus:
```suggestion
	private func setUpUI() {
```

Gifski/ConversionCompletedViewController.swift
```diff
@@ -0,0 +1,159 @@
...
+			panel.canCreateDirectories = true
+			panel.allowedFileTypes = [FileType.gif.identifier]
+			panel.directoryURL = inputUrl.directoryURL
+			panel.nameFieldStringValue = inputUrl.filenameWithoutExtension
+			panel.message = "Choose where to save the GIF"
+			panel.beginSheetModal(for: self.view.window!) { _ in }
+		}
+	}
+
+	private func setupDropView() {
```
sindresorhus:
```suggestion
	private func setUpDropView() {
```

Gifski/ConversionViewController.swift
```diff
@@ -0,0 +1,124 @@
...
+	private var isRunning = false
+
+	convenience init(conversion: Gifski.Conversion) {
+		self.init()
+
+		self.conversion = conversion
+	}
+
+	override func loadView() {
+		let wrapper = NSView(frame: CGRect(origin: .zero, size: Constants.defaultWindowSize))
```
sindresorhus:
```suggestion
		let wrapper = NSView(frame: Constants.defaultWindowSize.cgRect)
```

Gifski/ConversionViewController.swift
sindresorhus: What kind of crash though? Maybe it's a different one. Can you share?

Gifski/ConversionCompletedViewController.swift
sunshinejr: hmm, in the app the convention is always `setup` rather than `setUp` - do you prefer that I change all the places to `setUp` instead?

sunshinejr:
Hmm, I thought that I fixed the vibrancy by resequencing the lines but it seems that the bug is different  than what we thought - the vibrancy view makes the focus on itself and thus we don't see the blue highlight. It is visible afterwards because the next VC overlaps the vibrancy view and thus there is no vibrancy anymore but we can see the blue highlight.

So I think we should somehow stop the vibrancy view from stealing the focus (perhaps `canBecomeKeyView`?) and then on each VC we should add the vibrancy view probably.

Gifski/ConversionCompletedViewController.swift
sindresorhus: Because alone it can be used as a noun `setup`, but when you have `UI` afterwards it needs to be a verb `set up`. It could be `uiSetup`.

https://writingexplained.org/setup-vs-set-up-difference

sindresorhus:
> So I think we should somehow stop the vibrancy view from stealing the focus (perhaps canBecomeKeyView?) and then on each VC we should add the vibrancy view probably.

Sounds correct in theory, yes. But you never know with Cocoa.

Gifski/ConversionViewController.swift
sunshinejr: Zrzut ekranu 2019-07-8 o 14 57 15

sindresorhus:
Nice! I had no idea NSWindow has a built-in vibrancy view.

sindresorhus:
I noticed another issue. When it's done converting, the checkmark is drawn above the text `100%`:

Screen Shot 2019-07-08 at 23 51 27

sunshinejr:
@sindresorhus yeah I also noticed it and fixed it. Will update the PR shortly. Currently trying to figure out why my memory debugger says we have 70+ leaks currently, even when I replace `window.contentView` to be a simple `NSView` instead of using any of the created controllers at all. I fixed all retain cycles I found (didn't push the commit yet) but still got this weird state and I'm trying to figure out what's going on.

sunshinejr:
Okay, it seemed to be a weird Xcode state, I can't reproduce the issue anymore and I don't see any leaks/cycles. Got one issue in the debugger though:

> objc[66756]: __weak variable at 0x102705e30 holds 0x102700190 instead of 0x100f2c120. This is probably incorrect use of objc_storeWeak() and objc_loadWeak(). Break on objc_weak_error to debug.

sunshinejr:
Did some more investigation and it seems to be connected to the transition util? When I changed:
```swift
NSAnimationContext.runAnimationGroup({ _ in
	window.contentViewController?.view.animator().alphaValue = 0.0
	window.animator().setFrame(newWindowFrame, display: true)
}, completionHandler: {
	window.contentViewController = viewController
	viewController.view.animator().alphaValue = 1.0
	window.makeFirstResponder(viewController)
	completion?()
})
```

to 
```swift
window.contentViewController?.view.animator().alphaValue = 0.0
window.animator().setFrame(newWindowFrame, display: true)
window.contentViewController = viewController
viewController.view.animator().alphaValue = 1.0
window.makeFirstResponder(viewController)
completion?()
```
then the warning doesn't happen... I did try to debug on given symbol `objc_weak_error` but it doesn't say much to me...

sunshinejr:
False alarm, when removed the context it appeared anyways e.g. in the conversion completed view.

sunshinejr:
@sindresorhus I give up on the `objc_weak_error` in the console, tested the app for many days now and it doesn't seem to affect anything. I removed `unpublish` and I can't seem to reproduce the crash either.

I also removed some relict Save Panel hacks, fixed a bug when you selected "percent" based gif size and added focus on dimensions drop down in the Edit VC. Please take a look and see what we're missing.

sindresorhus:
> I give up on the objc_weak_error in the console, tested the app for many days now and it doesn't seem to affect anything.

Yeah, doesn't matter if it has no effect on anything. I have a feeling it's some kind of Swift bug.

> I removed unpublish and I can't seem to reproduce the crash either.

👌

sindresorhus:
There's still

> I noticed now that the progress circle get's less opacity/color when it shows the checkmark (before fading out). I don't think it was like this before.

and

> For the conversion completed view, the buttons fade in must earlier and faster than the thumbnail. In the current Gifski version, it's better synced.

sindresorhus:
Quick Look is no longer working. Nothing happens when I press Space and if I select "Quick Look" in the menu, I get:

Screen Shot 2019-07-18 at 17 10 47

sunshinejr:
@sindresorhus oh interesting, do you always have this problem with QuickLook? I cannot reproduce it on my videos 🤔

sunshinejr:
Hmm, actually, I can still reproduce the crash - after a successful GIF generation, I try another one and Cancel at 4% and it always crashes. Gonna see if I can figure out why is this happening.

sunshinejr:
Fixed the cancel crash (I think)! So we were double cancelling and depending if the second cancel was fast enough or not, the data was retained or not, and then crashed or not 😄 

We were double cancelling because cancelling of the progress cancelled the generator -> thus Gifski reported cancelled state and we then cancelled again... I could've skipped the cancelling on Gifski response.cancelled, but I just added a guard to cancel only if it's not cancelled yet, because I'm not sure if there are other cases when we could get cancel event from Gifski.

sindresorhus:
Nice catch!

sindresorhus:
I noticed one other small problem. One the conversion completed view, if I Command+H to hide the window and the show the window again, the thumbnail animation plays each time. It should only animate in when the view is presented, not when the window is made visible.

sunshinejr:
@sindresorhus 
- Conversion completed view should have better animation sync
- the bug with second animation on minimize should be fixed
- the bug with progress circle opacity should be fixed as well

I can't reproduce the issue with QuickLook so no idea how to approach this one...

sindresorhus:
Looks great now. Super nice work, @sunshinejr! Gifski is so much more usable now. 🙌

PR #101, opened 2019-07-28
Add ability to trim the video

### LINKED ISSUE #74: Support trimming
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/gifski-app/issues/74)

I realized that there's [`AVPlayerView`](https://developer.apple.com/documentation/avkit/avplayerview) and `AVPlayerView#beginTrimming` which makes it easy to create a trimming view.

See: https://developer.apple.com/documentation/avkit/avplayerview/implementing_trimming_macos

We could have a `Trim…` button in the save dialog.

---

---

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $200.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($200.00)
### Submitted pull Requests
- [#101 Implement trimming view](https://issuehunt.io/r/sindresorhus/gifski-app/pull/101)
---

### Tips

---
</details>

sunshinejr:
Fixes #74.
Zrzut ekranu 2019-07-28 o 18 30 35

How it currently works:
1. Drag a video.
2. If you want to trim a video (and the video can be trimmed) use the trimming slider and set the start and end time.
3. Now tap trim - the estimated duration will change and you are good to go! You can tap convert to finish the conversion.
4. If you changed your mind, tap Cancel - this will reset the trimming slider and you can trim it again with correct timings or just ignore the trimming and convert without it.

I think this is not ideal as you need to tap "Trim" to trim the video, ideally I would love to only have only the slider (without Trim/Cancel buttons) and once I'm finished changing the timings, just tap convert and boom. But unless there is a property that I'm missing, we can't just leave the slider so we would have to implement it ourselves which is not trivial at all.

Additionally, there are buttons in two places now which might be confusing... Unfortunately the API is really limiting in what we can and can't do with the buttons and slider.

@sindresorhus please let me know your thoughts on that!

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#74: Support trimming](https://issuehunt.io/repos/119822304/issues/74)
---

</details>


sindresorhus:
> Additionally, there are buttons in two places now which might be confusing... Unfortunately the API is really limiting in what we can and can't do with the buttons and slider.

I agree. It would be much nicer without the Trim/Cancel buttons. You could always do some good old view hierarchy searching. Search through the view hierarchy for the Trim/Cancel button, save a reference to them, hide them, and then synthetically click the Trim button when the user clicks Convert.

Screen Shot 2019-07-30 at 11 39 08

I usually use a method like this:

```swift
/// Get a subview matching a condition
func firstSubview(where matches: (NSView) -> Bool, deep: Bool = false) -> NSView? {
	for subview in subviews {
		if matches(subview) {
			return subview
		}

		if deep, let match = subview.firstSubview(where: matches) {
			return match
		}
	}

	return nil
}
```

Don't search by button title as it will be different in different locales, so I would search for `AVTrimView`, superview, and then find the buttons.

sindresorhus:
62010029-cbc4db00-b165-11e9-8d0c-2b496ecc8008

sindresorhus:
What do you think about having the adjustment controls below the trim view? I think those are more commonly used, so should be more easily reachable.

Gifski/EditVideoViewController.swift
```diff
@@ -278,6 +287,38 @@ final class EditVideoViewController: NSViewController {
...
+					self.startTime = startTime.seconds
+					self.endTime = endTime.seconds
+					self.estimateFileSize()
+				}
+			case .cancelButton:
+				// Recreating the player resets the start/end time on trimming view, couldn't find a better way to do it
+				self.playerView.player = AVPlayer(playerItem: AVPlayerItem(asset: self.asset))
+				self.startTime = nil
+				self.endTime = nil
+				self.estimateFileSize()
```
sindresorhus:
```suggestion
				self.estimateFileSize()
			@unknown default:
				assertionFailure("AVPlayerViewTrimResult received a new enum case. Please handle it.")
```

Gifski/Gifski.swift
```diff
@@ -28,6 +28,8 @@ final class Gifski {
 
 	struct Conversion {
 		let video: URL
+		let startTime: Double?
+		let endTime: Double?
```
sindresorhus: Instead of two separate variables, I would just use a closed range.

Gifski/EditVideoViewController.swift
```diff
@@ -278,6 +287,38 @@ final class EditVideoViewController: NSViewController {
...
+
+	private func beginTrimming() {
+		playerView.beginTrimming { [weak self] result in
+			guard let self = self else {
+				return
+			}
+
+			switch result {
+			case .okButton:
+				if let startTime = self.playerView.player?.currentItem?.reversePlaybackEndTime, let endTime = self.playerView.player?.currentItem?.forwardPlaybackEndTime {
```
sindresorhus:
```suggestion
				if
					let startTime = self.playerView.player?.currentItem?.reversePlaybackEndTime,
					let endTime = self.playerView.player?.currentItem?.forwardPlaybackEndTime
				{
```

Gifski/EditVideoViewController.swift
```diff
@@ -278,6 +287,38 @@ final class EditVideoViewController: NSViewController {
 		add(childController: videoDropController)
 	}
 
+	private func setupTrimmingView() {
```
sindresorhus:
```suggestion
	private func setUpTrimmingView() {
```

Gifski/EditVideoViewController.swift
sindresorhus: I use a simple util for this in my code:

```swift
struct Util {
	/// Warn about a new enum case when using "@unknown default".
	static func warnAboutUnknownDefaultInDebug(file: StaticString = #file, line: UInt = #line) {
		assertionFailure("Received a new enum case. Please handle it.", file: file, line: line)
	}
}
```

sindresorhus:
Tabbing is not working correctly. If I tab from the top controls, it then goes to the trimmer view, and then just repeats there forever and never ends up at the "Convert" button.

sunshinejr:
@sindresorhus I've added some hacks to remove these buttons, stretch the trim view and observe duration changes so we can get the time range. Also fixed stuff you talked about. Ready for another review. Really interested if these hacks work on Mojave.

Gifski/TrimmingAVPlayerView.swift
```diff
@@ -0,0 +1,47 @@
+import AVKit
+
+final class TrimmingAVPlayerView: AVPlayerView {
+	private var timeRangeObserver: NSKeyValueObservation?
+
+	func observeTrimmedTimeRange(_ updateClosure: @escaping (ClosedRange<Double>) -> Void) {
+		timeRangeObserver = player?.currentItem?.observe(\.duration, options: [.new]) { item, _ in
```
sindresorhus: `.duration` doesn't work on Mojave. It's never called, but `.forwardPlaybackEndTime` works fine. It's triggered when both changing the start and end trim. Can you check if that works on Catalina? If so, just add a note to switch to `.duration` when we target macOS 10.15.

Gifski/TrimmingAVPlayerView.swift
```diff
@@ -0,0 +1,47 @@
...
+				updateClosure(startTime...endTime)
+			}
+		}
+	}
+
+	func hideTrimButtons() {
+		// This method is a collection of hacks, so might be acting funky on different OS versions
+		let avTrimView = firstSubview(where: { $0.description.contains("AVTrimView") }, deep: true)
+
+		if let avTrimView = avTrimView {
```
sindresorhus: Use `guard` to reduce nesting. `let superview = avTrimView.superview` can be in the guard too.

Gifski/util.swift
```diff
@@ -2455,3 +2455,32 @@ extension NSViewController {
...
+				return match
+			}
+		}
+
+		return nil
+	}
+}
+
+extension NSLayoutConstraint {
+	func changing(firstItem: Any? = nil, firstAttribute: Attribute? = nil, relation: Relation? = nil, secondItem: NSView? = nil, secondAttribute: Attribute? = nil, multiplier: Double? = nil, constant: CGFloat? = nil) -> NSLayoutConstraint {
```
sindresorhus: Can you add a doc comment?

Gifski/EditVideoViewController.swift
```diff
@@ -273,11 +282,29 @@ final class EditVideoViewController: NSViewController {
...
 		let videoDropController = VideoDropViewController(dropLabelIsHidden: true)
 		add(childController: videoDropController)
 	}
 
+	private func setUpTrimmingView() {
+		trimmingObserver = playerView.observe(\.canBeginTrimming, options: .new) { [weak self] _, change in
+			if let canBeginTrimming = change.newValue, canBeginTrimming {
+				self?.beginTrimming()
+			}
+		}
```
sindresorhus: Maybe this can be moved to `TrimmingAVPlayerView`?

Gifski/TrimmingAVPlayerView.swift
```diff
@@ -0,0 +1,47 @@
...
+		timeRangeObserver = player?.currentItem?.observe(\.duration, options: [.new]) { item, _ in
+			let startTime = item.reversePlaybackEndTime.seconds
+			let endTime = item.forwardPlaybackEndTime.seconds
+			if !startTime.isNaN && !endTime.isNaN {
+				updateClosure(startTime...endTime)
+			}
+		}
+	}
+
+	func hideTrimButtons() {
```
sindresorhus: I would make this method private and call it inside `func viewDidMoveToSuperview() {}` in this class. Making it more self-contained is nice.

Gifski/TrimmingAVPlayerView.swift
```diff
@@ -0,0 +1,47 @@
+import AVKit
+
+final class TrimmingAVPlayerView: AVPlayerView {
+	private var timeRangeObserver: NSKeyValueObservation?
+
+	func observeTrimmedTimeRange(_ updateClosure: @escaping (ClosedRange<Double>) -> Void) {
+		timeRangeObserver = player?.currentItem?.observe(\.duration, options: [.new]) { item, _ in
+			let startTime = item.reversePlaybackEndTime.seconds
+			let endTime = item.forwardPlaybackEndTime.seconds
+			if !startTime.isNaN && !endTime.isNaN {
```
sindresorhus: Can it ever be NaN or are you just being cautious?

Gifski/TrimmingAVPlayerView.swift
```diff
@@ -0,0 +1,47 @@
...
+				updateClosure(startTime...endTime)
+			}
+		}
+	}
+
+	func hideTrimButtons() {
+		// This method is a collection of hacks, so might be acting funky on different OS versions
+		let avTrimView = firstSubview(where: { $0.description.contains("AVTrimView") }, deep: true)
```
sindresorhus: With:

```swift
extension NSObject {
	/// Returns the class name.
	class var simpleClassName: String {
		return String(describing: self)
	}

	/// Returns the class name of the instance.
	var simpleClassName: String {
		return type(of: self).simpleClassName
	}
}
```

You could just do 

```suggestion
		let avTrimView = firstSubview(where: { $0.simpleClassName == "AVTrimView" }, deep: true)
```

sindresorhus:
> Really interested if these hacks work on Mojave.

Works perfectly when https://github.com/sindresorhus/Gifski/pull/101#discussion_r309755650 is fixed.

Gifski/TrimmingAVPlayerView.swift
```diff
@@ -0,0 +1,47 @@
+import AVKit
+
+final class TrimmingAVPlayerView: AVPlayerView {
```
sindresorhus: Would be nice with a doc comments that explains what's special about this subclass.

Gifski/util.swift
```diff
@@ -2455,3 +2455,42 @@ extension NSViewController {
...
+
+extension NSLayoutConstraint {
+	func changing(
+		firstItem: Any? = nil,
+		firstAttribute: Attribute? = nil,
+		relation: Relation? = nil,
+		secondItem: NSView? = nil,
+		secondAttribute: Attribute? = nil,
+		multiplier: Double? = nil,
+		constant: CGFloat? = nil
```
sindresorhus: For consistency, this should be `Double` to then, like the above.

Gifski/TrimmingAVPlayerView.swift
sunshinejr: For some reason `viewDidMoveToSuperview` is not called - maybe due to the view being created in xib?

Gifski/TrimmingAVPlayerView.swift
sunshinejr: yeah, sometimes it returns noisy data so I had to do this otherwise it would crash

Gifski/TrimmingAVPlayerView.swift
sindresorhus: `func awakeFromNib()` then?

sunshinejr: It's too early to find the buttons - same thing when I tried just after `playerView.beginTrimming`

maybe if we transformed `TrimmingAVPlayerView` into a VC it would work?

sindresorhus: Yeah, most often doing a VC is a better option.

sunshinejr: It worked! Pushed the update.

sindresorhus:
@sunshinejr Thoughts on https://github.com/sindresorhus/Gifski/pull/101#issuecomment-516352222?

sunshinejr:
@sindresorhus hmm not sure what you mean with that 🤔

sindresorhus:
I think all the controls (dimensions, width, height, quality, etc) that are not the trim view should be below the trim view.

sunshinejr:
@sindresorhus ahh, gotcha, sure I will update it and we can compare!

sunshinejr:
@sindresorhus updated and I think it looks good!

sindresorhus:
Awesome work as usual, @sunshinejr 🙌

PR #107, opened 2019-08-05
Fix Quick Look when window is inactive on completion

### LINKED ISSUE #100: Quick Look doesn't work if the window is not focused
Quick Look doesn't work if the window is not focused (main/key) when the "conversion completed" screen is presented. Reproducible on macOS 10.14.6, but not on macOS 10.15 beta 3. See https://github.com/sindresorhus/Gifski/pull/94#issuecomment-512754498.

sindresorhus:
Fixes #100

---

With focus:

```
...
Responder chain:
(Main view controller: Optional(<Gifski.ConversionViewController: 0x600003b02220>)
  - <Gifski.ConversionViewController: 0x600003b02220>
  - <NSWindow: 0x600003e11300>
  - <Gifski.MainWindowController: 0x600002109ef0>
Responder chain:
(Main view controller: Optional(<Gifski.ConversionCompletedViewController: 0x10212bcf0>)
  - <Gifski.ConversionCompletedViewController: 0x10212bcf0>
  - <NSWindow: 0x600003e11300>
  - <Gifski.MainWindowController: 0x600002109ef0>
```

Without focus:

```
...
Responder chain:
(Main view controller: Optional(<Gifski.ConversionViewController: 0x600003b06680>)
  - <Gifski.ConversionViewController: 0x600003b06680>
  - <NSWindow: 0x600003e02600>
  - <Gifski.MainWindowController: 0x60000212d180>
```

Notice how the `ConversionCompletedViewController` never becomes first responder when the window is not active.

---

I'm gonna remove the debug utilities before merging.

### DIFF
--- a/Gifski/ConversionCompletedViewController.swift
+++ b/Gifski/ConversionCompletedViewController.swift
@@ -41,8 +41,8 @@ final class ConversionCompletedViewController: NSViewController {
 	override func viewDidAppear() {
 		super.viewDidAppear()
 
+		// This is needed for Quick Look to work.
 		self.view.window?.makeFirstResponder(self)
-		self.view.window?.printResponderChain()
 
 		if #available(macOS 10.14, *), defaults[.successfulConversionsCount] == 5 {
 			SKStoreReviewController.requestReview()
--- a/Gifski/ConversionCompletedViewController.swift
+++ b/Gifski/ConversionCompletedViewController.swift
@@ -41,6 +41,9 @@ final class ConversionCompletedViewController: NSViewController {
 	override func viewDidAppear() {
 		super.viewDidAppear()
 
+		self.view.window?.makeFirstResponder(self)
+		self.view.window?.printResponderChain()
+
 		if #available(macOS 10.14, *), defaults[.successfulConversionsCount] == 5 {
 			SKStoreReviewController.requestReview()
 		}
--- a/Gifski/ConversionViewController.swift
+++ b/Gifski/ConversionViewController.swift
@@ -42,14 +42,20 @@ final class ConversionViewController: NSViewController {
 		view = wrapper
 	}
 
-	override func viewDidLoad() {
-		super.viewDidLoad()
+	override func viewDidAppear() {
+		super.viewDidAppear()
+
+		view.window?.makeFirstResponder(self)
 
 		start(conversion: conversion)
 	}
 
 	/// Gets called when the Esc key is pressed.
-	/// Reference: https://stackoverflow.com/a/42440020
+	override func cancelOperation(_ sender: Any?) {
+		cancelConversion()
+	}
+
+	// TODO: Remove this when we target macOS 10.14.
 	@objc
 	func cancel(_ sender: Any?) {
 		cancelConversion()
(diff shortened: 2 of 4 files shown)


Gifski/util.swift
```diff
@@ -2449,7 +2449,6 @@ extension NSViewController {
 		}, completionHandler: {
 			window.contentViewController = viewController
 			viewController.view.animator().alphaValue = 1.0
-			window.makeFirstResponder(viewController)
```
sindresorhus: @sunshinejr This was the culprit. Why did you add it?

Gifski/util.swift
```diff
@@ -2449,7 +2449,6 @@ extension NSViewController {
 		}, completionHandler: {
```
sindresorhus: Unrelated, but why are we not using https://developer.apple.com/documentation/appkit/nsviewcontroller/1434431-present? We could use it with a custom animator: https://stackoverflow.com/a/28493373/64949

Gifski/util.swift
sunshinejr: Really interesting. So when I pushed the VC it wouldn't be the first responder so I had to do it manually.

Gifski/util.swift
sunshinejr: No particular reason, found that content switching API and used it.

Gifski/ConversionCompletedViewController.swift
```diff
@@ -41,6 +41,9 @@ final class ConversionCompletedViewController: NSViewController {
 	override func viewDidAppear() {
 		super.viewDidAppear()
 
+		self.view.window?.makeFirstResponder(self)
```
sunshinejr: really interesting: so making it a first responder in `viewDidAppear` is working but the generic way in `AppDelegate` with the app became active method is not? In that case we might need to see if other VCs also need to do that in `viewDidAppear`. Really strange.

sindresorhus: Yes, the problem seems to be making them first responder before their view is shown. Also really strange that it works in Catalina.

sunshinejr:
@sindresorhus after the fix I can't cancel with ESC on conversion VC. I think we should add the `.makeFirstRespoder` to all the VCs on `viewDidAppear`

sindresorhus:
> after the fix I can't cancel with ESC on conversion VC.

Fixed.

> I think we should add the .makeFirstRespoder to all the VCs on viewDidAppear

I don't think we should add it unless it's needed. I couldn't find any other cases it's needed.

Gifski/ConversionViewController.swift
```diff
@@ -42,16 +42,16 @@ final class ConversionViewController: NSViewController {
...
+		view.window?.makeFirstResponder(self)
 
 		start(conversion: conversion)
 	}
 
 	/// Gets called when the Esc key is pressed.
-	/// Reference: https://stackoverflow.com/a/42440020
-	@objc
-	func cancel(_ sender: Any?) {
+	override func cancelOperation(_ sender: Any?) {
```
sindresorhus: I also changed this as with the macOS 10.14 SDK, NSViewController has `cancelOperation` too.

PR #114, opened 2019-08-07
Ensure the trim range is valid

sindresorhus:
This hopefully fixes a crash report we received in Crashlytics:

```
TrimmingAVPlayerViewController.swift line 63
closure #1 in TrimmingAVPlayerView.observeTrimmedTimeRange(_:) + 63
```

### DIFF
--- a/Gifski/util.swift
+++ b/Gifski/util.swift
@@ -2515,3 +2515,18 @@ extension NSObject {
 		return type(of: self).simpleClassName
 	}
 }
+
+extension AVPlayerItem {
+	/// The playable range of the item.
+	/// Can be `nil` when the `.duration` is not available, for example, when the asset has not yet been fully loaded or if it's a live stream.
+	var playbackRange: ClosedRange<Double>? {
+		guard duration.isNumeric else {
+			return nil
+		}
+
+		let startTime = reversePlaybackEndTime.isNumeric ? reversePlaybackEndTime.seconds : 0
+		let endTime = forwardPlaybackEndTime.isNumeric ? forwardPlaybackEndTime.seconds : duration.seconds
+
+		return startTime < endTime ? startTime...endTime : endTime...startTime
+	}
+}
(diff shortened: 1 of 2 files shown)


Gifski/util.swift
```diff
@@ -2515,3 +2515,18 @@ extension NSObject {
...
+
+extension AVPlayerItem {
+	/// The playable range of the item.
+	/// Can be `nil` when the `.duration` is not available, for example, when the asset has not yet been fully loaded or if it's a live stream.
+	var playbackRange: ClosedRange<Double>? {
+		guard duration.isNumeric else {
+			return nil
+		}
+
+		let startTime = reversePlaybackEndTime.isNumeric ? reversePlaybackEndTime.seconds : 0
```
sindresorhus: Instead of checking if the seconds are `nan`, we check that the CMTime is numeric and valid, which is a more correct way to handle it, and if not, we fall back to the full range.

Gifski/util.swift
```diff
@@ -2515,3 +2515,18 @@ extension NSObject {
...
+		let endTime = forwardPlaybackEndTime.isNumeric ? forwardPlaybackEndTime.seconds : duration.seconds
+
+		return startTime < endTime ? startTime...endTime : endTime...startTime
```
sunshinejr:
```suggestion
		return startTime < endTime ? startTime...endTime : nil
```

this `endTime...startTime` shouldn't probably happen but in case it does, maybe we should just bail and return `nil`?

Gifski/util.swift
sunshinejr: ah nice catch 👍

Gifski/util.swift
```diff
@@ -2515,3 +2515,18 @@ extension NSObject {
 		return type(of: self).simpleClassName
 	}
 }
+
+extension AVPlayerItem {
+	/// The playable range of the item.
+	/// Can be `nil` when the `.duration` is not available, for example, when the asset has not yet been fully loaded or if it's a live stream.
+	var playbackRange: ClosedRange<Double>? {
```
sunshinejr: 👍

Gifski/util.swift
sindresorhus: I was making this generic as there are other use-cases than ours where it could happen. For example, when someone sets `reversePlaybackEndTime` manually. I don't really see any downside in *just handling it*. Do you?

sunshinejr: ah ye, I was just scared of this use-case maybe 😄 let’s try that 👍🏻

PR #121, opened 2019-08-12
Fix tooltip crash

sindresorhus:
This fixes a crash that could happen if a user quickly drags in a file right after the conversion completed view is presented, before the tooltip has been shown.

Fixes https://fabric.io/sindre-sorhus-projects/mac/apps/com.sindresorhus.gifski/issues/fbbde1a1afc20492a9eb6961bcedbc1e?time=last-thirty-days (Only accessible by maintainers)

### DIFF
--- a/Gifski/ConversionCompletedViewController.swift
+++ b/Gifski/ConversionCompletedViewController.swift
@@ -61,7 +61,11 @@ final class ConversionCompletedViewController: NSViewController {
 			wrapperView.fadeIn(duration: 0.5, delay: 0.15, completion: nil)
 		}
 
-		delay(seconds: 1) {
+		delay(seconds: 1) { [weak self] in
+			guard let self = self else {
+				return
+			}
+
 			self.tooltip.show(from: self.draggableFile, preferredEdge: .maxY)
 		}
 	}
--- a/Gifski/Tooltip.swift
+++ b/Gifski/Tooltip.swift
@@ -40,6 +40,11 @@ final class Tooltip: NSPopover {
 		of positioningView: NSView,
 		preferredEdge: NSRectEdge
 	) {
+		guard positioningView.window != nil else {
+			assertionFailure("Tooltip must be shown from a view with a window")
+			return
+		}
+
 		if !showOnlyOnce || (showOnlyOnce && defaults[showKey] < 1) {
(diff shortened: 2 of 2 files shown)


Gifski/Tooltip.swift
```diff
@@ -40,6 +40,11 @@ final class Tooltip: NSPopover {
 		of positioningView: NSView,
 		preferredEdge: NSRectEdge
 	) {
+		guard positioningView.window != nil else {
+			assertionFailure("Tooltip must be shown from a view with a window")
+			return
+		}
```
sindresorhus: This is just a preventive measure to ensure the tooltip can never accidentally crash the app. Without this, `super.show` could cause AppKit to emit a fatal error.

Gifski/Tooltip.swift
```diff
@@ -40,6 +40,11 @@ final class Tooltip: NSPopover {
...
+
 		if !showOnlyOnce || (showOnlyOnce && defaults[showKey] < 1) {
 			defaults[showKey] += 1
 			super.show(relativeTo: positioningRect, of: positioningView, preferredEdge: preferredEdge)
```
sindresorhus: @sunshinejr Just an idea, should be wrap this in `DispatchQueue.main.async {}` to ensure it's run in the next run loop, just as a preventive measure against race issues with other animations.

Gifski/ConversionCompletedViewController.swift
```diff
@@ -61,7 +61,11 @@ final class ConversionCompletedViewController: NSViewController {
 			wrapperView.fadeIn(duration: 0.5, delay: 0.15, completion: nil)
 		}
 
-		delay(seconds: 1) {
+		delay(seconds: 1) { [weak self] in
```
sunshinejr: ah another good one 👍

Gifski/Tooltip.swift
sunshinejr: hmm, good idea, we could try!

sindresorhus: Thinking about this more, doing this might causes other problems, like the tooltip not being shown if a modal dialog is presented until after it's dismissed. It might be better to wrap it yourself when needed.

PR #129, opened 2019-09-05
Improve the GIF size estimate

### LINKED ISSUE #41: Improve file size estimate
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/gifski-app/issues/41)

GifSki is my go-to gif-app. But,,,,

### Problem 😤
I always have to convert videos 3 times  before I get close to the size I want. Changing FPS, Quality and resolution to hit the sweet spot. Which in my case is usually under 5mb.

### Why is this relevant? 💡
I create a lot of gif videos to show how apps work etc. I host these on github, because GitHub is where work happens 👔. I need to be able to post these videos in GitHub comment fields etc. Github has a max limit of around 5.2mb So I want to keep my gif videos a bit under 5mb.

### Thinking 🤔
So Gifski is unique because it creates a unique Color-maps for each frame. Ball-park correct? So the file guesstimater of old doesn't work, you used to be able to do 256 * FramesInTotal = size. But you can't now because each frame is not defined from scratch.

### Solution 💰
✨ `FileSizeGuestimatorPro.swift` ✨ 😇

1. On a background thread (at startup, and while user interacts with settings)
2. `let segmentCount = (TotalFrameCount 240/4) = 60`
3. `let sizes:Int = (0..<4).indecies.map {i in GifSkiConverter.gifFrame(video:test.mp4,frameIdx:i* segmentCount).size }`
4. `let size:Int = sizes.reduce(0,+) * segmentCount `
5. `onBGComplete = {/*double the sampling, on each pass*/} //👈 rinse repeat until the size is hardned, or user changes variables or cancels`
6. Bonus point if these samples are re-used in the final conversion. 

---

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $182.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($180.00)
- [0maxxam0 0maxxam0](https://issuehunt.io/u/0maxxam0) ($2.00)
### Submitted pull Requests
- [#129 Improved Estimate](https://issuehunt.io/r/sindresorhus/gifski-app/pull/129)
---

### Tips

---
</details>

benkoska:
Fixes #41

When the estimateFileSize function is called it first cancels any ongoing estimations.
Then it runs the old estimate and show it in gray saying "Estimated size: 123Kb".
Then it starts to actually run the conversion of the Gif in the background and once it's done it will save the gif in the gifdata variable for later use and shows the size in green saying "File size: 123Kb".

When pressing the Convert button it first checks if there is gifdata is set. If so it will send it to the Conversion screen that will only have to save the data into a file so the conversion will complete almost instantly.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#41: Improve file size estimate](https://issuehunt.io/repos/119822304/issues/41)
---

</details>


sindresorhus:
Thanks for working on this. Can you include some text in the PR description on how you have implemented this? The diff is huge and I'd like to be able to read in words what's changed and why :)

sindresorhus:
Also, you've changed all of the tab indentation to spaces. You'll need to revert that.

DeveloperBen:
Ok, I will add a description and revert the tab indentations.

DeveloperBen:
I added the PR description. I am trying to figure out how to fix the tab indentations.

sindresorhus:
> Then it starts to actually run the conversion of the Gif in the background and once it's done it will 

That is a problem. It cannot wait until it's done to show the estimate. A GIF conversion can easily take 10 min or (a lot) more. Users will not be on the settings view that long. As commented in https://github.com/sindresorhus/Gifski/issues/41#issuecomment-431148672, the estimate needs to be constantly refined based on the background conversion.

DeveloperBen:
I initially implemented that (and I can revert) but I think that it’s very confusing when your adjusting the settings and the the estimate is constantly changing even though your not changing the settings. The estimate might also go up when the quality goes down because the initial estimate is very inaccurate.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

sindresorhus:
It wouldn’t constantly change. We would only update it on an interval. But how it’s implemented now is like not having it at all, since nobody will be on the view long enough for the estimate to finish, unless the video is very small.

sindresorhus:
To get a better representation of the overall size, we could start converting 10 samples (evenly spread out over the video duration) from the whole video concurrently. That would give us a more accurate estimate faster.

sindresorhus:
I’m open to other ideas.

DeveloperBen:
We should try a few things but have a maximum of two states rough and more accurate because even if it updates on an interval it will get confusing if it is changing a lot without you changing the values. That’s why we need to have the two colors so that it’s clear why the value changed

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kornelski:
The app now handles writing itself, so it knows exactly how much data has been written after each frame.

This means it can guess file size of the whole animation by dividing amount of data written so far by percentage of frames completed. Progress callback is a good place to calculate it.

DeveloperBen:
So should I update the estimate every 10% or so

kornelski:
You can update the estimate very frame. 

The estimate is likely to be very inaccurate in the beginning, so you could update the estimate very frame, but only after, say, 10 frames.

sindresorhus:
@kornelski What do you think of my idea to run multiple samples concurrently (unsure how many at once) from various places in the video to better represent the whole video? Imagine if the first half of the video has little movement and the other half has.

kornelski:
Yup, that would work as well. You'd have to ignore size of the first frame of each run (since usually it'd be encoded as a difference from a previous frame, but if you encode a cut separately, it'll force first frame of the cut to be a "keyframe").

DeveloperBen:
I changed the estimate to update it whenever a new frame is written. It does not wait for the whole thing to finish anymore.

Gifski/EditVideoViewController.swift
```diff
@@ -338,18 +342,33 @@ final class EditVideoViewController: NSViewController {
...
+					return
+				}
+
+				self.framesWritten = framesWritten
+				DispatchQueue.main.sync {
+					self.estimatedSizeLabel.stringValue = "Estimated size: " + self.formatter.string(fromByteCount: Int64(self.dataSize * (Int64(frameCount) / Int64(framesWritten))))
+				}
+			}
+			Gifski.setWriteCallback { size in
+				self.dataSize += Int64(size)
```
kornelski: the library supports only 1 callback at a time, so this will *replace* the other callback that gets the data.

DeveloperBen: Yes. I am also only converting on thing at a time. Either estimating or actually converting.

kornelski: I thought you also build gifData at the same time?

DeveloperBen: No I set gifData at the end of the conversion.

DeveloperBen: Anything else?

kornelski: Just empty gifData? gifData is appended to from a write callback.

DeveloperBen: the setWriteCallback does not actually override the writeCallback with Gifski it is called after everything else is run onWriteCallback.

DeveloperBen: Gifski.swift still writes the gifData and then I get it to save later on at the end of the conversion

kornelski: Ah, I assumed the setter was just forwarding it to the underlying library instead of adding its own layer of callback handling.

DeveloperBen:
Any other changes needed?

kornelski:
LGTM

DeveloperBen:
Perfect

DeveloperBen:
@sindresorhus 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Gifski/Gifski.swift
```diff
@@ -190,7 +204,11 @@ final class Gifski {
 						do {
 							try gifski.finish()
 							// Force unwrapping here is safe because we nil gifData in one single place after this.
-							completionHandlerOnce(.success(gifData! as Data))
+							guard let gifData = gifData as? Data else {
```
sindresorhus: This causes a compiler warning.

Gifski/Gifski.swift
```diff
@@ -190,7 +204,11 @@ final class Gifski {
 						do {
 							try gifski.finish()
 							// Force unwrapping here is safe because we nil gifData in one single place after this.
```
sindresorhus: This comment is outdated.

Gifski/Gifski.swift
```diff
@@ -49,6 +49,17 @@ final class Gifski {
...
+	private static var writeCallback: ((Int) -> Void)?
+	private static var onNewFrame: ((Int64) -> Void)?
+
+	static func setWriteCallback(_ callback: ((Int) -> Void)?) {
+		writeCallback = callback
+	}
+
+	static func setOnNewFrame(_ callback: ((Int64) -> Void)?) {
+			onNewFrame = callback
+		}
```
sindresorhus: These methods are moot. It's clearer to just assign the closure directly as it only supports one closure.

Gifski/Gifski.swift
```diff
@@ -49,6 +49,17 @@ final class Gifski {
 	// This is carefully nil'd and should remain private.
 	private static var gifData: NSMutableData?
 
+	private static var writeCallback: ((Int) -> Void)?
```
sindresorhus: The naming here is inconsistent with the below.

Gifski/EditVideoViewController.swift
```diff
@@ -310,6 +320,20 @@ final class EditVideoViewController: NSViewController {
 	}
 
 	private func estimateFileSize() {
+		// Cancle old estimate
```
sindresorhus: Typo

Gifski/ConversionViewController.swift
```diff
@@ -16,12 +16,14 @@ final class ConversionViewController: NSViewController {
 	private lazy var timeRemainingEstimator = TimeRemainingEstimator(label: timeRemainingLabel)
 
 	private var conversion: Gifski.Conversion!
+	private var gifData: Data?	
 	private var progress: Progress?
 	private var isRunning = false
 
-	convenience init(conversion: Gifski.Conversion) {
+	convenience init(_ conversion: Gifski.Conversion, _ gifData: Data?) {
```
sindresorhus: Don't do unrelated changes.

```suggestion
	convenience init(conversion: Gifski.Conversion, _ gifData: Data?) {
```

Gifski/EditVideoViewController.swift
```diff
@@ -318,10 +342,43 @@ final class EditVideoViewController: NSViewController {
 			}
 		}()
 		let frameCount = duration * frameRateSlider.doubleValue
-		let dimensions = resizableDimensions.changed(dimensionsType: .pixels).currentDimensions.value
-		var fileSize = (Double(dimensions.width) * Double(dimensions.height) * frameCount) / 3
-		fileSize = fileSize * (qualitySlider.doubleValue + 1.5) / 2.5
-		estimatedSizeLabel.stringValue = "Estimated size: " + formatter.string(fromByteCount: Int64(fileSize))
+
+		estimatedSizeLabel.textColor = NSColor(red: 0.13, green: 0.13, blue: 0.13, alpha: 1.0)
```
sindresorhus: Use the semantic system colors.

sindresorhus:
- The estimated size label is not legible in dark mode.
- The label should only be green when the estimation is done.
- I don't think it should "Starting Estimation". Just show the normal label with `…` as size.
- You added a lot of complications and duplicate logic to be able to reuse the GIF result from the estimation. I don't think it's worth it. The only time we can reuse it is for tiny videos and they will convert fast anyway.
- I tried it with [this](https://github.com/sindresorhus/Gifski/files/3609873/Test.-.super.short.mp4.zip) file and it was stuck on "Starting Estimation". It only changed when I touched any of the controls.
- With the same file, the *final* estimation said 1.3 MB, while the actual GIF after conversion ended up being 74 KB. So something is wrong with the estimation.

sindresorhus:
> The estimate is likely to be very inaccurate in the beginning, so you could update the estimate very frame, **but only after, say, 10 frames.**

Gifski/ConversionViewController.swift
sunshinejr: hmm this seems strange that we still need the `Conversion` object even though we had the gif data already 

I know that we need this object in the ConversionCompleted vc, but maybe we could try to think about better structure of this?

ideally I'd opt for an `enum` on this init that would tell us whether we need a gif conversion from the video or do we actually have the gif data already, e.g.:
```swift
enum GifSource { // <-- hate this name so this is only an example
    case conversionNeeded(Conversion)
    case gifDataAvailable(Data)
}
```

this would make it really clear in code with what data we work

Gifski/ConversionViewController.swift
```diff
@@ -77,24 +79,39 @@ final class ConversionViewController: NSViewController {
...
-					try result.get().write(to: gifUrl, options: .atomic)
+					try gifData.write(to: gifUrl, options: .atomic)
+
 					try? gifUrl.setMetadata(key: .itemCreator, value: "\(App.name) \(App.version)")
 					defaults[.successfulConversionsCount] += 1
 
 					self.didComplete(conversion: conversion, gifUrl: gifUrl)
 				} catch Gifski.Error.cancelled {
-					self.cancelConversion()
+					cancelConversion()
```
sunshinejr: I think this case shouldn't happen for `gifData` - `try gifData.write` shouldn't throw `Gifski.Error.cancelled` so we may remove this specialized `catch`

Gifski/ConversionViewController.swift
```diff
@@ -77,24 +79,39 @@ final class ConversionViewController: NSViewController {
...
+				Gifski.run(conversion, completionHandler: { result in
+					do {
+						try result.get().write(to: gifUrl, options: .atomic)
+					} catch Gifski.Error.cancelled {
+						self.cancelConversion()
+					} catch {
+						self.presentError(error, modalFor: self.view.window)
+						self.cancelConversion()
+					}
+					try? gifUrl.setMetadata(key: .itemCreator, value: "\(App.name) \(App.version)")
```
sunshinejr: can `gifUrl.setMetadata` && `defaults[. successfulConversionsCount] += 1` be moved to `self.didComplete()`?

Gifski/EditVideoViewController.swift
```diff
@@ -66,8 +71,13 @@ final class EditVideoViewController: NSViewController {
 			frameRate: frameRateSlider.integerValue
 		)
 
-		let convert = ConversionViewController(conversion: conversion)
-		push(viewController: convert)
+		if let gifData = gifData {
```
sunshinejr: this if shouldn't be needed as `gifData` is optional in the initializer

Gifski/EditVideoViewController.swift
```diff
@@ -310,6 +320,20 @@ final class EditVideoViewController: NSViewController {
 	}
 
 	private func estimateFileSize() {
+		// Cancle old estimate
+		estimateWorkItem?.cancel()
```
sunshinejr: maybe we could extract these 4 lines to a method e.g. `cancelEstimationProcess()` and then we could remove this comment above as well?

Gifski/EditVideoViewController.swift
```diff
@@ -318,10 +342,43 @@ final class EditVideoViewController: NSViewController {
...
 		let frameCount = duration * frameRateSlider.doubleValue
-		let dimensions = resizableDimensions.changed(dimensionsType: .pixels).currentDimensions.value
-		var fileSize = (Double(dimensions.width) * Double(dimensions.height) * frameCount) / 3
-		fileSize = fileSize * (qualitySlider.doubleValue + 1.5) / 2.5
-		estimatedSizeLabel.stringValue = "Estimated size: " + formatter.string(fromByteCount: Int64(fileSize))
+
+		estimatedSizeLabel.textColor = NSColor(red: 0.13, green: 0.13, blue: 0.13, alpha: 1.0)
+		estimatedSizeLabel.stringValue = "Starting Estimation..."
+		estimateWorkItem = DispatchWorkItem {
+			print("Total: \(frameCount)")
```
sunshinejr: ^^ print left in the code

Gifski/EditVideoViewController.swift
```diff
@@ -318,10 +342,43 @@ final class EditVideoViewController: NSViewController {
...
+
+			Gifski.setOnNewFrame { framesWritten in
+				DispatchQueue.main.sync {
+					self.estimatedSizeLabel.textColor = NSColor(red: 0.11, green: 0.37, blue: 0.13, alpha: 1.0)
```
sunshinejr: this is probably creating retain cycle as the `DispatchWorkItem` is retaining the parent by using `self` and the parent retains the work item by storing it as its property

Gifski/EditVideoViewController.swift
```diff
@@ -318,10 +342,43 @@ final class EditVideoViewController: NSViewController {
...
-		fileSize = fileSize * (qualitySlider.doubleValue + 1.5) / 2.5
-		estimatedSizeLabel.stringValue = "Estimated size: " + formatter.string(fromByteCount: Int64(fileSize))
+
+		estimatedSizeLabel.textColor = NSColor(red: 0.13, green: 0.13, blue: 0.13, alpha: 1.0)
+		estimatedSizeLabel.stringValue = "Starting Estimation..."
+		estimateWorkItem = DispatchWorkItem {
+			print("Total: \(frameCount)")
+
+			Gifski.setOnNewFrame { framesWritten in
+				DispatchQueue.main.sync {
```
sunshinejr: why `sync`? are we sure callback in `setOnNewFrame` is always gonna come from a thread different than main so we can avoid deadlock?

Gifski/EditVideoViewController.swift
```diff
@@ -318,10 +342,43 @@ final class EditVideoViewController: NSViewController {
...
+					self.estimatedSizeLabel.textColor = NSColor(red: 0.11, green: 0.37, blue: 0.13, alpha: 1.0)
+				}
+
+				if self.dataSize == 0 {
+					return
+				}
+
+				self.framesWritten = framesWritten
+				DispatchQueue.main.sync {
```
sunshinejr: again sync in here

Gifski/EditVideoViewController.swift
```diff
@@ -318,10 +342,43 @@ final class EditVideoViewController: NSViewController {
...
+					self.estimatedSizeLabel.stringValue = "Estimated size: " + self.formatter.string(fromByteCount: Int64(self.dataSize * (Int64(frameCount) / Int64(framesWritten))))
+				}
+			}
+			Gifski.setWriteCallback { size in
```
sunshinejr: also, how about having these two callbacks as arguments for the `run(conversion)` function? I don't like these static properties as they can be replaced any time and this would resolve in an undefined behavior by calling old conversion on new callbacks

Gifski/ConversionViewController.swift
DeveloperBen: sure

DeveloperBen:
@sindresorhus

sindresorhus:
If I try with [this video](https://github.com/sindresorhus/Gifski/files/3622518/Unicorn-Apple.ProRes.4444.XQ.with.Alpha.mov.zip) and then drag the right trim handle to the middle, it shows some weird estimations. It jumps between `Zero KB` and `41 MB`, while the final estimation is `4.6 MB`. There must be a bug somewhere as the estimation should never be able to be `Zero KB` for a video that is not empty.

sindresorhus:
The estimation jumps too much. It's distracting as a user. I don't think showing estimation every frame is working out. I think we should throttle it to max every 5 (?) seconds.

I also think the calculation/algorithm should put more weight into the existing guesstimate. I keep seeing jumps from `5 MB` to `244 MB` and then less than a second later back to `7 MB`. It should not be able to jump that much. We need to smoothen it out more.

sindresorhus:
When I drag in the video in https://github.com/sindresorhus/Gifski/pull/129#issuecomment-532302315, then trim it, and then click "Convert" right away, it's just stuck at 0% for a long time, then progresses, shows the checkmark, but never shows the completed screen.

sindresorhus:
You did not resolve all the review comments.

sindresorhus:
Currently it restarts the estimation if you tab between controls. It should not do that.

Gifski/EditVideoViewController.swift
```diff
@@ -318,10 +343,41 @@ final class EditVideoViewController: NSViewController {
...
+					self.gifData = data
+					self.estimatedSizeLabel.textColor = NSColor(named: "green")
+					self.estimatedSizeLabel.stringValue = "File size: " + self.formatter.string(fromByteCount: Int64(data.count))
+				} catch {
+					self.presentError(error, modalFor: self.view.window)
+				}
+			}
+		}
+		guard let workItem = estimateWorkItem else {
+			// This should never happen as we define estimateWorkItem above
```
sindresorhus: Would be good to use `assertionFailure()` just in case it actually happens. The comment could be the message.

sindresorhus:
It would be great if you could spend more time on perfecting and thoroughly testing this PR before requesting a review again :)

Gifski/EditVideoViewController.swift
```diff
@@ -343,16 +344,12 @@ final class EditVideoViewController: NSViewController {
 		}()
 		let frameCount = duration * frameRateSlider.doubleValue
 
-		estimatedSizeLabel.textColor = NSColor(red: 0.13, green: 0.13, blue: 0.13, alpha: 1.0)
-		estimatedSizeLabel.stringValue = "Starting Estimation..."
+		estimatedSizeLabel.textColor = NSColor(named: "gray")
+		estimatedSizeLabel.stringValue = "..."
```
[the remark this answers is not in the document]
DeveloperBen: Did not know there is a character …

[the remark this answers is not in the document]
kornelski: … is typed with alt+:

kornelski:
This measurement would be very useful now that we have lossy compression, because it added another hard-to-estimate factor.

DeveloperBen:
I can work on it right away.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

sindresorhus:
Closing for lack of activity.

PR #134, opened 2019-09-14
Confirm quitting the app during conversion

sindresorhus:
To prevent people accidentally quitting.

Screenshot 2019-09-14 at 16 07 02

The conversion continues behind it.

The wording is taken from the Photos app when exporting:

image

### DIFF
--- a/Gifski/AppDelegate.swift
+++ b/Gifski/AppDelegate.swift
@@ -52,6 +52,26 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
 	func applicationShouldTerminateAfterLastWindowClosed(_ sender: NSApplication) -> Bool { true }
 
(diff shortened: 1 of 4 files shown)


Gifski/AppDelegate.swift
```diff
@@ -52,6 +52,26 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
 	func applicationShouldTerminateAfterLastWindowClosed(_ sender: NSApplication) -> Bool { true }
 
+	func applicationShouldTerminate(_ sender: NSApplication) -> NSApplication.TerminateReply {
+		if mainWindowController.isConverting {
+			let response = NSAlert.showModal(
+				for: mainWindowController.window,
+				message: "Do you want to continue converting?",
+				informativeText: "Gifski is currently converting your video. If you quit, the conversion will be cancelled.",
```
yakov116: Maybe just me. Bit think it should be `converting a video`

PR #142, opened 2019-10-07
Fix CircularProgress crash

sindresorhus:
There has been 21 crashes in v2.1.0 for this:

```
Fatal Exception: NSInternalInconsistencyException
NSWindow drag regions should only be invalidated on the Main Thread!
6  Gifski                         0x1068d8325 CircularProgress.isFinished.setter (<compiler-generated>)
7  Gifski                         0x1068d7de0 CircularProgress.progress.setter + 107 (CircularProgress.swift:107)
8  Gifski                         0x1068d8a50 closure #2 in CircularProgress.progressInstance.didset + 161 (CircularProgress.swift:161)
9  Gifski                         0x10691b24f thunk for @escaping @callee_guaranteed (@guaranteed NSProgress, @in_guaranteed NSKeyValueObservedChange<Bool>) -> () (<compiler-generated>)
10 libswiftFoundation.dylib       0x7fff72311bb6 $s10Foundation27_KeyValueCodingAndObservingPAAE7observe_7options13changeHandlerAA05NSKeyC11ObservationCs0B4PathCyxqd__G_So0kcF7OptionsVyx_AA0kC14ObservedChangeVyqd__GtctlFySo8NSObjectC_AOyypGtcfU_
11 libswiftFoundation.dylib       0x7fff723a2fb9 $s10Foundation27_KeyValueCodingAndObservingPAAE7observe_7options13changeHandlerAA05NSKeyC11ObservationCs0B4PathCyxqd__G_So0kcF7OptionsVyx_AA0kC14ObservedChangeVyqd__GtctlFySo8NSObjectC_AOyypGtcfU_TA
12 libswiftFoundation.dylib       0x7fff723a2e17 $s10Foundation21NSKeyValueObservationC019_swizzle_me_observeC010forKeyPath2of6change7contextySSSg_ypSgSDySo8NSStringCypGSgSvSgtFTf4dnndn_n
13 libswiftFoundation.dylib       0x7fff72311316 $s10Foundation21NSKeyValueObservationC019_swizzle_me_observeC010forKeyPath2of6change7contextySSSg_ypSgSDySo8NSStringCypGSgSvSgtFTo
14 Foundation                     0x7fff48a2ff08 NSKeyValueNotifyObserver
15 Foundation                     0x7fff48a2f984 NSKeyValueDidChange
16 Foundation                     0x7fff48bb9ead NSKeyValueDidChangeWithPerThreadPendingNotifications.llvm.7538459074364299148
17 CoreFoundation                 0x7fff4681b867 (Missing)
18 Foundation                     0x7fff48a8e12a -[NSProgress _setValueForKeys:settingBlock:]
19 Foundation                     0x7fff48b05a14 (Missing)
20 Foundation                     0x7fff48a8f75a (Missing)
21 Gifski                         0x1068f6e87 @objc closure #2 in static Gifski.run(_:completionHandler:) + 93 (Gifski.swift:93)
22 Gifski                         0x106924a5d _$LT$gifski..progress..ProgressCallback$u20$as$u20$gifski..progress..ProgressReporter$GT$::increase::h38ffd2e7bb6b7e57
23 Gifski                         0x106931aa0 gifski::Writer::write::haa6ec2bf93d87d93
24 Gifski                         0x106929bd1 gifski::c_api::gifski_write_sync_internal::hd59452488b4436f5
25 Gifski                         0x106939337 std::sys_common::backtrace::__rust_begin_short_backtrace::hf426f0f1bb683314
26 Gifski                         0x10692aa5c std::panicking::try::do_call::hfa9c1991c45d90cf (.llvm.8325536553258157362)
27 Gifski                         0x1069c088c __rust_maybe_catch_panic + 31 (lib.rs:31)
28 Gifski                         0x10692d160 core::ops::function::FnOnce::call_once$u7b$$u7b$vtable.shim$u7d$$u7d$::h0e110dce52ed5d44
29 Gifski                         0x10699489e _$LT$alloc..boxed..Box$LT$F$GT$$u20$as$u20$core..ops..function..FnOnce$LT$A$GT$$GT$::call_once::ha44ca87baccc45af + 220 (alloc.rs:220)
30 Gifski                         0x1069bf97e std::sys::unix::thread::Thread::new::thread_start::hceb9fd3b79bf176c + 104 (alloc.rs:104)
```

Seems the macOS 10.15 SDK got stricter. I have not been able to reproduce this though. So this is just a shot in the dark. I'm just making sure anything UI related is done on the main process.

### DIFF
--- a/Gifski.xcodeproj/xcshareddata/xcschemes/Gifski.xcscheme
+++ b/Gifski.xcodeproj/xcshareddata/xcschemes/Gifski.xcscheme
@@ -59,8 +59,6 @@
       useCustomWorkingDirectory = "NO"
       ignoresPersistentStateOnLaunch = "NO"
       debugDocumentVersioning = "YES"
-      stopOnEveryThreadSanitizerIssue = "YES"
-      stopOnEveryMainThreadCheckerIssue = "YES"
       migratedStopOnEveryIssue = "YES"
       debugServiceExtension = "internal"
       allowLocationSimulation = "YES">
--- a/Gifski/Vendor/CircularProgress+Util.swift
+++ b/Gifski/Vendor/CircularProgress+Util.swift
@@ -225,3 +225,12 @@ extension NSWindow {
 		attachedSheet != nil
 	}
 }
+
+
+func assertMainThread(
+	function: StaticString = #function,
+	file: String = #file,
+	line: UInt = #line
+) {
+	assert(Thread.isMainThread, "\(function) in \(file.nsString.lastPathComponent):\(line) must run on the main thread!")
+}
--- a/Gifski/Vendor/CircularProgress.swift
+++ b/Gifski/Vendor/CircularProgress.swift
@@ -82,7 +82,6 @@ public final class CircularProgress: NSView {
 	The progress value in the range `0...1`.
 
 	- Note: The value will be clamped to `0...1`.
-	- Note: Can be set from a background thread.
 	*/
 	@IBInspectable public var progress: Double {
 		get {
@@ -132,12 +131,12 @@ public final class CircularProgress: NSView {
 			_isFinished = newValue
 
 			if _isFinished {
-				self.isIndeterminate = false
+				isIndeterminate = false
 
-				if !self.isCancelled, self.showCheckmarkAtHundredPercent {
-					self.progressLabel.string = ""
-					self.cancelButton.isHidden = true
-					self.successView.isHidden = false
+				if !isCancelled, showCheckmarkAtHundredPercent {
+					progressLabel.string = ""
+					cancelButton.isHidden = true
+					successView.isHidden = false
 				}
 			}
 		}
@@ -439,10 +438,10 @@ public final class CircularProgress: NSView {
 			_isIndeterminate = newValue
 			didChangeValue(for: \.isIndeterminate)
 
-			if self._isIndeterminate {
-				self.startIndeterminateState()
+			if _isIndeterminate {
+				startIndeterminateState()
 			} else {
-				self.stopIndeterminateState()
+				stopIndeterminateState()
 			}
 		}
 	}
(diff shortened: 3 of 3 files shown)


sindresorhus:
// @Ybrin @SergeyKuryanov @DeveloperBen  In case you have time to review.

sindresorhus:
I'm starting to think that allowing any method to be called on any thread was not a good idea in general. But I don't really want to do major changes to it at this time. SwiftUI will make all this much simpler.

kornelski:
The next few lines also crash:

```swift

			if !progressLabel.isHidden {
				progressLabel.string = "\(Int(_progress * 100))%"
				successView.isHidden = true
			}
```

SergeyKuryanov:
In general I am not fan of threading code like this and limit it to some services. This makes testing and debugging much easier. But if we need to allow to call properties from any thread, how about using some helper like this:
```swift
extension DispatchQueue {
    static func syncOnMainIfNeeded(_ work: () -> Void) {
        if Thread.isMainThread {
            work()
        } else {
            DispatchQueue.main.sync(execute: work)
        }
    }
}
```

To wrap _all setters_ inside it:
```swift
set {
    DispatchQueue.syncOnMainIfNeeded {
        // setting values, update UI, etc.
    }
}

sindresorhus:
@kornelski How are you able to get it to crash?

kornelski:
Just by running a conversion. Catalina complains `successView` is modified on non-main thread.

sindresorhus:
> Seems like this should fix it. But why do you even call isFinished from other threads? UI classes like that should actually always require being called from the UI thread. Just like in the Apple APIs.

When I made CircularProgress I naively thought it would be nice if users could call it from any thread. I now see the problem with that thinking.

sindresorhus:
I've updated this PR to make sure CircularProgress is only called on the main thread. I've added a lot of asserts to ensure none of public properties are ever called on a non-main thread.

sindresorhus:
@SergeyKuryanov That does look tempting from the outside, but it has a lot of downsides. See: http://blog.benjamin-encz.de/post/main-queue-vs-main-thread/

PR #146, opened 2019-10-18
Add Internet Access Policy

sindresorhus:
See https://obdev.at/iap/index.html

This makes it clearer for people using Little Snitch (most popular firewall) why we connect to the server and it makes it less likely they will deny it.

### DIFF
--- a/Gifski/InternetAccessPolicy.json
+++ b/Gifski/InternetAccessPolicy.json
@@ -1,7 +1,7 @@
 {
 	"ApplicationDescription": "Gifski converts videos to high-quality GIFs.",
 	"DeveloperName": "Sindre Sorhus",
-	"Website": "https://sindresorhus.com/dato",
+	"Website": "https://sindresorhus.com/gifski",
 	"Connections" : [
 		{
 			"Host" : "*.crashlytics.com",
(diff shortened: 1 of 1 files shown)


Gifski/InternetAccessPolicy.json
```diff
@@ -0,0 +1,21 @@
...
+	"ApplicationDescription": "Gifski converts videos to high-quality GIFs.",
+	"DeveloperName": "Sindre Sorhus",
+	"Website": "https://sindresorhus.com/dato",
+	"Connections" : [
+		{
+			"Host" : "*.crashlytics.com",
+			"NetworkProtocol" : "TCP",
+			"Port" : "443",
+			"Purpose" : "Gifski sends crash reports to this server.",
+			"DenyConsequences" : "If you deny this connection, crash reports will not be sent and bugs might stay unfixed."
```
sindresorhus: I'm not 100% happy about this sentence.

sunshinejr:
```suggestion
			"DenyConsequences" : "If you deny this connection, we won't be able to receive your crash reports. And if we don't know about them, we can't fix them."
```

not 100% happy with this one too but maybe you'll find it helpful 😅

sindresorhus:
```diff
-If you deny this connection, we won't be able to receive your crash reports. And if we don't know about them, we can't fix them.
+If you deny this connection, the developers of Gifski won't be notified of any crashes and they can't fix what they don't know about.
```

?

I feel like "your" implies that the user is actually sending the crash reports explicitly, which is not the case. I also think "we" is not clear enough in this context.

sunshinejr: @sindresorhus good point, looks better 👍

Issue #137, opened 2019-10-01
Make non-looping GIF


schneiderf:
Feature request: I would like to have the option to make non-looping gifs. Adobe Photoshop for example offers that option in its legacy "Save for web" exporter. Would be really great to see, thank you!

---
Gifski 2.0.1 (19)
macOS 10.14.6
iMacPro1,1

sindresorhus:
We should add a checkbox for "Infinite loop" and a text field (disabled by default) that when "Infinite loop" is unchecked, you can specify a loop count.

sunshinejr:
@sindresorhus I only found an option to add a flag `--once` to the gifski cli to remove looping the gif. Do you think we could start with just a checkbox if it's infinite or not? Loop count might be more time-consuming.

kornelski:
I expose only `once: bool` flag, because anything in between seems niche.

sindresorhus:
Yeah, lets go with only a checkbox. Loop count is indeed too niche.

PR #147, opened 2019-10-23
Add option for non-looping output

yonihemi:
Resolves https://github.com/sindresorhus/Gifski/issues/137

Note that QuickLook doesn't seem to respect iteration counts and loops the preview in any case.
Correct output can be verified with [ExifTool](https://www.sno.phy.queensu.ca/~phil/exiftool/), which should print `Animation Iterations : 1` when Infinite Loop option is disabled.

### DIFF
--- a/Gifski/Constants.swift
+++ b/Gifski/Constants.swift
@@ -12,6 +12,7 @@ extension NSColor {
 extension Defaults.Keys {
 	static let outputQuality = Key<Double>("outputQuality", default: 1)
 	static let successfulConversionsCount = Key<Int>("successfulConversionsCount", default: 0)
+	static let singleIteration = Key<Bool>("singleIteration", default: false)
 }
 
 struct Constants {
--- a/Gifski/EditVideoViewController.swift
+++ b/Gifski/EditVideoViewController.swift
@@ -21,6 +21,7 @@ final class EditVideoViewController: NSViewController {
 	@IBOutlet private var frameRateSlider: NSSlider!
 	@IBOutlet private var frameRateLabel: NSTextField!
 	@IBOutlet private var qualitySlider: NSSlider!
+	@IBOutlet private var loopButton: NSButton!
 
 	@IBOutlet private var widthTextField: IntTextField!
 	@IBOutlet private var heightTextField: IntTextField!
@@ -65,7 +66,8 @@ final class EditVideoViewController: NSViewController {
 			timeRange: timeRange,
 			quality: Defaults[.outputQuality],
 			dimensions: resizableDimensions.changed(dimensionsType: .pixels).currentDimensions.value,
-			frameRate: frameRateSlider.integerValue
+			frameRate: frameRateSlider.integerValue,
+			singleIteration: Defaults[.singleIteration]
 		)
 
 		let convert = ConversionViewController(conversion: conversion)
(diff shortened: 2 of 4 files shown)


sindresorhus:
@yonihemi It's recommended to do PRs from separate branches, not `master` ;)

Gifski/Constants.swift
```diff
@@ -12,6 +12,7 @@ extension NSColor {
 extension Defaults.Keys {
 	static let outputQuality = Key<Double>("outputQuality", default: 1)
 	static let successfulConversionsCount = Key<Int>("successfulConversionsCount", default: 0)
+	static let singleIteration = Key<Bool>("singleIteration", default: false)
```
sindresorhus: I don't think "iteration" is the correct term here. I would name it `loopGif`.

Gifski/EditVideoViewController.swift
```diff
@@ -248,12 +251,23 @@ final class EditVideoViewController: NSViewController {
 			self.estimateFileSize()
 		}
 
+		loopButton.onAction = { [weak self] _ in
```
sindresorhus: Instead of manually handling this, in Interface Builder, you can bind the checkbox directly to the userdefaults key.

Gifski/Gifski.swift
```diff
@@ -32,17 +32,19 @@ final class Gifski {
...
 		- Parameter frameRate: Clamped to 5...30. Uses the frame rate of `input` if not specified.
 		*/
-		init(video: URL, timeRange: ClosedRange<Double>? = nil, quality: Double = 1, dimensions: CGSize? = nil, frameRate: Int? = nil) {
+		init(video: URL, timeRange: ClosedRange<Double>? = nil, quality: Double = 1, dimensions: CGSize? = nil, frameRate: Int? = nil, singleIteration: Bool = false) {
 			self.video = video
 			self.timeRange = timeRange
 			self.quality = quality
 			self.dimensions = dimensions
 			self.frameRate = frameRate
+			self.singleIteration = singleIteration
```
sindresorhus: Can you add a short doc comment parameter description for this.

Gifski/EditVideoViewController.xib
```diff
@@ -14,6 +14,7 @@
                 <outlet property="frameRateLabel" destination="5Ke-gz-J2u" id="VaC-Et-vm7"/>
                 <outlet property="frameRateSlider" destination="Rej-Dl-z8Y" id="cIA-dv-KJw"/>
                 <outlet property="heightTextField" destination="bzf-LM-SxO" id="67N-9X-vzq"/>
+                <outlet property="loopButton" destination="Mk2-3o-6CT" id="lFW-e9-tIT"/>
```
sindresorhus: `loopButton` => `loopCheckbox`

sindresorhus:
For reference, here's how it looks like now:

Screenshot 2019-10-23 at 14 05 31

I think we can make it just `Infinite Loop` => `Loop`.

The "Estimated File Size" looks off somehow, but I can't exactly put my finger on it.

Gifski/EditVideoViewController.swift
yonihemi: Sure thing. I was trying to keep the same style as other properties' handling so changes will be focused on changed functionality.

kornelski:
"Loop forever"?

Gifski/ConversionCompletedViewController.swift
```diff
@@ -23,6 +23,13 @@ final class ConversionCompletedViewController: NSViewController {
 		maxWidth: 260
 	)
 
+	private lazy var copyTooltip = Tooltip(
```
sunshinejr: ah yeah due to the single branch the changes from the other PR got here too 😢

Gifski/EditVideoViewController.swift
sunshinejr: @sindresorhus wow didn't know that! great idea

Gifski/ConversionCompletedViewController.swift
yonihemi: My bad. Will fix this.

yonihemi:
Sorry, can't change the source branch so I've opened another pull request.

PR #148, opened 2019-10-23
Display visual feedback after clicking the “Copy” button

yonihemi:
Found it confusing to not have any feedback for Copy, so added a tooltip.

### DIFF
--- a/Gifski/ConversionCompletedViewController.swift
+++ b/Gifski/ConversionCompletedViewController.swift
@@ -23,13 +23,6 @@ final class ConversionCompletedViewController: NSViewController {
 		maxWidth: 260
 	)
 
-	private lazy var copyTooltip = Tooltip(
-		identifier: "copyCompletedTips",
-		text: "GIF copied to clipboard.",
-		showOnlyOnce: false,
-		maxWidth: 260
-	)
-
 	convenience init(conversion: Gifski.Conversion, gifUrl: URL) {
 		self.init()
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
I'm not sure about this. Why would you expect any feedback? You don't get any feedback if you press the "Edit => Copy" menu item in any app.

yonihemi:
You might be right, it could be me getting used to iOS-y feedback style, but there is a subtle indication in the flickering of the Edit menu which doesn't happen here (clicking button or pressing Cmd-C). Feel free to reject if this feels superfluous.

sindresorhus:
@yonihemi I'll need to think about this more.

Note to self: If we decide to go for this, the tooltip needs to automatically disappear after a set amount of seconds, maybe 4 seconds.

Gifski/ConversionCompletedViewController.swift
```diff
@@ -23,6 +23,13 @@ final class ConversionCompletedViewController: NSViewController {
 		maxWidth: 260
 	)
 
+	private lazy var copyTooltip = Tooltip(
+		identifier: "copyCompletedTips",
+		text: "GIF copied to clipboard.",
```
sindresorhus:
```suggestion
		text: "GIF copied to the clipboard",
```

sindresorhus:
An alternative approach:

sunshinejr:
Ah yeah some feedback on copying (either via button or command + c) would be great. Tooltip might be just a bit too much, but I like the idea of updating the "Copy" button for a second and go back to its original state.

yonihemi:
Agreed, this is more elegant. Changed implementation to the one suggested on the button itself.

PR #150, opened 2019-10-23
Present the open panel when clicking initial view

yonihemi:
Hits me Every. Single. Time.

### DIFF
--- a/Gifski/MainWindowController.swift
+++ b/Gifski/MainWindowController.swift
@@ -36,6 +36,9 @@ final class MainWindowController: NSWindowController {
 		window.makeKeyAndOrderFront(nil)
 
 		DockProgress.style = .circle(radius: 55, color: .themeColor)
+
+		let clickGestureRecognizer = NSClickGestureRecognizer(target: self, action: #selector(open(_:)))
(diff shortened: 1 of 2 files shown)


sindresorhus:
Should we indicate to the user that they can just click?

"Drop a Video" => "Drop a Video or Click Here" (Not really happy about it being more verbose though. Maybe it's not worth it)

sindresorhus:
This doesn't work, if I drag and drop a video, then click "Cancel", and then try to click "Drop a Video".

yonihemi:
@sindresorhus Fixed.

> Should we indicate to the user that they can just click?

IMO not worth it as will become too long. Silly people like me don't read anyway (hence the clicking).

sindresorhus:
I noticed an annoying issue with this now. It now presents the open panel when you drag the window around. I think it should only activate on an explicit click, not dragging. (Note that the user can click and drag anywhere in the window to drag the window).

sindresorhus:
@yonihemi Would you be able to fix the above? If not, I'll have to revert this PR.

sindresorhus:
Reverted

PR #151, opened 2019-10-23
Update IAP wording

sunshinejr:


Gifski/InternetAccessPolicy.json
```diff
@@ -8,7 +8,7 @@
 			"NetworkProtocol" : "TCP",
 			"Port" : "443",
 			"Purpose" : "Gifski sends crash reports to this server.",
-			"DenyConsequences" : "If you deny this connection, the developers of Gifski won't be notified of any crashes and they can't fix what they don't know about."
```
sunshinejr: I felt like the "and they can't fix what they don't know about" is a little too condescending, even though we don't have that intentions in mind. Not sure on the fix but I think we should update the sentence to be a little more nicer than this :D

sindresorhus: I agree. This sounds nicer.

sindresorhus:
> any crashes that happen on your device

This makes it sound like we would receive reports about **any** crashes, even ones not related to Gifski.

Gifski/InternetAccessPolicy.json
```diff
@@ -8,7 +8,7 @@
...
+			"DenyConsequences" : "If you deny this connection, the developers of Gifski won't be notified of any crashes that happen on your device and thus won't be able to fix it."
```
sindresorhus: How about this?

```suggestion
			"DenyConsequences" : "If you deny this connection, the Gifski developers won't be notified if the app crashes and thus won't be able to fix it."
```

sunshinejr: ah very nice 👍

PR #153, opened 2019-10-24
Add a simple welcome screen

sindresorhus:
I'm planning a much nicer welcome screen in the future using SwiftUI, but I want to add something like this for now as a lot users seem to not know about the "Send feedback" button and instead uses App Store reviews for bug reports, which is annoying as we can't really get enough information to fix it there.

Anything else we should mention in the welcome screen?

### DIFF
--- a/Gifski/MainWindowController.swift
+++ b/Gifski/MainWindowController.swift
@@ -17,7 +17,7 @@ final class MainWindowController: NSWindowController {
 		NSAlert.showModal(
 			for: window,
 			message: "Welcome to Gifski!",
-			informativeText: "If you have any feedback, bug reports, or feature requests, use the “Send Feedback” button in the “Help” menu. All submissions are answered by a human and bugs are quickly fixed.\n\nIt's better to submit a bug report through the feedback button than as an App Store review since we cannot contact you for more information in a review."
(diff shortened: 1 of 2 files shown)


Gifski/MainWindowController.swift
```diff
@@ -9,6 +9,18 @@ final class MainWindowController: NSWindowController {
...
 
+	private func showWelcomeScreen() {
+		guard App.isFirstLaunch else {
+			return
+		}
+
+		NSAlert.showModal(
+			for: window,
+			message: "Welcome to Gifski!",
+			informativeText: "If you have any feedback, bug reports, or feature requests, kindly use the “Send Feedback” button in the “Help” menu. We will respond to all submissions and it is likely that those reported issues will be fixed immediately.\n\nPlease note that it is preferable to submit a bug report through the feedback button than as an App Store review since the review will not allow us to contact you for more information regarding the bug."
```
sunshinejr:
```suggestion
			informativeText: "If you have any feedback, bug reports or feature requests, kindly use the “Send Feedback” button in the “Help” menu. We will respond to all submissions and it is likely that reported issues will be fixed immediately.\n\nPlease note that it is preferable to submit a bug report through the feedback button than as an App Store review since the review will not allow us to contact you for more information regarding the bug."
```

I think this is a very descriptive paragraph, nice job!

sindresorhus: I like the Oxford comma: https://www.grammarly.com/blog/what-is-the-oxford-comma-and-why-do-people-care-so-much-about-it/

sunshinejr: 👍

Issue #119, opened 2019-08-08
Error: "Error.generateFrameFailed"


sindresorhus:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2480%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/gifski-app/issues/119)

This is not a crash as we handle it, but it's presented to the user and still annoying. This is a very common error that we also had with earlier versions. 133 users have been affected by this in total. 😔

Error:

```
Gifski.Error.generateFrameFailed(Error Domain=AVFoundationErrorDomain Code=-11832 "Cannot Open" UserInfo={NSLocalizedFailureReason=This media cannot be used., NSLocalizedDescription=Cannot Open, NSUnderlyingError=0x60c00024f3f0 {Error Domain=NSOSStatusErrorDomain Code=-12431 "(null)"}})
```

Stacktrace:

```
# Version: 2.0.0 (18)
# OS Version: 10.13.6 (17G8030)
# Device: MacBook Pro 15", Late 2011

#0. Thread
0  Gifski                         0x10fe745ab CLSUserLoggingRecordError
1  Gifski                         0x10fe58c2f -[Crashlytics recordError:withAdditionalUserInfo:]
2  Gifski                         0x10fd80ed7 @objc AppDelegate.application(_:willPresentError:) (<compiler-generated>)
3  AppKit                         0x7fff51ce7cc3 (Missing)
4  AppKit                         0x7fff52408b5d (Missing)
5  Gifski                         0x10fd85400 closure #1 in closure #1 in ConversionViewController.start(conversion:) + 1930 (util.swift:1930)
6  Gifski                         0x10fd863f7 partial apply for closure #1 in closure #1 in ConversionViewController.start(conversion:) (<compiler-generated>)
7  Gifski                         0x10fd5fa26 closure #1 in closure #1 in static Gifski.run(_:completionHandler:) + 65 (Gifski.swift:65)
8  Gifski                         0x10fd62712 partial apply for closure #1 in closure #1 in static Gifski.run(_:completionHandler:) (<compiler-generated>)
```

Relevant lines:

https://github.com/sindresorhus/Gifski/blob/7285f880644fb83b33eee64ddccbd634535ab221/Gifski/Gifski.swift#L194-L196

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $80.00)

[...]

PR #155, opened 2019-10-24
Show a more actionable error message on "Frame failed" error

sindresorhus:
https://github.com/sindresorhus/Gifski/issues/119 is still an issue for a lot of users according to Crashlytics, but I have no idea how to reproduce it.

### DIFF
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -1,4 +1,4 @@
-import Foundation
+import Cocoa
 import AVFoundation
 import Crashlytics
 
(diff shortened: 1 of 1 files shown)


Gifski/Gifski.swift
```diff
@@ -201,7 +201,17 @@ final class Gifski {
 				case .failure where result.isCancelled:
 					completionHandlerOnce(.failure(.cancelled))
 				case let .failure(error):
-					completionHandlerOnce(.failure(.generateFrameFailed(error)))
+					// TODO: Remove this when we've been able to track down the issue.
+					completionHandlerOnce(.failure(.cancelled))
+					DispatchQueue.main.async {
+						NSAlert.showModal(
+							message: "Gifski was unable generate frames from the video",
+							informativeText: "We have been trying to track down this issue for a long time, but we have been unable to reproduce it. It would be awesome if you could send an email to sindresorhus@gmail.com with some information about the video file you tried to convert so we can fix this issue.",
```
sunshinejr:
```suggestion
							informativeText: "We have been trying to track down this issue for a long time, but we have been unable to reproduce it. It would be awesome if you could send an email to sindresorhus@gmail.com with the video or some information about the video file you tried to convert so we can fix this issue.",
```
do you think we could ask for the video specifically? or just info about it?

and yeah I think this may be an edge case on specific video files, wonder what it could be

sindresorhus: I got a lead in the iOS Slack channel about what it could be: https://github.com/sindresorhus/Gifski/issues/119#issuecomment-545843036

PR #156, opened 2019-10-30
Add share extension

### LINKED ISSUE #81: Share extension
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/gifski-app/issues/81)

I think it would be useful to create a [Share extension](https://developer.apple.com/design/human-interface-guidelines/macos/extensions/share-extensions/) for Gifski. That way, any video/screen recording app with a system "Share" button would be able to send a video to Gifski.

This would be especially useful for the built-in macOS screen recording feature. When you're done recording there, you get a thumbnail at the bottom-right of the screen. If you click it, a preview window opens, and in the toolbar, there's a "Share" button. So the user could then quickly send it to Gifski after recording without even having to save the video file.

Thoughts?

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $100.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($100.00)
### Submitted pull Requests
- [#156 Create Share extension](https://issuehunt.io/r/sindresorhus/gifski-app/pull/156)
---

### Tips

---
</details>

Ybrin:
Fixes #81 

I am using a "hack" by saving the shared video in an app container and opening the main application with the help of a WebView in the share extension view.    
The share extension is closed once Gifski is opened. This is a little bit ugly right now as you can see the empty share extension for a few moments. I don't see another workaround as the WebView needs some time to open Gifski.    
I added a simple CSS loading animation into the WebView so it doesn't look too ugly while loading, especially if the file shared is rather big and the copy process takes a few more moments.

The rest is handled by Gifski as it was always. The URL is opened and Gifski starts the Edit process.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#81: Share extension](https://issuehunt.io/repos/119822304/issues/81)
---

</details>


Ybrin:
@sindresorhus @sunshinejr

Ybrin:
Btw, I am using the hack because share extensions are actually not made for things like opening the main application. They are more like for messenger apps sharing posts or something...

I saw that most non-messenger apps solve it like that so I guess it should be fine. Even if it is technically a hack...

Share with Gifski/loadingHtml.html
```diff
@@ -0,0 +1,61 @@
+<html>
```
sindresorhus: I would prefer a more Mac-like spinner like this one: https://codepen.io/jmak/pen/LsCet

Share with Gifski/Share_with_Gifski.entitlements
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.app-sandbox</key>
+	<true/>
+	<key>com.apple.security.application-groups</key>
+	<array>
+		<string>$(TeamIdentifierPrefix).gifski_video_share_group</string>
```
sindresorhus:
```suggestion
		<string>$(TeamIdentifierPrefix).com.sindresorhus.Gifski.shareGroup</string>
```

?

Share with Gifski/Info.plist
```diff
@@ -0,0 +1,44 @@
...
+	<key>CFBundleVersion</key>
+	<string>1</string>
+	<key>LSMinimumSystemVersion</key>
+	<string>$(MACOSX_DEPLOYMENT_TARGET)</string>
+	<key>NSExtension</key>
+	<dict>
+		<key>NSExtensionAttributes</key>
+		<dict>
+			<key>NSExtensionActivationRule</key>
+			<string>TRUEPREDICATE</string>
```
sindresorhus: It should only be shown on https://github.com/sindresorhus/Gifski/blob/f68bd012b5322f43d6730e308f69071bfb85f24d/Gifski/Info.plist#L18-L20

Share with Gifski/Info.plist
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>AppIdentifierPrefix</key>
+	<string>$(AppIdentifierPrefix)</string>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>$(DEVELOPMENT_LANGUAGE)</string>
+	<key>CFBundleDisplayName</key>
+	<string>Share with Gifski</string>
```
sindresorhus:
```suggestion
	<string>Gifski</string>
```

Share with Gifski/loadingHtml.html
sindresorhus: Actually, why do you even need a web spinner? You can just overlay a native view over the webview with a native spinner.

Share with Gifski/ShareViewController.swift
```diff
@@ -0,0 +1,92 @@
...
+					}
+				} else if error != nil {
+					self.errorOpenMainApp()
+				}
+			}
+		}
+    }
+
+	private func errorOpenMainApp() {
+		if let gifski = URL(string: "gifski://error?shareExtension=true") {
```
sindresorhus: Why `error`? I assume it's because you need to specify a hostname. I would pick like `noop`.

Share with Gifski/ShareViewController.swift
```diff
@@ -0,0 +1,92 @@
...
+				if let url = url {
+					let shareUrl = "\(url.lastPathComponent)"
+
+					let appIdentifierPrefix = Bundle.main.infoDictionary!["AppIdentifierPrefix"] as! String
+					let appGroupShareVideUrl = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: "\(appIdentifierPrefix).gifski_video_share_group")?.appendingPathComponent(shareUrl)
+
+					try? FileManager.default.removeItem(at: appGroupShareVideUrl!)
+					try! FileManager.default.copyItem(at: url, to: appGroupShareVideUrl!)
+
+					if let gifski = URL(string: "gifski://\(shareUrl.addingPercentEncoding(withAllowedCharacters: .urlHostAllowed)!)?shareExtension=true") {
```
sindresorhus: I think the share URL should be a query parameter, not as a hostname.

Share with Gifski/ShareViewController.swift
```diff
@@ -0,0 +1,92 @@
...
+			typeIdentifier = "com.apple.m4v-video"
+		} else if item.hasItemConformingToTypeIdentifier("com.apple.quicktime-movie") {
+			typeIdentifier = "com.apple.quicktime-movie"
+		} else {
+			errorOpenMainApp()
+			return
+		}
+
+		item.loadFileRepresentation(forTypeIdentifier: typeIdentifier) { url, error in
+			DispatchQueue.main.sync {
```
sindresorhus: Only the webview stuff needs to be wrapped in this.

Share with Gifski/ShareViewController.swift
```diff
@@ -0,0 +1,92 @@
...
+		} else if item.hasItemConformingToTypeIdentifier("com.apple.m4v-video") {
+			typeIdentifier = "com.apple.m4v-video"
+		} else if item.hasItemConformingToTypeIdentifier("com.apple.quicktime-movie") {
+			typeIdentifier = "com.apple.quicktime-movie"
+		} else {
+			errorOpenMainApp()
+			return
+		}
+
+		item.loadFileRepresentation(forTypeIdentifier: typeIdentifier) { url, error in
```
sindresorhus: Use `guard` for the error here.

Share with Gifski/ShareViewController.swift
sindresorhus: Where is the error handled in the main app?

Share with Gifski/ShareViewController.swift
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+	private func errorOpenMainApp() {
+		if let gifski = URL(string: "gifski://error?shareExtension=true") {
+			let request = URLRequest(url: gifski)
+			self.webView.navigationDelegate = self
+			self.webView.load(request)
+		}
+
+		DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
```
sindresorhus: Can you add a comment why this needs to be delayed?

sindresorhus:
Why do you need the web view? There's https://developer.apple.com/documentation/foundation/nsextensioncontext/1416791-open

Share with Gifski/Share_with_Gifski.entitlements
Ybrin: Apple says it should be `$(TeamIdentifierPrefix).<name>` for macOS apps. I can rename the group to `shareGroup` though if you prefer.

Reference: https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_application-groups

Ybrin:
@sindresorhus The open function of `NSExtensionContext` cannot be used from a share extension. It's stated in the docs that only some extensions are allowed to use it. That's the reason for the WebView workaround

Share with Gifski/Share_with_Gifski.entitlements
sindresorhus: Ah, ok. Leave it as is then.

sindresorhus:
Did you try the `performSelector` workarounds mentioned in https://stackoverflow.com/questions/24297273/openurl-not-work-in-action-extension and https://stackoverflow.com/questions/25727489/how-to-launch-a-parent-ios-app-from-its-app-extension ?

Ybrin:
Ok `NSWorkspace.shared.open` works...i will delete the WebView stuff then...

Share with Gifski/ShareViewController.swift
Ybrin: Done

Share with Gifski/ShareViewController.swift
Ybrin: Are you sure that FileManager stuff is thread safe? If yes I will move it out.

Share with Gifski/ShareViewController.swift
Ybrin: Done

Share with Gifski/ShareViewController.swift
Ybrin: Done

Share with Gifski/ShareViewController.swift
Ybrin: Yes it is, nvm. I just checked it...

Ybrin: Done

Share with Gifski/Info.plist
Ybrin: Done

Ybrin:
@sindresorhus Done. I guess I got everything you mentioned.

sindresorhus:
Can you mention the feature in the readme here: https://github.com/sindresorhus/Gifski#features

And also a quick guide on how to use it for screen recording purposes? As described in https://github.com/sindresorhus/Gifski/issues/81. Maybe in a `Tips` section.

Ybrin:
@sindresorhus Done. I included the tip for the screen recordings directly into the features section. This is quite on the top of the readme, which is a great place for this feature.

Gifski.xcodeproj/project.pbxproj
```diff
@@ -436,6 +541,65 @@
...
+				DEVELOPMENT_TEAM = YG56YK5RN5;
+				ENABLE_HARDENED_RUNTIME = YES;
+				GCC_C_LANGUAGE_STANDARD = gnu11;
+				INFOPLIST_FILE = "Share with Gifski/Info.plist";
+				LD_RUNPATH_SEARCH_PATHS = (
+					"$(inherited)",
+					"@executable_path/../Frameworks",
+					"@executable_path/../../../../Frameworks",
+				);
+				MACOSX_DEPLOYMENT_TARGET = 10.14;
```
sindresorhus: Don't set the deployment target explicitly. This is the wrong target.

Share with Gifski/ShareViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+		} else if item.hasItemConformingToTypeIdentifier("com.apple.m4v-video") {
+			typeIdentifier = "com.apple.m4v-video"
+		} else if item.hasItemConformingToTypeIdentifier("com.apple.quicktime-movie") {
+			typeIdentifier = "com.apple.quicktime-movie"
+		} else {
+			errorOpenMainApp()
+			return
+		}
+
+		item.loadFileRepresentation(forTypeIdentifier: typeIdentifier) { url, _ in
```
sindresorhus: Why not use `loadDataRepresentation` since we copy it later on anyway?

Share with Gifski/ShareViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+
+			let appIdentifierPrefix = Bundle.main.infoDictionary!["AppIdentifierPrefix"] as! String
+			let appGroupShareVideUrl = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: "\(appIdentifierPrefix).gifski_video_share_group")?.appendingPathComponent(shareUrl)
+
+			try? FileManager.default.removeItem(at: appGroupShareVideUrl!)
+			try! FileManager.default.copyItem(at: url, to: appGroupShareVideUrl!)
+
+			let gifski = URL(string: "gifski://shareExtension?path=\(shareUrl.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)!)")!
+
+			DispatchQueue.main.sync {
```
sindresorhus: Does it really have to be `sync`?

Share with Gifski/ShareViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+				NSWorkspace.shared.open(gifski)
+				self.extensionContext?.completeRequest(returningItems: [], completionHandler: nil)
+			}
+		}
+    }
```
sindresorhus: Wrong indentation.

Share with Gifski/Info.plist
```diff
@@ -0,0 +1,54 @@
...
+			<string>SUBQUERY (
+    extensionItems,
+    $extensionItem,
+    SUBQUERY (
+        $extensionItem.attachments,
+        $attachment,
+        ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO &quot;public.mpeg-4&quot; ||
+        ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO &quot;com.apple.m4v-video&quot; ||
+        ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO &quot;com.apple.quicktime-movie&quot;
+    ).@count == $extensionItem.attachments.@count
```
sindresorhus: Use tab-indentation.

Share with Gifski/ShareViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+
+		var typeIdentifier: String
+		if item.hasItemConformingToTypeIdentifier("public.mpeg-4") {
+			typeIdentifier = "public.mpeg-4"
+		} else if item.hasItemConformingToTypeIdentifier("com.apple.m4v-video") {
+			typeIdentifier = "com.apple.m4v-video"
+		} else if item.hasItemConformingToTypeIdentifier("com.apple.quicktime-movie") {
+			typeIdentifier = "com.apple.quicktime-movie"
+		} else {
+			errorOpenMainApp()
```
sindresorhus: I think this should pass a more useful error to the main app. Instead of making the `error` query parameter a bool, just make it the error message.

Share with Gifski/ShareViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+			return
+		}
+
+		item.loadFileRepresentation(forTypeIdentifier: typeIdentifier) { url, _ in
+			guard let url = url else {
+				self.errorOpenMainApp()
```
sindresorhus: Same here. More useful message.

Side note: Are you sure the share extension cannot present the error message itself?

Share with Gifski/ShareViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+			let gifski = URL(string: "gifski://shareExtension?path=\(shareUrl.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)!)")!
+
+			DispatchQueue.main.sync {
+				NSWorkspace.shared.open(gifski)
+				self.extensionContext?.completeRequest(returningItems: [], completionHandler: nil)
+			}
+		}
+    }
+
+	private func errorOpenMainApp() {
```
sindresorhus:
```suggestion
	private func openMainAppAndPresentError() {
```

Share with Gifski/ShareViewController.swift
```diff
@@ -0,0 +1,53 @@
...
+				self.errorOpenMainApp()
+				return
+			}
+
+			let shareUrl = "\(url.lastPathComponent)"
+			let appIdentifierPrefix = Bundle.main.infoDictionary!["AppIdentifierPrefix"] as! String
+			let appGroupShareVideUrl = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: "\(appIdentifierPrefix).gifski_video_share_group")?.appendingPathComponent(shareUrl)
+
+			try? FileManager.default.removeItem(at: appGroupShareVideUrl!)
+			try! FileManager.default.copyItem(at: url, to: appGroupShareVideUrl!)
```
sindresorhus: Too much force-unwrapping. You can not be sure all of this is going to work.

Share with Gifski/ShareViewController.swift
```diff
@@ -0,0 +1,53 @@
...
+
+			let gifski = URL(string: "gifski://shareExtension?path=\(shareUrl.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)!)")!
```
sindresorhus: There are nicer ways to add a query to a URL than string interpolation.

Share with Gifski/Info.plist
```diff
@@ -0,0 +1,52 @@
...
+	<key>AppIdentifierPrefix</key>
+	<string>$(AppIdentifierPrefix)</string>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>$(DEVELOPMENT_LANGUAGE)</string>
+	<key>CFBundleDisplayName</key>
+	<string>Gifski</string>
+	<key>CFBundleExecutable</key>
+	<string>$(EXECUTABLE_NAME)</string>
+	<key>CFBundleIconFile</key>
+	<string>icon</string>
```
sindresorhus: I don't think we need this or the bundled generic icon. You can just remove those.

Gifski/AppDelegate.swift
```diff
@@ -54,13 +54,29 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 			return
 		}
 
+		var sharedVideoUrl = videoUrl
+
+		if videoUrl.host == "shareExtension" {
+			if let path = videoUrl.queryParameters["path"] {
+				let appIdentifierPrefix = Bundle.main.infoDictionary!["AppIdentifierPrefix"] as! String
+				let appGroupShareVideUrl = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: "\(appIdentifierPrefix).gifski_video_share_group")?.appendingPathComponent(path.removingPercentEncoding!)
+				sharedVideoUrl = appGroupShareVideUrl!
```
sindresorhus: Typo and too much force-unwrapping.

Gifski.xcodeproj/project.pbxproj
```diff
@@ -436,6 +541,65 @@
...
+				MTL_FAST_MATH = YES;
+				PRODUCT_BUNDLE_IDENTIFIER = "com.sindresorhus.Gifski.Share-with-Gifski";
```
sindresorhus:
```suggestion
				PRODUCT_BUNDLE_IDENTIFIER = "com.sindresorhus.Gifski.ShareExtension";
```

And other places it's used.

Gifski.xcodeproj/project.pbxproj
```diff
@@ -436,6 +541,65 @@
...
+				CLANG_ENABLE_OBJC_WEAK = YES;
+				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
+				CODE_SIGN_ENTITLEMENTS = "Share with Gifski/Share_with_Gifski.entitlements";
+				CODE_SIGN_IDENTITY = "Mac Developer";
+				CODE_SIGN_STYLE = Automatic;
+				COMBINE_HIDPI_IMAGES = YES;
+				DEVELOPMENT_TEAM = YG56YK5RN5;
+				ENABLE_HARDENED_RUNTIME = YES;
+				GCC_C_LANGUAGE_STANDARD = gnu11;
+				INFOPLIST_FILE = "Share with Gifski/Info.plist";
```
sindresorhus:
```suggestion
				INFOPLIST_FILE = "ShareExtension/Info.plist";
```

And other places it's used.

Share with Gifski/Share_with_Gifski.entitlements
```diff
@@ -0,0 +1,16 @@
...
+	</array>
+	<key>com.apple.security.files.user-selected.read-only</key>
+	<true/>
+	<key>com.apple.security.network.client</key>
```
sindresorhus: I don't think this is needed now that we don't use WebKit anymore?

Share with Gifski/Share_with_Gifski.entitlements
```diff
@@ -0,0 +1,16 @@
...
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.app-sandbox</key>
+	<true/>
+	<key>com.apple.security.application-groups</key>
+	<array>
+		<string>$(TeamIdentifierPrefix).gifski_video_share_group</string>
+	</array>
+	<key>com.apple.security.files.user-selected.read-only</key>
```
sindresorhus: Is this really needed?

Share with Gifski/ShareViewController.swift
sindresorhus: Actually, ignore this. Extensions have limited memory, so this is not a good idea.

Gifski.xcodeproj/project.pbxproj
Ybrin: I changed it to 10.13 just like the main Gifski app.

Share with Gifski/ShareViewController.swift
Ybrin: Sorry, I tested it multiple times and the docs also explicitly state it. Those two calls have to be on the main thread. If not, the share extension crashes randomly sometimes.

Share with Gifski/ShareViewController.swift
Ybrin: This seems resolved in your last commit...

Share with Gifski/Info.plist
Ybrin: Done

Share with Gifski/ShareViewController.swift
Ybrin: It seems better to me to handle all errors in one place and not some in the main app and some in the extension. I will pass the error message though.

Share with Gifski/ShareViewController.swift
Ybrin: Done

Share with Gifski/ShareViewController.swift
Ybrin: Actually it's well documented in apples libraries. `appGroupShareVideUrl ` can never be nil as it will never change. `FileManager.default.copyItem ` will only throw if either the destination file exists, which is impossible because of the line before, or you don't have permissions, which is impossible in an app group.

But I changed it to handle this case.

Share with Gifski/Info.plist
Ybrin: Done

Gifski/AppDelegate.swift
Ybrin: Done

Gifski.xcodeproj/project.pbxproj
Ybrin: Done

Share with Gifski/ShareViewController.swift
sindresorhus: Why is it better? Generally, it's better to show an error as close to the context/source as possible. If it is possible to show an error alert in the share extension, I would prefer that.

Share with Gifski/ShareViewController.swift
```diff
@@ -7,7 +7,7 @@ final class ShareViewController: NSViewController {
 		super.loadView()
 
 		guard let item = (self.extensionContext?.inputItems[0] as? NSExtensionItem)?.attachments?.first else {
-			errorOpenMainApp()
+			errorOpenMainApp(message: "Shared item does not contain an attachment")
```
sindresorhus: `Shared` => `The shared`

Gifski.xcodeproj/project.pbxproj
Ybrin: Done

Share with Gifski/Share_with_Gifski.entitlements
Ybrin: Removed it

Ybrin: Yes it doesn't need the network client entitlement. I tested it.

Share with Gifski/Share_with_Gifski.entitlements
Ybrin: Seems like it works without it. I removed it.

Share with Gifski/ShareViewController.swift
Ybrin: Done

Share with Gifski/ShareViewController.swift
Ybrin: I made it with URLComponents now.

Ybrin:
@sindresorhus I am done with your suggestions. I think I didn't forget anything.

Share with Gifski/ShareViewController.swift
Ybrin: I don't see how actually. I tried it initially and now again. We don't have access to a window in an extension as it runs in the context of another app and `runModal()` crashes.
Maybe some kind of custom view embedded into the alert view which shows the message and an OK button?

sindresorhus:
Regarding https://github.com/sindresorhus/Gifski/pull/156#discussion_r344441606, you could just present a simple error view with a text field and a OK button in `ShareViewController`.

ShareExtension/ShareViewController.swift
```diff
@@ -0,0 +1,76 @@
+import Cocoa
+
+final class ShareViewController: NSViewController {
+	override var nibName: NSNib.Name? { NSNib.Name("ShareViewController") }
+
+	override func loadView() {
```
sindresorhus: There's a SwiftLint violation here. I think it's a false-positive, so you need to add an ignore comment.

ShareExtension/ShareViewController.swift
```diff
@@ -0,0 +1,76 @@
+import Cocoa
+
+final class ShareViewController: NSViewController {
+	override var nibName: NSNib.Name? { NSNib.Name("ShareViewController") }
```
sindresorhus:
```suggestion
	override var nibName: NSNib.Name? { "ShareViewController" }
```

ShareExtension/ShareViewController.swift
Ybrin: Done

Ybrin:
@sindresorhus I changed it. The error will now be shown directly in the extension view and not redirected to the main app.

sindresorhus:
Looks good to me now. Really nice work, @Ybrin 👍

PR #165, opened 2019-12-02
Handle video track duration being shorter than total duration

### LINKED ISSUE #119: Error: "Error.generateFrameFailed"

### LINKED ISSUE #164: Crashes when almost complete
So here is the video I'm trying to convert.  The video was taken with a Samsung phone running Android 4.4,; I have no idea if it matters? 
Action taken was reduce to 25%, lowered the quality, FPS was 15.  The conversion stopped at 96% displayed.

My first video that failed was to large to email , so, I have made a shorter video that also failed with the same error message, for your reference the filename is Gifski.mp4: [Gifski.mp4.zip](https://github.com/sindresorhus/Gifski/files/3909493/Gifski.mp4.zip)

---
Gifski 2.3.0 (26) - com.sindresorhus.Gifski
macOS 10.13.6
iMac10,1

sindresorhus:
Fixes #119
Fixes #164

Someone finally submitted a video that reproduces the issue and I discovered why it's happening. Apparently, a video file can contain a video track that is shorter than the total file duration. We generated frame time points for the whole asset, even outside the video track. That's why `generateCGImagesAsynchronously` errored. We tried to get a frame from the video when there were none.

### DIFF
--- a/Gifski/VideoValidator.swift
+++ b/Gifski/VideoValidator.swift
@@ -57,7 +57,10 @@ struct VideoValidator {
 		}
 
 		// We already specify the UTIs we support, so this can only happen on invalid video files or unsupported codecs.
-		guard asset.isVideoDecodable else {
+		guard
+			asset.isVideoDecodable,
+			let firstVideoTrack = asset.firstVideoTrack
+		else {
 			NSAlert.showModalAndReportToCrashlytics(
 				for: window,
 				message: "The video file is not supported.",
@@ -91,6 +94,25 @@ struct VideoValidator {
 			)
 
 			return .failure
+		}
+
+		// If the video track duration is shorter than the total asset duration, we extract the video track into a new asset to prevent problems later on.
+		guard firstVideoTrack.isFullDuration else {
+			guard
+				let newAsset = firstVideoTrack.extractToNewAsset(),
+				let newVideoMetadata = newAsset.videoMetadata
+			else {
+				NSAlert.showModalAndReportToCrashlytics(
+					for: window,
+					message: "Cannot read the video.",
+					informativeText: "Please open an issue on https://github.com/sindresorhus/Gifski or email sindresorhus@gmail.com. ZIP the video and attach it.\n\nInclude this info:",
+					debugInfo: asset.debugInfo
+				)
+
+				return .failure
+			}
+
+			return .success(newAsset, newVideoMetadata)
 		}
 
 		return .success(asset, videoMetadata)
--- a/Gifski/VideoValidator.swift
+++ b/Gifski/VideoValidator.swift
@@ -96,7 +96,7 @@ struct VideoValidator {
 			return .failure
 		}
 
-		// If the video track duration is shorter than the total asset duration, we extract the video track into a new asset to prevent problems later on.
+		// If the video track duration is shorter than the total asset duration, we extract the video track into a new asset to prevent problems later on. If we don't do this, the video will show as black in the trim view at the duration where there's no video track, and it will confuse users. Also, if the user trims the video to just the black no video track part, the conversion would continue, but there's nothing to convert, so it would be stuck at 0%.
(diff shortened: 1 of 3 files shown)


Gifski/VideoValidator.swift
```diff
@@ -93,6 +96,25 @@ struct VideoValidator {
 			return .failure
 		}
 
+		// If the video track duration is shorter than the total asset duration, we extract the video track into a new asset to prevent problems later on.
+		guard firstVideoTrack.isFullDuration else {
```
sunshinejr: do we still need to do this after you changed the code to base on the track instead of the asset?

sunshinejr:
Ah this is a very good find! I'm wondering - can we actually create a video with multiple video tracks?

Gifski/VideoValidator.swift
sindresorhus: Yes, because the part of the video where there's no video track will still show in the trimmer view if not, which might confuse users, and also if the user trims it to only that part, then the conversion will start, but nothing will happen as there's nothing to convert.

sindresorhus: I've expanded the comment about it: https://github.com/sindresorhus/Gifski/pull/165/commits/ebe1e7e58c01127ddb46467d1b6a381f0194ef11

sindresorhus:
> can we actually create a video with multiple video tracks?

Yes, it's technically [supported by MP4](https://video.stackexchange.com/a/20937/1114), but not very common, and most apps don't handle it, and we definitely shouldn't care for Gifski.

sunshinejr:
@sindresorhus ah makes sense 👍

PR #167, opened 2019-12-03
Add “Open” button to the drop view

### LINKED ISSUE #158: Click the initial drop view to present the open panel
See https://github.com/sindresorhus/Gifski/pull/150

sindresorhus:
On macOS 10.15 only.

#### Before

Screenshot 2019-12-03 at 14 14 12

#### After

Screenshot 2019-12-03 at 13 53 29

---

Fixes #158

### DIFF
--- a/Gifski/VideoDropView.swift
+++ b/Gifski/VideoDropView.swift
@@ -14,11 +14,7 @@ class DropView<CompletionType>: SSView {
 		}
 	}
 
-	var dropView: NSView? {
-		didSet {
-			dropView?.isHidden = dropView == nil
-		}
-	}
+	var dropView: NSView?
 
 	private let dropLabel = with(Label()) {
 		$0.textColor = .secondaryLabelColor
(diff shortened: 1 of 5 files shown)


Gifski/VideoDropView.swift
```diff
@@ -14,6 +14,12 @@ class DropView<CompletionType>: SSView {
 		}
 	}
 
+	var dropView: NSView? {
+		didSet {
+			dropView?.isHidden = dropView == nil
```
sunshinejr: this is really interesting! but isn't it effectively the same as `dropView?.isHidden = false`?

Gifski/VideoDropViewController.swift
```diff
@@ -1,10 +1,35 @@
 import AppKit
+import SwiftUI
+
+@available(macOS 10.15, *)
+struct DropCenterView: View {
+	var body: some View {
+		VStack {
```
sunshinejr: 👌

Gifski/VideoDropView.swift
sindresorhus: You're probably right. I'm just gonna drop it as we don't really need that handling at all. We only set `dropView` once. I did it for the label as back then I thought it was smart to make the drop view as reusable as possible, but that was honestly a mistake. It's just made it harder to customize it.

PR #173, opened 2020-01-29
Support 60 FPS

### LINKED ISSUE #161: Allow 60 FPS?
I've gotten a couple of requests for 60 FPS. We currently limit it to 30 FPS. I do realize it doesn't even usually make sense with 30 FPS for GIF, but their use-case is very short clips to upload to https://dribbble.com.

Thoughts?

sindresorhus:
Fixes #161

@kornelski If the user sets 60 FPS, it will just use 50 FPS, right? Or do I need to clamp the actual value passed to gifski-lib at 50 FPS?

### DIFF
--- a/Gifski/Constants.swift
+++ b/Gifski/Constants.swift
@@ -1,6 +1,12 @@
 import Cocoa
 import Defaults
 
+struct Constants {
+	static let defaultWindowSize = CGSize(width: 360, height: 240)
+	static let backgroundImage = NSImage(named: "BackgroundImage")!
+	static let allowedFrameRate = 5.0...60.0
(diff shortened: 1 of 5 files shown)


kornelski:
gifski doesn't take fps, but delay that is multiple of 10ms (per GIF spec), and it will store the value you give as-is. 

But if you specify delay=1 (floor(100/60)), it'll mean 100fps, and browsers will ignore it.

sindresorhus:
Done. I've confirmed that the delay is `2`.

KartikChugh:
My slider only goes up to 30 fps for some reason

PR #175, opened 2020-01-30
Update libgifski

### LINKED ISSUE #174: Update lib gif.ski
In 0.10 I've changed interface from taking frame delay, to taking absolute time in seconds (float) when the frame should be presented (aka PTS). This makes the library manage delays internally and it supports arbitrary framerates.

If you pass PTS straight from the codec, it'll give smoothest framerate. See how to limit fps in a PTS-friendly way: https://github.com/ImageOptim/gifski/commit/f1e8b59039999c7a577f3c437e4e7c61a4780f4c#diff-fbfbc756b97d21883f66ee9308049c58R79

cc #173

sindresorhus:
Fixes #174

### DIFF
--- a/gifski-api/src/encodegifsicle.rs
+++ b/gifski-api/src/encodegifsicle.rs
@@ -76,7 +76,7 @@ impl Encoder for Gifsicle<'_> {
             };
             gfs.screen_width = image.width() as _;
             gfs.screen_height = image.height() as _;
-            gfs.loopcount = if settings.once { 1 } else { -1 };
+            gfs.loopcount = if settings.once { 1 } else { 0 }; // 0 is correct, -1 is not
             unsafe {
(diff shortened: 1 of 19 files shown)


gifski-api/src/encodegifsicle.rs
```diff
@@ -76,7 +76,7 @@ impl Encoder for Gifsicle<'_> {
             };
             gfs.screen_width = image.width() as _;
             gfs.screen_height = image.height() as _;
-            gfs.loopcount = if settings.once {1} else {0}; // 0 means infinite
+            gfs.loopcount = if settings.once { 1 } else { -1 };
```
sindresorhus: @kornelski The change from `0` to `-1` was intentional, right?

kornelski: Good catch. That's a regression. I didn't backport the fix from the app. Fixed in master now.

PR #176, opened 2020-02-02
Fix stuttering

sindresorhus:
Some users have reported that the GIFs they produce with Gifski 2.6.0 sometimes have stuttering. It turns out that I had some misconceptions about the `result.actualTime` behavior.

Copy of my comment from the code (which I have written for posterity):

> Even if you specify `.requestedTimeToleranceAfter = .zero` and `.requestedTimeToleranceBefore = .zero`, it's not guaranteed that the frame will be generated at the exact given times, or that `result.requestedTime` and `result.actualTime` will be the same. If there are not enough frames in the source video, `generateCGImagesAsynchronously` might produce duplicate frames and `result.actualTime` will be the time of the previous frame, hence not the same as `result.requestedTime`. If you're generating some kind of video/animation from the frames, use `result.requestedTime` and not `result.actualTime` for the presentation timestamp.

Of course none of this behavior is documented by Apple.

### DIFF
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -222,6 +222,9 @@ final class Gifski {
 				value: frameForTimes.map { $0.seconds }
 			)
 
+			var frameNumber = 0
+			var previousActualTime: CMTime?
+
 			generator.generateCGImagesAsynchronously(forTimePoints: frameForTimes) { [weak self] result in
 				guard let self = self else {
 					return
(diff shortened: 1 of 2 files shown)


kornelski:
Skipping of frames with identical PTS is good. 

But I think using `requestedTime` instead of `actualTime` is wrong. I don't agree with Apple's advice here. `requestedTime` is taken from wishful thinking from the request, not actual video timing.

sindresorhus:
> I don't agree with Apple's advice here.

It's not Apple's advice. It's my advice to myself based on my findings.

kornelski:
https://github.com/ImageOptim/gifski/commit/1e763a0d0c058c6f5a97c19560b404fd92d63081

sindresorhus:
> requestedTime is taken from wishful thinking from the request, not actual video timing.

Good point. I changed to use `requestedTime` initially before doing the duplicate frame skipping thing. Since we're skipping duplicate frames, it should be fine to use `actualTime`.

sindresorhus:
> ImageOptim/gifski@1e763a0

Awesome. I think this behavior should be documented in the doc comments so implementers know they can just let libgifski handle the frame deduplication.

sindresorhus:
For posterity, here's the sample video that ended up with stuttering: [6k3rx.mp4.zip](https://github.com/sindresorhus/Gifski/files/4144323/6k3rx.mp4.zip) (set 30 FPS in the editor)

kornelski:
works fine in the CLI version which uses actual PTS: 30fps

sindresorhus:
Closing this in favor of https://github.com/sindresorhus/Gifski/commit/15db51b06ceac0db2b89e1341568d79f115ba975.

Hopefully, the libgifski changes completely fixed this now.

PR #177, opened 2020-02-05
Fix trimming from the start

sindresorhus:
I got a report that the first frame froze for a long time on their GIF. After a lot of debugging, it turns out after the 60 FPS changes, I forgot to adjust the presentation timestamp passed to libgifski when the video is trimmed.

### DIFF
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -253,7 +253,7 @@ final class Gifski {
 							bytesPerRow: UInt32(image.bytesPerRow),
 							height: UInt32(image.height),
 							pixels: buffer,
-							presentationTimestamp: result.actualTime.seconds
+							presentationTimestamp: max(0, result.actualTime.seconds - startTime)
 						)
 					} catch {
 						completionHandlerOnce(.failure(.addFrameFailed(error)))


Gifski/Gifski.swift
```diff
@@ -253,7 +253,7 @@ final class Gifski {
 							bytesPerRow: UInt32(image.bytesPerRow),
 							height: UInt32(image.height),
 							pixels: buffer,
-							presentationTimestamp: result.actualTime.seconds
+							presentationTimestamp: max(0, result.actualTime.seconds - startTime)
```
sindresorhus: @kornelski `result.actualTime.seconds - startTime` is often a negative value as the actual time might be before our requested time, hence before the startTime. Does libgifski handle negative PTS? Should I just pass it negative to libgifski or is it correct to handle it here? If passing it to libgifski works, I assume it will just throw it away, but wouldn't that mean, we'll lose a frame?

Example value: `-0.012468888888889751`.

sindresorhus:
I've got confirmation from the reporter that this does fix the problem.

Gifski/Gifski.swift
kornelski: It doesn't handle negative PTS.

sindresorhus:
Thanks for reviewing  🙌

PR #178, opened 2020-02-06
Improve the legibility of the progress circle in dark mode

sindresorhus:
We use the user's accent color for the progress circle. Because the window is vibrant, the progress circle is not very legible in dark mode:

Screenshot 2020-02-06 at 15 52 40

I tried making the color brighter, but I couldn't make it look very good.

I suggest we make it just white in dark mode. This looks pretty good:

Screenshot 2020-02-06 at 15 39 43

I've also changed the line width from 2 to 3.

The question then is what should we do in light mode?

The blue accent color looks fine:

Screenshot 2020-02-06 at 15 54 29

But if the user (for some crazy reason) has orange as accent color, it becomes less legible:

Screenshot 2020-02-06 at 15 48 45

And this is how using `.labelColor`, which is used in dark mode, would look in light mode:

Screenshot 2020-02-06 at 15 37 54

Looks pretty good to me.

**So the question is, should light mode use accent color or just label color like dark mode**? I'm leaning towards the latter.

### DIFF
--- a/Gifski/Assets.xcassets/ProgressCircleColor.colorset/Contents.json
+++ b/Gifski/Assets.xcassets/ProgressCircleColor.colorset/Contents.json
@@ -1,0 +1,28 @@
+{
+  "info" : {
+    "version" : 1,
+    "author" : "xcode"
+  },
+  "colors" : [
+    {
+      "idiom" : "mac",
+      "color" : {
+        "platform" : "osx",
+        "reference" : "controlAccentColor"
+      }
+    },
+    {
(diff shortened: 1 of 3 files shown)


sunshinejr:
@sindresorhus imo using `accentColor` is a pretty nice touch that makes the app more personalized. On the other hand I also like the black/white (`labelColor`), probably even better, but I wonder if it's not gonna be too "boring".

sindresorhus:
Alright, let's keep the accent color in light mode.

PR #179, opened 2020-02-06
Loop the playback when “Loop Forever” is enabled

sindresorhus:
I tried using [`AVPlayerLooper`](https://developer.apple.com/documentation/avfoundation/avplayerlooper) first, but that removes the editing controls when looping. The looping seems smooth using the current method anyway.

### DIFF
--- a/Gifski/EditVideoViewController.swift
+++ b/Gifski/EditVideoViewController.swift
@@ -302,6 +302,12 @@ final class EditVideoViewController: NSViewController {
 		playerViewController = TrimmingAVPlayerViewController(playerItem: AVPlayerItem(asset: asset)) { [weak self] _ in
 			self?.estimateFileSize()
 		}
+
+		Defaults.observe(.loopGif) { [weak self] in
+			self?.playerViewController.loopPlayback = $0.newValue
+		}
+			.tieToLifetime(of: self)
+
 		add(childController: playerViewController, to: playerViewWrapper)
 	}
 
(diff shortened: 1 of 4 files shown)


kornelski:
Have you tested with airplay? Hopefully without audio macOS won't give it the lag, but videos with audio have a 2-second delay on seek.

sindresorhus:
Just tested now. When it plays to the end, it pauses for 2 seconds to make the audio catch up, before beginning from the start again. I also tested with `AVPlayerLooper` and same result.

Do we even want audio when playing there at all? I don't really see the point personally. So one fix could be to just remove all audio tracks before presenting the editor.

sindresorhus:
I tried https://developer.apple.com/documentation/avfoundation/avplayer/1387441-allowsexternalplayback, but it seems to have no impact on audio.

kornelski:
Removing audio from preview would make a lot of sense. I'm not familiar with the APIs to advise how.

sindresorhus:
I did find a way to force the audio to play just in the editor. Using https://developer.apple.com/documentation/avfoundation/avplayer/1390717-audiooutputdeviceuniqueid. But it's a nightmare to find the built-in output device ID (thousands of lines of low-level code).

Let's go with removing the audio.

sindresorhus:
Audio removed.

sindresorhus:
I'm gonna manually test this a bit more before merging. I want to ensure there are no regressions.

PR #181, opened 2020-02-09
Fix some memory leaks

sindresorhus:
I noticed some tiny memory leaks when debugging.

### DIFF
--- a/Gifski/TimeRemainingEstimator.swift
+++ b/Gifski/TimeRemainingEstimator.swift
@@ -10,13 +10,21 @@ final class TimeRemainingEstimator {
 	/// Begin fade out when remaining time reaches this amount.
 	let fadeOutThreshold: TimeInterval = 1
 
-	var progress: Progress? {
+	weak var progress: Progress? {
 		didSet {
-			progressObserver = progress?.observe(\.fractionCompleted) { sender, _ in
+			progressObserver = progress?.observe(\.fractionCompleted) { [weak self] sender, _ in
+				guard let self = self else {
+					return
+				}
+
 				self.percentComplete = sender.fractionCompleted
 			}
 
-			cancelObserver = progress?.observe(\.isCancelled) { sender, _ in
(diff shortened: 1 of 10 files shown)


Gifski/TimeRemainingEstimator.swift
```diff
@@ -10,13 +10,21 @@ final class TimeRemainingEstimator {
 	/// Begin fade out when remaining time reaches this amount.
 	let fadeOutThreshold: TimeInterval = 1
 
-	var progress: Progress? {
+	weak var progress: Progress? {
```
sindresorhus: The progress instance was retained in so many places:
- By the conversion view controller
- Here
- In DockProgress (https://github.com/sindresorhus/DockProgress/commit/3365c2be401b47bf4ab4fe7aa0ffbf7d1a9658fc)

Gifski/TimeRemainingEstimator.swift
```diff
@@ -10,13 +10,21 @@ final class TimeRemainingEstimator {
...
 		didSet {
-			progressObserver = progress?.observe(\.fractionCompleted) { sender, _ in
+			progressObserver = progress?.observe(\.fractionCompleted) { [weak self] sender, _ in
```
sindresorhus: Xcode complains that the observers leak, but I don't know any way to `nil` them when `progress` becomes nil from being weak, not from being set to `nil`. There's no way to observe when a weak variable becomes `nil`...

sunshinejr: why do you only want to know when it becomes `nil` due to weak referencing, not to `nil`ing it manually?

sindresorhus: I want both, but in this case, only knowing when it becomes nil through weak would be useful as we never manually `nil` it. However, it doesn't really matter in this case as `TimeRemainingEstimator` is released when the conversion is done anyway.

PR #182, opened 2020-02-10
Ensure there are enough frames

sindresorhus:
Previously, if there were not enough frames, the progress bar would show and just stay there without anything happening.

Not enough frames could happen on very short videos or if the user trims too much.

Screenshot 2020-02-10 at 20 16 19

### DIFF
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -8,6 +8,7 @@ final class Gifski {
 	enum Error: LocalizedError {
 		case invalidSettings
 		case unreadableFile
+		case notEnoughFrames(Int)
 		case generateFrameFailed(Swift.Error)
 		case addFrameFailed(Swift.Error)
 		case writeFailed(Swift.Error)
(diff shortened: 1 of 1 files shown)


Gifski/Gifski.swift
```diff
@@ -19,6 +20,8 @@ final class Gifski {
 				return "Invalid settings."
 			case .unreadableFile:
 				return "The selected file is no longer readable."
+			case let .notEnoughFrames(frameCount):
+				return "An animated GIF requires at least 2 frames. Your video only has \(frameCount) frame\(frameCount == 1 ? "" : "s")."
```
sindresorhus: I know this is not true, but it's easier to enforce 2 frames than having users being confused why their GIF is not playing because it only has 1 frame.

yakov116: I think `Your video contains 0 frames` is a better way to word it IMHO

sindresorhus: Agreed

PR #185, opened 2020-03-27
Fix looping the playback when video is trimmed from the start

### LINKED ISSUE #184: Looping in the editor should loop only in the selected area
After the last update the video playback in the editor started looping when "Loop forever" is checked, _which is awesome_, I just think that it makes more sense to only loop within the selected area. Currently looping starts from the beginning of the video.

sindresorhus:
Fixes #184

### DIFF
--- a/Gifski/util.swift
+++ b/Gifski/util.swift
@@ -2964,8 +2964,19 @@ extension Error {
 
 
 extension AVPlayer {
+	/**
+	Seek to the start of the playable range of the video.
+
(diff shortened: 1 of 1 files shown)


sindresorhus:
@silvenon Can you try it out?

[Gifski.app.zip](https://github.com/sindresorhus/Gifski/files/4394046/Gifski.app.zip)

*(You need to right-click the app and select "Open" to open it)*

silvenon:
Works great! 👌

PR #188, opened 2020-04-06
Fix handling of videos with blank frames at the start

### LINKED ISSUE #186: "Couldn't open file" when started recording & progress isn't reported correctly.
This is my workflow:
1. Run simulator recording
```
xcrun simctl io booted recordVideo "$1"
```
1. Stop & open the file with Gifski
1. Tap convert

What happens is that most of the time I tap convert it says "Couldn't open file", I have to wait (or tap on buttons, not sure which one works) and it converts eventually. Also when it finally starts converting, the progress reported isn't entirely accurate since it finishes at about 50%.

There is no particular video that triggers that as it is consistently reproducible. I'm thinking if this has something to do with APFS?

sindresorhus:
Fixes #186 

Can be tested with the fixture in #186.

This issue has been eating at me for a year and I finally figured out why it happens. 🎉

I would really prefer this solution instead:

> TODO: Ideally, we should trim blank frames when initially reading the video in `VideoValidator.swift`, but I don't know a way to detect blank frames.

I'll try to find a solution for the above, if not, this PR is a good workaround for now.

Let me know if anyone has any idea of how to do the above.

### DIFF
--- a/Gifski/util.swift
+++ b/Gifski/util.swift
@@ -322,12 +322,13 @@ extension AVAssetImageGenerator {
 		let isFinished: Bool
 	}
 
+	/// - Note: If you use `result.completedCount`, don't forget to update its usage in each `completionHandler` call as it can change if frames are ignored.
 	func generateCGImagesAsynchronously(
 		forTimePoints timePoints: [CMTime],
 		completionHandler: @escaping (Swift.Result<CompletionHandlerResult, Error>) -> Void
 	) {
 		let times = timePoints.map { NSValue(time: $0) }
-		let totalCount = times.count
+		var totalCount = times.count
 		var completedCount = 0
 
 		generateCGImagesAsynchronously(forTimes: times) { requestedTime, image, actualTime, result, error in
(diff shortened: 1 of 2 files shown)


Gifski/util.swift
```diff
@@ -348,6 +349,17 @@ extension AVAssetImageGenerator {
 					)
 				)
 			case .failed:
+				// TODO: Ideally, we should trim blank frames when initially reading the video in `VideoValidator.swift`, but I don't know a way to detect blank frames.
```
sunshinejr, line 352: I think this workaround is actually a pretty good fix, maybe even better than trimming the video as this covers the case when a video has a blank frame in the middle of a video? not sure if possible though

sindresorhus: Yes, I would still keep this fix even if I find a way to trim. The reasons I want to trim is that then it wouldn't waste time trying to decode empty frames and also because the empty frames show up in the editor preview.

sindresorhus: > video has a blank frame in the middle of a video

It is possible if someone has composed multiple videos in a composition but has not done the timing correctly and left a gap between them.

PR #189, opened 2020-04-06
Fix a crash that can sometimes happen when cancelling

sindresorhus:
I'm yet again looking into the crash that sometimes happen when you cancel the conversion.

This time, I'm trying the trick mentioned in https://forums.developer.apple.com/message/255978#255978 to be able to create normal callbacks from a C-based callbacks at the GifskiWrapper level. This makes it easier to ensure a weak self and prevent accidentally retaining self.

However, when trying to convert, I'm seeing a "Use of stack memory after return" address sanitizer issue at https://github.com/sindresorhus/Gifski/pull/189/files#diff-1df33fd7967eaf75e9f4c10c855f03f7R94 and https://github.com/ImageOptim/gifski/blob/a94b2828b33014bdf1704a97c3040acb023c8a2e/src/c_api.rs#L316 @kornelski Are you sure there's nothing in libgifski that is incorrectly retaining the pointer or something?

### DIFF
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -71,6 +71,13 @@ final class Gifski {
 		progress = Progress(parent: .current())
 
 		let completionHandlerOnce = Once().wrap { [weak self] (_ result: Result<Data, Error>) -> Void in
+			if
+				case .failure(let error) = result,
+				case .cancelled = error
+			{
+				try? self?.gifski?.finish()
+			}
+
 			self?.gifski?.release()
 
 			DispatchQueue.main.async {
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -47,10 +47,8 @@ final class Gifski {
 		var loopGif: Bool = true
 	}
 
-	// TODO: `NSMutableData` is not thread-safe. We should use a lock when writing to it.
 	private var gifData = NSMutableData()
 	private var progress: Progress!
-	private let writeQueue = DispatchQueue(label: "com.sindresorhus.Gifski.writeQueue", qos: .utility)
 
 	// TODO: Split this method up into smaller methods. It's too large.
 	/**
@@ -95,10 +93,14 @@ final class Gifski {
 			return
 		}
 
-		gifski.setProgressCallback(context: &progress) { context in
-			let progress = context!.assumingMemoryBound(to: Progress.self).pointee
-			progress.completedUnitCount += 1
-			return progress.isCancelled ? 0 : 1
+		gifski.setProgressCallback { [weak self] in
+			guard let self = self else {
+				return 1
+			}
+
+			self.progress.completedUnitCount += 1
+
+			return self.progress.isCancelled ? 0 : 1
 		}
 
 		gifski.setWriteCallback { [weak self] bufferLength, bufferPointer in
@@ -106,13 +108,7 @@ final class Gifski {
 				return 0
 			}
 
-			self.writeQueue.async { [weak self] in
-				guard let self = self else {
-					return
-				}
-
-				self.gifData.append(bufferPointer, length: bufferLength)
-			}
+			self.gifData.append(bufferPointer, length: bufferLength)
 
 			return 0
 		}
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -5,6 +5,9 @@ import Crashlytics
 var conversionCount = 0
 
 final class Gifski {
+	deinit {
+		print("GIFSKI DEINIT")
+	}
 	enum Error: LocalizedError {
 		case invalidSettings
 		case unreadableFile
@@ -49,6 +52,7 @@ final class Gifski {
 
 	private var gifData = NSMutableData()
 	private var progress: Progress!
+	private var gifski: GifskiWrapper?
 
 	// TODO: Split this method up into smaller methods. It's too large.
 	/**
@@ -67,6 +71,8 @@ final class Gifski {
 		progress = Progress(parent: .current())
 
 		let completionHandlerOnce = Once().wrap { [weak self] (_ result: Result<Data, Error>) -> Void in
+			self?.gifski?.release()
+
 			DispatchQueue.main.async {
 				guard
 					let self = self,
@@ -88,7 +94,9 @@ final class Gifski {
 			fast: false
 		)
 
-		guard let gifski = GifskiWrapper(settings: settings) else {
+		self.gifski = GifskiWrapper(settings: settings)
+
+		guard let gifski = gifski else {
 			completionHandlerOnce(.failure(.invalidSettings))
 			return
 		}
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -5,9 +5,6 @@ import Crashlytics
 var conversionCount = 0
 
 final class Gifski {
-	deinit {
-		print("GIFSKI DEINIT")
-	}
 	enum Error: LocalizedError {
 		case invalidSettings
 		case unreadableFile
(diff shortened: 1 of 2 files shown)


kornelski:
That line points to use of the data you've provided…

That looks suspicious:

```swift
var callback = callback
```

aren't vars on stack? If so then `&callback` is a on-stack temporary and can't be given out.

sindresorhus:
> aren't vars on stack? If so then &callback is a on-stack temporary and can't be given out.

I honestly don't know. I was following the example in https://forums.developer.apple.com/message/255978#255978 written by an Apple employee. I assumed he knows what he's doing...

sindresorhus:
> In Swift I guess that's done by making an object and putting something as a property in that object? So assign callback to something like self.callback and then pass a pointer to self.callback to the library.

Tried this now in https://github.com/sindresorhus/Gifski/pull/189/commits/ae0a7859a488a048e9998235473912be26b94ce2 and I get the exact same sanitation error.

sindresorhus:
> That line points to use of the data you've provided…

I'm aware. I was just mentioning it as libgifski is also part of the stack trace. Just trying to rule out libgifski as the problem.

Screenshot 2020-04-06 at 17 18 20

kornelski:
That discussion was about blocks that are executed on the same stack frame, i.e. convincing swift that `@escaping` is not needed.

In this case the writing will be run on another thread, so it is escaping the current stack.

kornelski:
Could you be setting write callback twice? Overwrite of the property would invalidate the previous value.

sindresorhus:
> Could you be setting write callback twice? Overwrite of the property would invalidate the previous value.

No, I've confirmed it's only set once.

sindresorhus:
I think I have found a solution in https://github.com/sindresorhus/Gifski/pull/189/commits/feecd24ec93b6e7de2b41608d0f9fe6133faaff0. I have not been able to reproduce the cancel crash since then.

Gifski/GifskiWrapper.swift
```diff
@@ -68,18 +68,45 @@ final class GifskiWrapper {
...
+	func setProgressCallback(_ callback: @escaping ProgressCallback) {
+		progressCallback = callback
+
+		gifski_set_progress_callback(
+			pointer,
+			{ context in // swiftlint:disable:this opening_brace
+				let this = Unmanaged<GifskiWrapper>.fromOpaque(context!).takeUnretainedValue()
+				return Int32(this.progressCallback())
+			},
+			Unmanaged.passRetained(self).toOpaque()
```
sindresorhus, line 84: I have no idea how manual memory management works. It only works when passed as retained here. If I use `. passUnretained()`, it crashes as usual.

@kornelski Do you think this handling is correct? I guess `gifski_set_progress_callback` does not retain the `context` for me?

I have a feeling the problem before was that the passed in `context` got deallocated when Gifski cancelled the conversion, but libgifski still tried to call the callback afterwards, resulting in a crash. Any thoughts on that theory?

kornelski: Yeah, C API doesn't even know retaining exists. Never retains anything.

kornelski: Does passRetained increase refcount? If so, it may be leaking memory, because there won't be anything to decrease the refcount later. But leaks are better than crashes :)

sindresorhus: Could the C API check whether `context` is currently allocated and only call the callback if it is?

I’ll look into how to decrease the ref count when it’s not longer needed.

kornelski: Unfortunately such thing is not possible. It would be a life-saver for all user-after-free C bugs :) 

The closest approximation would be to take the callback as a weak reference, but C doesn't have them either.

Swift's memory management is unclear to me: it's automatic magic, except where it suddenly isn't :(

Maybe it would help to write an Objective-C wrapper for the C API?

Automobileservices4:
#189

sindresorhus:
@kornelski I found a way to manually release self: https://github.com/sindresorhus/Gifski/pull/189/commits/c7af700d6af3839fe28c2131fbf21a97682335da `GifskiWrapper` is now correctly deinit'd. However, I still sometimes get a crash for deallocated memory. So it seems that libgifski continues writing even after we cancel everything. Is there any way you could add a `.cancel()` method or something to libgifski? If not, I think the Objective-C wrapper you mentioned is the only way to properly fix this. There's no way for the Swift code to know when libgiski will finish writing and no way to prevent the callback to be called when we don't want it to.

kornelski:
after [gifski_finish](https://docs.rs/gifski/0.10.0/gifski/c_api/fn.gifski_finish.html) returns it's guaranteed there will be no more writes.

sindresorhus:
I didn't realize we have to call `gifski_finish()` when cancelling. I don't think the docs says anything about that, by the way.

sindresorhus:
With https://github.com/sindresorhus/Gifski/pull/189/commits/77befeda95ffd098ea9f20404102c02db99ad263, I'm not longer able to crash it on cancel. 🎉

sindresorhus:
Hmm, thinking about this a bit more. We should probably call `.finish()` on all errors except `. writeFailed` since that error is caused by calling `.finish()` in the first place. @kornelski Thoughts?

sindresorhus:
@kornelski Unrelated, but is https://github.com/ImageOptim/gifski/commit/062e73c49648bd9e485fc434baefacc679702574 safe to update here? I mean, is it done?

kornelski:
latest master commit in gifski is OK. But the last frame pts change doesn't affect the app, so no need to update.

kornelski:
You always have to call `gifski_finish()`, otherwise you'll leak memory.

sindresorhus:
> You always have to call gifski_finish(), otherwise you'll leak memory.

That could be made clearer in the docs.

sindresorhus:
Alright. I now ensure that `gifski_finish` is always called and only once.

sindresorhus:
The PR should be ready now: https://github.com/sindresorhus/Gifski/pull/189/files

Does it look ok?

Gifski/GifskiWrapper.swift
```diff
@@ -107,4 +139,12 @@ final class GifskiWrapper {
 	func finish() throws {
 		try wrap { gifski_finish(pointer) }
```
kornelski, line 157: I suggest setting the pointer to null ASAP, so that multiple calls can't crash it:

``
var tmp = pointer;
pointer = null;
gifski_finish(tmp);
```

File: Gifski/Gifski.swift [hunk 3]
```diff
@@ -68,6 +68,10 @@ final class Gifski {
 		progress = Progress(parent: .current())
 
 		let completionHandlerOnce = Once().wrap { [weak self] (_ result: Result<Data, Error>) -> Void in
+			// Ensure libgifski finishes no matter what.
+			try? self?.gifski?.finish()
+			self?.gifski?.release()
```
sindresorhus, line 73: I tried moving the `.release()` call into `.finish()`, but for some reason it doesn't release GifskiWrapper then. Looking at it, it should behave exactly the same, but it doesn't...

Gifski/GifskiWrapper.swift
sindresorhus: I already prevent multiple calls to `finish()`: https://github.com/sindresorhus/Gifski/pull/189/files#diff-1df33fd7967eaf75e9f4c10c855f03f7R153-R155

kornelski: But that's not thread-safe. `gifski_finish()` can take a while, so two calls to `finish()` one after another will call  `gifski_finish()` twice, then set the flag twice.

sindresorhus: Good point. Done.

sindresorhus:
Thanks for reviewing 🙌

sindresorhus:
It looks like this fixed the issue. Crash-free so far:

Screenshot 2020-05-07 at 01 56 48

PR #195, opened 2020-06-11
Warn users when copying that some apps cannot handle pasting animated GIFs

sindresorhus:
Someone complained in a review that Gifski produces a broken GIF when copying, not understanding that it's [Chrome (and Chrome-based apps) that suck](https://twitter.com/sindresorhus/status/1271001198422683648). This makes it clear.

Why two alerts? Because from experience a lot of users would just click away the first one without reading.

The alerts are only shown once.

I have the same alerts in my [Jiffy app](https://sindresorhus.com/jiffy) too.

**Feedback wanted on the alert text.**

### DIFF
--- a/Gifski/ConversionCompletedViewController.swift
+++ b/Gifski/ConversionCompletedViewController.swift
@@ -120,6 +120,21 @@ final class ConversionCompletedViewController: NSViewController {
 			self.copyButton.title = "Copied!"
 			self.copyButton.isEnabled = false
 
+			App.runOnce(identifier: "copyWarning") {
+				NSAlert.showModal(
+					for: self.copyButton.window,
+					message: "The GIF was copied to the clipboard.",
+					informativeText: "However…",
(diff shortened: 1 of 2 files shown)


kornelski:
Where does copying work? Maybe the option should be removed, or kept out of view only in Edit -> Copy menu.

sindresorhus:
> Where does copying work?

Most apps: Safari, Messages, Notes, Mail, Bear, Agenda, etc.

> Maybe the option should be removed, or kept out of view only in Edit -> Copy menu.

I don't want to limit functionality just because Chrome sucks.

kornelski:
In that case I'd phrase it as a "Chrome bug warning!" and say it affects Chrome and Chromium-based apps like Slack.

sindresorhus:
Well, it's not *just* Chrome, but Chrome is where most users would encounter the problem.

Issue #130, opened 2019-09-06
Display estimate as a range


kornelski:
The estimate could be displayed as a range, e.g.: 0.5MB-2MB. This could be done simply by calculating estimate twice, with different assumptions (fudge factors).

Displaying it as a range would communicate to users that it's not an actual expected file size, and would communicate how imprecise the estimate is, so they would be less surprised when it's off.

sindresorhus:
> with different assumptions (fudge factors).

What would the different assumptions be?

kornelski:
Magic constants for compression ratio and influence of quality.

jsejcksn:
Any kind of improved accuracy for expectations is good, I think.

<img width="201" alt="estimated" src="https://user-images.githubusercontent.com/868251/74587104-1566fd00-4fb4-11ea-9ef2-618ece3680fd.png">
<img width="126" alt="actual" src="https://user-images.githubusercontent.com/868251/74587105-17c95700-4fb4-11ea-83d4-f9e3acb2828e.png">

sindresorhus:
From https://github.com/sindresorhus/Gifski/pull/205#issuecomment-714612557:

> For lower/upper guesstimate I had in mind plugging in different constants/assumptions into the algorithm.

What would those constants/assumptions be? This issue is a bit vague on how exactly it should be solved.

kornelski:
```swift
		let frameCount = duration * frameRateSlider.doubleValue
		let dimensions = resizableDimensions.changed(dimensionsType: .pixels).currentDimensions.value
		var fileSize = (Double(dimensions.width) * Double(dimensions.height) * frameCount) / 3
		fileSize = fileSize * (qualitySlider.doubleValue + 1.5) / 2.5
```

The `/ 3` part is a total guess of compression ratio. It could be /5 and /2 for example.

sindresorhus:
The estimation has been completely rewritten since this was filed. It now does an actual partial GIF generation (sampling every 5th frame) and scales the result, rather than relying on the naive formula with guessed compression ratios. The naive estimate is only used as a brief fallback while the real estimate computes. A range display would add UI complexity for little benefit given the current accuracy. Closing as outdated.

PR #205, opened 2020-10-15
Improve file size estimate

### LINKED ISSUE #41: Improve file size estimate

sindresorhus:
I tried out many different algorithms and variants, and in the end, found that this one produced the most accurate estimate:
- Get all the normal frame time codes.
- Pick 5 consecutive samples from 5 evenly distributed places in the video.
- Convert with current settings.
- Divide by the original frame count compared to the one used in the estimate.

Fixes #41 

---

Here is a build: [Gifski - with estimate.zip](https://github.com/sindresorhus/Gifski/files/5388439/Gifski.-.with.estimate.zip)

**Please try it out on various videos and check the estimate compared to the final result. The most important part is that the estimate never shows less than the actual file.**

Note: The code and look are not done. It should be cleaned up a lot, but I would like to finalize the algorithm first.

Some potential optimization would be to run the 5 samples concurrently, but I don't plan to do that in this PR. I'll add a TODO comment.

---

Open questions:

- [x] Should we still show the naive estimate (the one we currently use) while the good estimate is being generated? Currently, the estimate takes from 5 - 15 seconds. The downside of showing the naive one is that it can be woefully incorrect sometimes.

### DIFF
--- a/Gifski/ConversionViewController.swift
+++ b/Gifski/ConversionViewController.swift
@@ -80,7 +80,7 @@ final class ConversionViewController: NSViewController {
 				return
 			}
 
-			gifski.run(conversion) { result in
+			gifski.run(conversion, isEstimation: false) { result in
 				do {
 					let gifUrl = try self.generateTemporaryGifUrl(for: conversion.video)
 					try result.get().write(to: gifUrl, options: .atomic)
--- a/Gifski/EditVideoViewController.swift
+++ b/Gifski/EditVideoViewController.swift
@@ -49,6 +49,17 @@ final class EditVideoViewController: NSViewController {
 		maxWidth: 300
 	)
 
+	private var conversionSettings: Gifski.Conversion {
+		.init(
+			video: inputUrl,
+			timeRange: timeRange,
+			quality: Defaults[.outputQuality],
+			dimensions: resizableDimensions.changed(dimensionsType: .pixels).currentDimensions.value,
+			frameRate: frameRateSlider.integerValue,
+			loopGif: Defaults[.loopGif]
+		)
+	}
+
 	convenience init(
 		inputUrl: URL,
 		asset: AVAsset,
@@ -65,16 +76,7 @@ final class EditVideoViewController: NSViewController {
 
 	@IBAction
 	private func convert(_ sender: Any) {
-		let conversion = Gifski.Conversion(
-			video: inputUrl,
-			timeRange: timeRange,
-			quality: Defaults[.outputQuality],
-			dimensions: resizableDimensions.changed(dimensionsType: .pixels).currentDimensions.value,
-			frameRate: frameRateSlider.integerValue,
-			loopGif: Defaults[.loopGif]
-		)
-
-		let convert = ConversionViewController(conversion: conversion)
+		let convert = ConversionViewController(conversion: conversionSettings)
 		push(viewController: convert)
 	}
 
@@ -340,7 +342,17 @@ final class EditVideoViewController: NSViewController {
 		selectPredefinedSizeBasedOnCurrentDimensions()
 	}
 
+	private func setEstimatedFileSize(_ string: String) {
+		estimatedSizeLabel.stringValue = "Estimated File Size: \(string)"
+	}
+
+	private var gifski: Gifski?
+
 	private func estimateFileSize() {
+		Debouncer.debounce(delay: 0.5, action: _estimateFileSize)
+	}
+
+	private func _estimateFileSize() {
 		let duration: Double = {
 			guard let timeRange = timeRange else {
 				return videoMetadata.duration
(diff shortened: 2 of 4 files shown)


kornelski:
Looks good.

I see you're adding 10% just in case. I have an idea to make it more scientific: measure frame sizes in bytes, and compute standard deviation of frame sizes. If the deviation is large, then the estimate is uncertain and should be inflated. If all frame sizes are about the same, then the estimate is likely to be accurate.

kornelski:
Alternative solution, which I think we've discussed previously, is to start the actual final conversion in the background, and use it for the estimate. When user presses start, instead of restarting, just reuse the in-progress conversion. This will have disadvantage of using frames from the beginning for the estimate, but OTOH it will make conversion seem faster, since it will get a head start.

sindresorhus:
> I have an idea to make it more scientific: measure frame sizes in bytes, and compute standard deviation of frame sizes. If the deviation is large, then the estimate is uncertain and should be inflated. If all frame sizes are about the same, then the estimate is likely to be accurate.

That's a good idea. I'll try it out.

> This will have disadvantage of using frames from the beginning for the estimate

I tried that too and the estimate was much worse. We really evenly spread out samples to get an accurate estimate. I rather have a more accurate estimate over a slightly shorter conversion time.

sunshinejr:
@sindresorhus this is looking really good! tried few recordings from apps (this is mostly my use case for Gifski), and the file size was always a bit bigger, but not that much. E.g. I saw naive 59mb, then updated to 35,1 and then it generated 34,9. This is a really big improvement for me. And personally, I'd skip the "naive" and maybe add a loader or a "Estimating filesize" text?

kornelski:
Regarding naive estimate:

* Show a min-max range. "20MB-50MB" makes it clearer how imprecise it is.

* Once you get any better estimate, remember the ratio between the naive and the better estimates. Apply that ratio to later naive estimates. This will give you decent estimates real time for changing parameters.

sindresorhus:
> Show a min-max range. "20MB-50MB" makes it clearer how imprecise it is.

How should we get the range? Just expand the current naive estimate both ways to get a lower/upper bound, or do you have something more clever in mind?

> Once you get any better estimate, remember the ratio between the naive and the better estimate. Apply that ratio to later naive estimates. This will give you decent estimates real time for changing parameters.

That's a good idea.

kornelski:
For lower/upper guesstimate I had in mind plugging in different constants/assumptions into the algorithm.

sindresorhus:
> For lower/upper guesstimate I had in mind plugging in different constants/assumptions into the algorithm.

Let's continue this in https://github.com/sindresorhus/Gifski/issues/130.

sindresorhus:
I don't have time to implement and test all these things right now, but I've opened an issue to track them: https://github.com/sindresorhus/Gifski/issues/211

sindresorhus:
I think it's more important to get this out there now. A lot of people have complained about inaccurate estimates.

PR #207, opened 2020-10-16
Limit frame rates to 50 FPS

### LINKED ISSUE #200: Show warning when framerate is set above 50 fps
Due to the fact that nearly all browsers—still to this day—only support a maximum frame delay of 20 ms, GIFs exported at 60 fps (16 ms frame delay) will render at 10 fps (100 ms frame delay) in a browser. Finder happily plays back short frame delay GIFs, so I was mislead into thinking this would work on the internet, or in my case Slack.

It seems there's a use case for exporting GIFs with a framerate above 50 fps (see #161), so I think we should display a warning to the user if they exceed this 50 fps limit, and maybe even convert this into a snapping UISlider.
CleanShot 2020-08-07 at 13 14 14@2x

Gifski Warning Mockup

What do you all think of the copy in the warning message and the general approach?

sindresorhus:
I went with the simple solution of just capping it. I don't see the use of 60 FPS if it doesn't work well in browsers. 50 FPS should be more than enough for GIFs anyway.

It also shows a warning now the first time the user tries to open a video with >50 FPS:

Screen Shot 2020-10-16 at 15 32 31

Fixes #200

---

// @phaseOne

### DIFF
--- a/Gifski/Constants.swift
+++ b/Gifski/Constants.swift
@@ -4,7 +4,7 @@ import Defaults
 struct Constants {
 	static let defaultWindowSize = CGSize(width: 360, height: 240)
 	static let backgroundImage = NSImage(named: "BackgroundImage")!
-	static let allowedFrameRate = 5.0...60.0
+	static let allowedFrameRate = 5.0...50.0
 }
 
 extension NSColor {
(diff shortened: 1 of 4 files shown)


kornelski:
Alert when opening a video may be too much, since user may not even intend to use 60fps. 

I'd default the slider to 20fps (personally I think any higher framerate is too much for GIF), and display this warning only if user drags the slider to 60fps. After it's been displayed once, cap the slider to 50fps.

sindresorhus:
> Currently, we default to half the frame rate if it's more than 24, but I'm happy to just default it to 20. I would never use more than 20 either.

https://github.com/sindresorhus/Gifski/pull/210

PR #217, opened 2020-11-17
Make older Gifski release clearer in the readme

janbrasna:
Only added a contextual link down to the archive version from system requirements up there just for convenience… +fixed link text a11y

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -17,7 +17,7 @@ Gifski supports all the video formats that macOS supports (`.mp4` or `.mov` with
 
 **[Blog post](https://blog.sindresorhus.com/gifski-972692460aa5)** &nbsp;&nbsp; **[Product Hunt](https://www.producthunt.com/posts/gifski-2)**
 
-Requires macOS 10.14 or later.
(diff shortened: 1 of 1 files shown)


sindresorhus:
I think we can just remove the FAQ item and move the link directly to the download section. We could move the 10.14 note there too.

janbrasna:
I figured there would be more similar cases now with new Universal Binaries et al., so possibly there's a new macOS support breakdown coming soon anyways… (Not sure if the new arch build targets coming up mean more stringent macOS support, or the new Xcode output is still ok for 10.15 & 11 at the same time…)

sindresorhus:
Nothing will change regarding supported macOS versions in the near future. Universal binaries are backwards compatible.

janbrasna:
Gotcha. Reshuffled. Not too sure about the wording tho, tried to sound as helpful as possible (wanting to add something along the lines of a release date on 12/2019 to the v2.4.0 for a time frame reference what to expect regarding how old is it from today… but couldn't make it sound casual) — so feel free to fix up if you can get it better;)

PR #218, opened 2020-11-25
Add a loop count setting

0bmay:
This changes Gifski's gif animation looping modes from none - forever ti add in an in between for looping the animation 1 - 9999 files along with the existing No loops and Forever loops.

I sent in a feature request this morning, and when I looked into what it would take, I challenged myself to add it into the awesome tool.

This code base is based off the latest master for  sindresorhus/Gifski and the latest revision to kornelski/rust-ffmpeg


Gifski/EditVideoViewController.swift
```diff
@@ -41,6 +42,9 @@ final class EditVideoViewController: NSViewController {
 	private let formatter = ByteCountFormatter()
 	private var playerViewController: TrimmingAVPlayerViewController!
 
+	// a loop of 1 is times shown of 2..  a loop of 2 is timesShown of 3
+	private var timesShown: Int = 1
```
sindresorhus, line 46: Use the wording `loopCount` instead of `timesShown` everywhere.

gifski-api/gifski.h
```diff
@@ -53,6 +53,10 @@ typedef struct GifskiSettings {
    * Lower quality, but faster encode.
    */
   bool fast;
+  /**
+   * 0+, but useful range is 1-10. Recommended to set to 5. If times_shown is > 0 once should be set to true
+   */
+  uint16_t times_shown;
```
sindresorhus, line 59: You need to upstream these changes (any changes in the `gifski-api` directory) to https://github.com/ImageOptim/gifski, which is the source of truth.

Gifski.xcodeproj/project.pbxproj
```diff
@@ -509,7 +509,7 @@
 				CODE_SIGN_IDENTITY = "Apple Development";
 				CODE_SIGN_STYLE = Automatic;
 				COMBINE_HIDPI_IMAGES = YES;
-				DEVELOPMENT_TEAM = YG56YK5RN5;
+				DEVELOPMENT_TEAM = 8JPUXG2X2Z;
```
sindresorhus, line 512: This should not be committed.

Gifski/EditVideoViewController.swift
```diff
@@ -314,6 +320,28 @@ final class EditVideoViewController: NSViewController {
 		updateTextFieldsMinMax()
 	}
 
+	private func setUpTimesShownTextField() {
+		timesShownTextField.onBlur = { [weak self] timesShown in
```
sindresorhus, line 324: Use a `guard let self = self else` guard.

Gifski/Gifski.swift
```diff
@@ -97,7 +98,8 @@ final class Gifski {
 			height: UInt32(conversion.dimensions?.height ?? 0),
 			quality: UInt8(conversion.quality * 100),
 			once: !conversion.loopGif,
-			fast: false
+			fast: false,
+			times_shown: UInt16(conversion.timesShown)
```
sindresorhus, line 102: I think the `times_shown` parameter should be optional.

Gifski/Gifski.swift
```diff
@@ -45,6 +45,7 @@ final class Gifski {
 		var dimensions: CGSize?
 		var frameRate: Int?
 		var loopGif: Bool = true
+		var timesShown: Int = 9999
```
sindresorhus, line 48:
```suggestion
		var timesShown: Int?
```

sindresorhus (changes requested):

sindresorhus:
- The textfield is too wide.
- You need to add a stepper control on the right side of the textfield.
	https://stackoverflow.com/questions/702829/integrate-nsstepper-with-nstextfield
- The textfield should not allow negative values.
- The textfield should only allow numbers.
- The textfield should not be allowed to be empty.
- When "loop forever" is enabled, the textfield should be disabled.
- "loop forever" can be renamed to "forever".
- In the Rust implementation, make sure the loop count is also passed to gifsicle, which is used when the quality slider is not 100%.

sindresorhus:
> I sent in a feature request this morning, and when I looked into what it would take, I challenged myself to add it into the awesome tool.

I already decline this, but seeing as you're willing to put an effort into this, I'm ok with adding ut 👍🏻

sindresorhus:
These labels should be vertically trailing aligned:

Screen Shot 2020-11-25 at 10 10 09

sindresorhus:
- Having any number in the loop count text field should not disable the "forever" checkbox.

0bmay:
Sindre, thank you for your feedback. I am working on the changes.

Gifski/Gifski.swift
0bmay: Done

gifski-api/gifski.h
0bmay: I will add in a pull request for that project as well.

sindresorhus:
> The textfield should not allow negative values.

This is not fixed.

sindresorhus:
Is it really useful to allow numbers up to 9999? I think we should maybe set the limit at 100.

sindresorhus:
There needs to be slight more margin between the text field and the stepper.

sindresorhus:
If you choose a loop count, I think the preview player should loop. Ideally, it should loop according to the loop count, but not sure how easy that would be to achieve.

sindresorhus:
> These labels should be vertically trailing aligned:

Actually, I was wrong. It should be aligned to the slider:

Screen Shot 2020-11-26 at 10 07 56

Let's also rename it back to "Loop Forever" as otherwise there's just too much empty space and it looks weird.

0bmay:
> Is it really useful to allow numbers up to 9999? I think we should maybe set the limit at 100.

Good point, anything more than 100 (or 30+) is not much different than looping forever anyway.

0bmay:
Again, thank you for the feedback.

0bmay:
The underlying gifski library works as expected for no looping, forever loops or X loops as tested via the command line tool.  I need to think more on how the looping is set via the UI as right now it's not working as expected.

Gifski/EditVideoViewController.swift
```diff
@@ -401,7 +402,7 @@ final class EditVideoViewController: NSViewController {
 		let heightMinMax = resizableDimensions.heightMinMax
 		widthTextField.minMax = Int(widthMinMax.lowerBound)...Int(widthMinMax.upperBound)
 		heightTextField.minMax = Int(heightMinMax.lowerBound)...Int(heightMinMax.upperBound)
-		loopCountTextField.minMax = 0...9999
+		loopCountTextField.minMax = 0...100
```
sindresorhus, line 405: Can you move the range to https://github.com/sindresorhus/Gifski/blob/5c2d437d162948da78565e663df1d2e8cf5253d6/Gifski/Constants.swift#L7 ? And reuse is in both places.

sindresorhus:
- The "loop forever" is not perfectly aligned to the right of the above slider.

sindresorhus:
I also noticed that a lot of UI elements got slightly moved compared to `master`. They are now unaligned. The Dimensions dropdown, the buttons, etc. Ideally, nothing should be moved.

Screen Recording 2020-11-27 at 17 24 11

Gifski/EditVideoViewController.swift
0bmay: Waiting to push...

0bmay:
> I also noticed that a lot of UI elements got slightly moved compared to `master`. They are now unaligned. The Dimensions dropdown, the buttons, etc. Ideally, nothing should be moved.
> 
> Screen Recording 2020-11-27 at 17 24 11

Thanks "Add missing constraints"... will revert.

Gifski/Constants.swift
```diff
@@ -5,6 +5,7 @@ struct Constants {
 	static let defaultWindowSize = CGSize(width: 360, height: 240)
 	static let backgroundImage = NSImage(named: "BackgroundImage")!
 	static let allowedFrameRate = 5.0...50.0
+	static let repeatCountRange = 0...100
```
sindresorhus, line 8:
```suggestion
	static let loopCountRange = 0...100
```

sindresorhus:
Super nitpick, but I noticed the stepper is 1px off vertically from the text field. If you have the correct constraints, this might just be how the bounds of the stepper/textfields are set up, and then it's ok, but might be something to look into.

Screen Shot 2020-11-28 at 08 35 24

sindresorhus:
This looks pretty good to me otherwise. Now it's just waiting for https://github.com/ImageOptim/gifski/pull/151

sindresorhus:
I think the stepper should be 1 point closer to the text field (from 9 to 8, since 8 is the magic distance number).

Gifski/Constants.swift
0bmay: the name here is a preview of the changes to come to the upstream gifski library that I'm working on.  I'll make the change to the code here now, as it's more reflective of the current app's code.

0bmay:
> Super nitpick, but I noticed the stepper is 1px off vertically from the text field. If you have the correct constraints, this might just be how the bounds of the stepper/textfields are set up (or maybe something to do with the baseline), and then it's ok, but might be something to look into.
> 
> Screen Shot 2020-11-28 at 08 35 24

I believe I had it aligned center + bottom and now have it center + top.  My stepper for Xcode 12.2 looks different than yours, but I wouldn't think that would account for any pixel offsets, but stranger things have happened.  I have updated the code.

sindresorhus:
> I believe I had it aligned center + bottom and now have it center + top.

I believe the first baseline of the stepper should be constrained to the first baseline of the text field.

I also noticed that the stepper is currently constrained to the slider, which is incorrect.

sindresorhus:
Everything here looks good. Now just waiting for https://github.com/ImageOptim/gifski/pull/151.

0bmay:
updated code to use the upstream changes to [ImageOptim/gifski#151](https://github.com/ImageOptim/gifski/pull/151).

sindresorhus:
Just noticed that https://github.com/sindresorhus/Gifski/pull/218#issuecomment-734045735 is not done.

sindresorhus:
Found a bug. When opening a video and the editor shows, the "loop count" text field is not disabled even though the "loop forever" checkbox is checked.

Screen Shot 2020-12-18 at 13 52 28

---

(In hindsight, this feature would have been much more easily implemented in SwiftUI as we wouldn't have weird state bugs like this)

0bmay:
> Just noticed that [#218 (comment)](https://github.com/sindresorhus/Gifski/pull/218#issuecomment-734045735) is not done.

I think I have found a bug in NSImageView. If loading an animated gif into an NSImageView the animation will be shown at total times equal to the loop count (x) setting and not looped x times.

e.g. Loop Count is 1, the animation should be shown twice, or repeat the animation 1 time.
NSImageView sees the Loop Count as the number of times to show the animation and a value of 1 means no looping.

Also, a Loop Count of 0 in the gif is show the animation once and stop.
NSImageView sees a Loop Count of 0 as Loop Forever.

What I can do to fix the non looping issue is to not animate the image if we are not looping, but I think that would defeat the purpose of the preview window.

So how the animation is shown in an NSImaveView is not how any other tool would show the animated gif unless the animated gif's setting is loop forever.

sindresorhus:
> e.g. Loop Count is 1, the animation should be shown twice, or repeat the animation 1 time.
NSImageView sees the Loop Count as the number of times to show the animation and a value of 1 means no looping.

Can you report this to Apple?

> So how the animation is shown in an NSImaveView is not how any other tool would show the animated gif unless the animated gif's setting is loop forever.

Yeah, I think this applies to Quick Look too. At least I know it ignores the loop count completely.

sindresorhus:
Maybe we should have a warning (NSAlert) the first time the user sets the loop count in the UI to just inform them that it usually works, but the preview and Quick Look will not show the correct result>

0bmay:
Issue reported to Apple Feedback ID [FB8947153](https://feedbackassistant.apple.com/feedback/8947153)

I can add a one time alert when the looping/repeat value is not forever.

I know there is code out there that splits the animated gif out into individual frames and controls the loop that way, but I think that might be overkill for a screen that is shown for a short period of time and is a smaller view of the resulting animation.

sindresorhus:
> I can add a one time alert when the looping/repeat value is not forever.

👍  You can use: https://github.com/sindresorhus/Gifski/blob/5c2d437d162948da78565e663df1d2e8cf5253d6/Gifski/EditVideoViewController.swift#L274-L282

> I know there is code out there that splits the animated gif out into individual frames and controls the loop that way, but I think that might be overkill for a screen that is shown for a short period of time and is a smaller view of the resulting animation.

Agreed, that's too much.

Gifski/EditVideoViewController.swift
```diff
@@ -299,6 +299,19 @@ final class EditVideoViewController: NSViewController {
...
 	}
 
+	private func showConversionCompletedAnimationWarningIfNeeded() {
+		// TODO: This function can be removed once NSImageView respects the looping counter in the GIF header.
+		SSApp.runOnce(identifier: "convCompleteWarning") {
+			DispatchQueue.main.async { [self] in
+				NSAlert.showModal(
+					for: view.window,
+					message: "Animated GIF Preview Limitation",
+					informativeText: "The Conversion Completed Preview after processing will not reflect the image repetitions selected. However, the exported file will refelct the options selected and repeat correctly."
```
sindresorhus, line 309: I think this could be more succinctly written.

And as I commented earlier, Quick Look ignores the loop count and just always loops forever, which should also be mentioned.

"image repetitions" => "loop count"

Also typo.

Gifski/EditVideoViewController.swift
```diff
@@ -299,6 +299,19 @@ final class EditVideoViewController: NSViewController {
 		}
 	}
 
+	private func showConversionCompletedAnimationWarningIfNeeded() {
+		// TODO: This function can be removed once NSImageView respects the looping counter in the GIF header.
```
sindresorhus, line 303: Can you mirror the FB you submitted to Apple over at https://github.com/feedback-assistant/reports (follow the template) and then link to it here?

Gifski/EditVideoViewController.swift
```diff
@@ -299,6 +299,19 @@ final class EditVideoViewController: NSViewController {
...
+		SSApp.runOnce(identifier: "convCompleteWarning") {
```
sindresorhus, line 304:
```suggestion
		SSApp.runOnce(identifier: "gifLoopCountWarning") {
```

Gifski/EditVideoViewController.swift
0bmay: Done and thank you! I didn't know this resource existed.

Gifski/EditVideoViewController.swift
```diff
@@ -282,6 +299,20 @@ final class EditVideoViewController: NSViewController {
...
 
+	private func showConversionCompletedAnimationWarningIfNeeded() {
+		// TODO: This function can be removed once NSImageView respects the looping counter in the GIF header.
+		// https://github.com/feedback-assistant/reports/issues/187
+		SSApp.runOnce(identifier: "gifLoopCountWarning") {
+			DispatchQueue.main.async { [self] in
+				NSAlert.showModal(
+					for: view.window,
+					message: "Animated GIF Preview Limitation",
+					informativeText: "The after conversion preview and Quick Look will may not loop as expect, but the exorted file is valid."
```
sindresorhus, line 310: There are 3 typos in this text. I think it also could be more explicit about in what case it may not loop as expected.

Gifski/EditVideoViewController.swift
sindresorhus: This comment is now outdated. The code should not be removed when it's fixed as it also applies to Quick Look. Rather the text should be modified to only mention Quick Look when it's fixed.

sindresorhus:
This looks good now. Thank you for your hard work :)

Issue #220, opened 2021-01-15
Mac app ridiculously slow


soshimee:
I just checked out Gifski, then I realized it wasn't working, so I tried the CLI version. CLI version worked fine, but Mac App Store one could barely process 64x64 sized videos.
Maybe related to Big Sur update?

kornelski:
Does it apply to all videos?

Can you use Activity Monitor to create a spindump report of the app while it's processing?

soshimee:
Dunno, can you give me some sample videos? I don't have that many videos so... The current video format that I'm using is mp4.

fregante:
Just as another point: Today I tried converting a 12MB 720p MOV screencast on Big Sur and it suggested 4 minutes build time and >120MB output. I don't know if that's standard.

kornelski:
Oh, that's totally expected. That's how the awful GIF format from 1989 works.

GIF was designed for "16p" animations. Throwing 720p content at it is pushing GIF to its limits, so expect it to be incredibly slow and unfanthomanbly huge.

kornelski:
@sindresorhus Maybe the app could warn about creating overkill GIFs? Warn when resolution is >= 720p, and/or when the total size of all frames is huge.

soshimee:
I'm fine with big outputs, but it takes HOURS to do that on the Mac app but
takes only several minutes on the CLI app.

sindresorhus:
@soshimee It did indeed get slower with Big Sur. Would you be able to share the 64x64 video with me? (You can send it privately too). Gifski is slow with 720P video, but a 64x64 video should be no problem.

PR #221, opened 2021-01-22
Fix “cannot decode” error & improve performance

### LINKED ISSUE #202: Failed to generate frame: Cannot Decode
macOS v.11.0 Beta (20A5364e)
Gifski v.2.9.0 (39)

**Screenshot:**
Bildschirmfoto 2020-09-10 um 16 26 45

**Video recording:**
[Bildschirmaufnahme 2020-09-10 um 16.18.34.zip](https://github.com/sindresorhus/Gifski/files/5202315/Bildschirmaufnahme.2020-09-10.um.16.18.34.zip)

sindresorhus:
Fixes #202
Related #220

### DIFF
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -179,15 +179,6 @@ final class Gifski {
 			let generator = AVAssetImageGenerator(asset: asset)
 			generator.appliesPreferredTrackTransform = true
 
-			// `.zero` tolerance is much slower and fails a lot on macOS 11. (macOS 11.1)
-			if #available(macOS 11, *) {
-				generator.requestedTimeToleranceBefore = CMTime(seconds: 0.1, preferredTimescale: .video)
-				generator.requestedTimeToleranceAfter = CMTime(seconds: 0.1, preferredTimescale: .video)
-			} else {
-				generator.requestedTimeToleranceBefore = .zero
-				generator.requestedTimeToleranceAfter = .zero
-			}
-
 			// This improves the performance a little bit.
 			if let dimensions = conversion.dimensions {
 				generator.maximumSize = CGSize(widthHeight: dimensions.longestSide)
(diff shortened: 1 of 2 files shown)


Gifski/Utilities.swift
```diff
@@ -374,11 +378,30 @@ extension AVAssetImageGenerator {
...
 					}
 
-					// macOS 11 started throwing “decode failed” error for some frames in screen recordings. As a workaround, we ignore these. We throw an error if more than 50% of the frames could not be decoded.
+					// macOS 11 (still an issue in macOS 11.1) started throwing “decode failed” error for some frames in screen recordings. As a workaround, we ignore these. We throw an error if more than 50% of the frames could not be decoded.
 					if error.code == .decodeFailed {
-						decodeFailureFrameCount += 1
-						totalCount -= 1
-						Crashlytics.recordNonFatalError(error: error, userInfo: ["requestedTime": requestedTime.seconds])
+						var newActualTime = CMTime.zero
+						if let image = try? self.copyCGImage(at: requestedTime, actualTime: &newActualTime) {
```
sindresorhus, line 384: This seems to funnily succeed even when `generateCGImagesAsynchronously` failed at the same requested time 🤷🏻‍♂️

Gifski/Gifski.swift
```diff
@@ -178,14 +178,29 @@ final class Gifski {
 
 			let generator = AVAssetImageGenerator(asset: asset)
 			generator.appliesPreferredTrackTransform = true
-			generator.requestedTimeToleranceBefore = .zero
-			generator.requestedTimeToleranceAfter = .zero
+
+			// `.zero` tolerance is much slower and fails a lot on macOS 11. (macOS 11.1)
+			if #available(macOS 11, *) {
+				generator.requestedTimeToleranceBefore = CMTime(seconds: 0.1, preferredTimescale: .video)
```
sindresorhus, line 184: @kornelski Do you think it's safe to do 0.1 threshold here? Maybe it should be based on the requested FPS, but not exactly sure what heuristics to use.

kornelski: that seems risky above 10fps. I'm worried that decoder could take advantage of it to snap to keyframes, so you'd have stutter around them.

I'd use `0.5/fps`, so that it's at worst half of delay in either direction.

sindresorhus: > I'd use 0.5/fps, so that it's at worst half of frame time in either direction.

👍

PR #222, opened 2021-01-22
Upgrade libgifski

sindresorhus:
https://github.com/ImageOptim/gifski/commit/41285587116a1a3db5401f1b070167476ceb8581

@kornelski Is it safe to upgrade to this commit?


Gifski/Gifski.swift
```diff
@@ -96,8 +96,9 @@ final class Gifski {
 			width: UInt32(conversion.dimensions?.width ?? 0),
 			height: UInt32(conversion.dimensions?.height ?? 0),
 			quality: UInt8(conversion.quality * 100),
+			once: false,
```
sindresorhus, line 99: @kornelski I assume `once` is ignored if `loop_count` is specified?

kornelski: Oops. No. This header got out of sync. Update to a newer commit.

PR #238, opened 2021-04-20
Split the `Gifski.run()` method into multiple smaller methods

ejensen:
When adding support for reversing gifs (PR #239), I noticed this TODO
> // TODO: Split this method up into smaller methods. It's too large.

I split up this 260 line `run` method into 4 methods


sindresorhus:
This looks great. Thanks for the cleanup! 🙌

If you notice anything else that could be improved, feel free to open an issue or PR about it ;)

PR #239, opened 2021-04-20
Add setting to create bounced (yo-yo) GIFs

### LINKED ISSUE #234: Support back-and-forth (yo-yo)
A few people have requested this over the years. I'm still not sure we should do it. There are endless amounts of editing features we could add but I would like to keep Gifski focused on converting and simple. Show your opinion on this with GitHub Reactions.

---

If we were to implement this, we would have to take the video track element, copy it, reverse the copy, and then put the copy after the original.

ejensen:
Adds support for creating bounced gifs. Bouncing a GIF appends the reversed sequence of frames to the end of the forward playing frames

The option is shown as a new checkbox next to the loop checkbox
editor

The editor previews the bouncing (when enabled)
editor playback

Issue #234 calls this "back-and-forth" or "yo-yo". I used "bounce" since that's what the iOS Photos app uses (screenshot below), but I'm fine with renaming the checkbox whatever you prefer
ios

Builds upon PR #238
Closes #234

### DIFF
--- a/Gifski/EditVideoViewController.swift
+++ b/Gifski/EditVideoViewController.swift
@@ -426,7 +426,13 @@ final class EditVideoViewController: NSViewController {
 		Defaults.observe(.loopGif) { [weak self] in
 			self?.playerViewController.loopPlayback = $0.newValue
 		}
-			.tieToLifetime(of: self)
+		.tieToLifetime(of: self)
+
+		Defaults.observe(.bounceGif) { [weak self] in
+			self?.playerViewController.bouncePlayback = $0.newValue
+			self?.estimateFileSize()
+		}
+		.tieToLifetime(of: self)
 
 		add(childController: playerViewController, to: playerViewWrapper)
 	}
--- a/Gifski/EditVideoViewController.swift
+++ b/Gifski/EditVideoViewController.swift
@@ -325,64 +325,28 @@ final class EditVideoViewController: NSViewController {
 		}
 	}
 
-	private func showKeyframeRateWarningIfNeeded() {
+	private func showKeyframeRateWarningIfNeeded(maximumKeyframeInterval: Double = 30) {
 		DispatchQueue.global(qos: .utility).async { [weak self] in
 			guard
-				Defaults[.bounceGif],
-				let asset = self?.asset,
-				let videoTrack = asset.firstVideoTrack,
-				let reader = try? AVAssetReader(asset: asset)
+				let keyframeInfo = self?.asset.firstVideoTrack?.getKeyframeInfo(),
+				keyframeInfo.keyframeInterval > maximumKeyframeInterval
 			else {
 				return
 			}
 
-			let trackReaderOutput = AVAssetReaderTrackOutput(track: videoTrack, outputSettings: nil)
-			reader.add(trackReaderOutput)
+			print("Low keyframe interval \(keyframeInfo.keyframeInterval)")
 
-			guard reader.startReading() else {
-				return
-			}
-
-			var frameCount: UInt = 0
-			var keyframeCount: UInt = 0
-
-			while true {
-				guard self != nil else {
-					reader.cancelReading()
+			DispatchQueue.main.async { [weak self] in
+				guard let self = self else {
 					return
 				}
 
-				guard let sampleBuffer = trackReaderOutput.copyNextSampleBuffer() else {
-					reader.cancelReading()
-					break
-				}
-
-				if CMSampleBufferGetNumSamples(sampleBuffer) > 0 {
-					frameCount += 1
-
-					if
-						let attachments = CMSampleBufferGetSampleAttachmentsArray(sampleBuffer, createIfNecessary: false) as? [NSDictionary],
-						attachments.first?[kCMSampleAttachmentKey_NotSync] == nil
-					{
-						keyframeCount += 1
-					}
-				}
-			}
-
-			let keyframeInterval = frameCount / keyframeCount
-			if keyframeInterval > 30 {
-				print("Low keyframe interval \(keyframeInterval)")
-
-				DispatchQueue.main.async { [weak self] in
-					if let self = self {
-						NSAlert.showModal(
-							for: self.view.window,
-							title: "Reverse Playback Preview Limitation",
-							message: "Reverse playback may stutter when the video has a low keyframe rate.\nThe GIF will not have the same stutter.",
-							defaultButtonIndex: -1
-						)
-					}
-				}
+				NSAlert.showModal(
+					for: self.view.window,
+					title: "Reverse Playback Preview Limitation",
+					message: "Reverse playback may stutter when the video has a low keyframe rate. The GIF will not have the same stutter.",
+					defaultButtonIndex: -1
+				)
 			}
 		}
 	}
@@ -498,10 +462,12 @@ final class EditVideoViewController: NSViewController {
 		.tieToLifetime(of: self)
 
 		playerRateObserver = playerViewController.playerView.player?.observe(\.rate, options: [.new]) { [weak self] _, change in
-			if change.newValue != 0 {
-				SSApp.runOnce(identifier: "lowKeyframeRateWarning") {
-					self?.showKeyframeRateWarningIfNeeded()
-				}
+			guard change.newValue == -1 else {
+				return
+			}
+
+			SSApp.runOnce(identifier: "lowKeyframeRateWarning") {
+				self?.showKeyframeRateWarningIfNeeded()
 			}
 		}
 
--- a/Gifski/EditVideoViewController.swift
+++ b/Gifski/EditVideoViewController.swift
@@ -43,6 +43,7 @@ final class EditVideoViewController: NSViewController {
 	private let formatter = ByteCountFormatter()
 	private var playerViewController: TrimmingAVPlayerViewController!
 	private var playerRateObserver: NSKeyValueObservation?
+	private var isKeyframeRateChecked = false
 
 	private var timeRange: ClosedRange<Double>? { playerViewController?.timeRange }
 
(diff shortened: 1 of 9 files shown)


sindresorhus:
I noticed that the player stutters a little bit when bouncing back sometimes:

https://user-images.githubusercontent.com/170270/115362419-b6ba9600-a1eb-11eb-8bd8-28ccba23c78d.mov

It also pauses a little bit too long. Not sure if there's anything we can do about that though.

Test file: [60fps - short.mp4.zip](https://github.com/sindresorhus/Gifski/files/6341871/60fps.-.short.mp4.zip)

Gifski/Gifski.swift
```diff
@@ -380,15 +389,29 @@ final class Gifski {
...
 					bytesPerRow: pixels.bytesPerRow,
 					pixels: pixels.bytes,
-					presentationTimestamp: max(0, result.actualTime.seconds - startTime)
+					presentationTimestamp: presentationTimestamp
 				)
+
+				if conversion.bounce {
+					try gifski?.addFrame(
+						pixelFormat: .argb,
+						frameNumber: result.totalCount * 2 - result.completedCount - 1,
```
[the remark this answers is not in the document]
sindresorhus: Maybe also a short code comment that explains the calculation, for future reference.

sindresorhus:
You can actually see this stutter in the final GIF too:

60fps - short

But interestingly only if you save it and open it in Quick Look. In Chrome, it plays fine. The pause at the end is still too long in Chrome too though.

sindresorhus (changes requested):

sindresorhus:
- Can you update the readme and app store description to mention support for bouncing? Should probably use the word (yo-yo) in parens in case the user is more familiar with that term.
- The "size estimation" is not updated when the "Bounce" checkbox is toggled.
- Can you add a `.toolTip` to the "Bounce" checkbox that explains to the user what it is? Not everyone is familiar with the bounce effect. It should also tell the user that bouncing will double the GIF size.

ejensen:
I pushed some changes to address a number of items pointed out. There are still a few things I'm working on:
- [x] Remove the delay at the apex (middle) of the bounce
- [x] Update the size estimation when the bounce checkbox is toggled
- [x] Add code comments on how the bouncing logic works
- [x] Add a tooltip to the bounce checkbox
- [ ] Fix the frame stutter in the video editor preview 
  - It is caused by `AVPlayer`'s handling of videos with low keyframe rates - jumping to the nearest keyframe
  - The attached video only has 1 keyframe in the 214 frame video, causing a more noticeable stutter than others
- [ ] Add documentation of the bounce option to readme
- [ ] Mention the bounce option in the App Store description

sindresorhus:
Can you not force push? 🙏 It makes it hard to see what changed. Just add new commits. I will squash on merge anyway.

ejensen:
> Can you not force push when doing changes? 🙏 (To resolve merge conflict its fine though) It makes it hard to see what changed. Just add new commits. I will squash on merge anyway.

Noted. I'll push new commits in further updates. You can see the diff in the force push by clicking the ["force-push"](https://github.com/sindresorhus/Gifski/compare/936823748148cd64772896cdeee1ba7fdc12528d..0d0e17dedb8b4da2fccdeee940ee1e841bda48cf) text in the update added by Github

ejensen:
After trying a number of things to attempt to remove the stutter in the reverse playback preview, I discovered Quicktime and all the other AVFoundation-based playback apps I tried, experience the same playback stutter when playing that video in reverse:
https://user-images.githubusercontent.com/915431/115647423-9caab180-a2d8-11eb-9439-75b77c77a999.mp4

Using ffmpeg to adjust the keyframe interval resolved the stuttering in reverse playback in both Gifski and Quicktime
`ffmpeg -i input.mp4 -vcodec libx264 -x264-params keyint=30 -acodec copy output.mp4`

Resulting video:
https://user-images.githubusercontent.com/915431/115647408-96b4d080-a2d8-11eb-8c49-ab8e662b0c5d.mp4

Smooth reverse preview playback in Gifski:
https://user-images.githubusercontent.com/915431/115648753-dda3c580-a2da-11eb-80de-7dad27a0c56d.mp4

My instinct tells me that trying to re-keyframe videos before preview playback will not be efficient. I think we have a few other options:
1. Check the keyframe rate of the video and if it is too low for smooth reverse preview playback, provide a warning in the preview UI informing the user that the reverse playback is choppy in the preview but will be smooth in the GIF
2. Check the keyframe rate of the video and if it is too low for smooth reverse preview playback, don't attempt to bounce the preview
3. Leave the preview playback as-is and update the readme & App Store description with instructions on how to avoid the preview video playback stutter

Let me know what you think is the best course

Gifski/EditVideoViewController.xib
```diff
@@ -246,9 +246,25 @@
                                         </connections>
                                     </button>
                                     <stepper horizontalHuggingPriority="750" verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="x03-mN-zhh">
-                                        <rect key="frame" x="159" y="14" width="19" height="28"/>
+                                        <rect key="frame" x="159" y="34" width="19" height="28"/>
                                         <stepperCell key="cell" continuous="YES" alignment="left" maxValue="100" id="d96-4O-hec"/>
                                     </stepper>
+                                    <button verticalHuggingPriority="750" verticalCompressionResistancePriority="751" translatesAutoresizingMaskIntoConstraints="NO" id="6Zb-wc-GCx">
+                                        <rect key="frame" x="197" y="17" width="72" height="18"/>
+                                        <string key="toolTip">When enabled the GIFs will playback normally then playback in reverse. Resulting in a GIF that is double the length and double the filesize.</string>
```
sindresorhus, line 254:
```suggestion
                                        <string key="toolTip">Makes the GIF play normally and then in reverse, resulting in a GIF that is double the length and double the file size.</string>
```

sindresorhus:
Thanks for investigating this. I think we should go with `1`:

> Check the keyframe rate of the video and if it is too low for smooth reverse preview playback, provide a warning in the preview UI informing the user that the reverse playback is choppy in the preview but will be smooth in the GIF

You can use `SSApp.runOnce` to make sure the message is only shown once.

---

Probably worth filing a Feedback Assistant report about this too.

ejensen:
I pushed the commits that add the warning about video preview playback on low keyframe rate videos
warning

Gifski/Gifski.swift
```diff
@@ -219,7 +225,7 @@ final class Gifski {
 	private func imageGenerator(
 		for conversion: Conversion,
 		jobKey: String
-	) -> Result<(generator: AVAssetImageGenerator, times: [CMTime]), Error> {
+	) -> Result<(generator: AVAssetImageGenerator, times: [CMTime], fps: Int), Error> {
```
sindresorhus, line 228: I think this could just use `throws` instead of `Result`. I generally only use `Result` for async or when I need to preserve the Error type.

sindresorhus (changes requested):

Gifski/EditVideoViewController.swift
```diff
@@ -434,6 +497,14 @@ final class EditVideoViewController: NSViewController {
 		}
 		.tieToLifetime(of: self)
 
+		playerRateObserver = playerViewController.playerView.player?.observe(\.rate, options: [.new]) { [weak self] _, change in
+			if change.newValue != 0 {
```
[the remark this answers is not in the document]
ejensen: Updated to use `guard` and changed the logic to only show the warning once the video begins playing in reverse (instead of whenever the video starts playing and the bounce option is enabled)

Gifski/EditVideoViewController.swift
```diff
@@ -324,6 +325,68 @@ final class EditVideoViewController: NSViewController {
...
+		DispatchQueue.global(qos: .utility).async { [weak self] in
+			guard
+				Defaults[.bounceGif],
+				let asset = self?.asset,
+				let videoTrack = asset.firstVideoTrack,
+				let reader = try? AVAssetReader(asset: asset)
+			else {
+				return
+			}
+
```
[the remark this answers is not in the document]
ejensen: I moved the logic to an extension on `AVAssetTrack`. I made the method synchronous since the logic itself is synchronous. The caller of the method can dispatch to a background queue if needed.

Gifski/Gifski.swift
ejensen: I generally agree, but In this case, `Error` is the specialized `Gifski.Error` not the general `Swift.Error`. Switching it to a throwing method loses the specialized Error type that its caller expects.

Gifski/EditVideoViewController.swift
```diff
@@ -425,7 +453,23 @@ final class EditVideoViewController: NSViewController {
...
+			self?.estimateFileSize()
+		}
+		.tieToLifetime(of: self)
+
+		playerRateObserver = playerViewController.playerView.player?.observe(\.rate, options: [.new]) { [weak self] _, change in
+			guard change.newValue == -1 else {
+				return
+			}
+
+			SSApp.runOnce(identifier: "lowKeyframeRateWarning") {
```
sindresorhus, line 469: I just realized that this will only show the warning if the first video we shows that plays in reverse has low key frame rate, otherwise `showKeyframeRateWarningIfNeeded()` will never be hit. So in the likely scenario, most users will not see the warning even if they eventually load a video with a low key frame rate.

ejensen: Commit becc38f changes the behavior so the warning is shown once per-conversion that has a low keyframe rate. And includes the option to suppress the warning for all future conversions:
alert-with-suppression

sindresorhus:
This looks great. Thank you so much for working on this. 🙌

PR #241, opened 2021-05-06
Use NSViewController.present() with custom animator #108

### LINKED ISSUE #108: Use `NSViewController.present()` with custom animator
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/gifski-app/issues/108)

Currently, we do this: https://github.com/sindresorhus/Gifski/blob/56511ac4c7deb68b2e39490e0bef45180634b128/Gifski/util.swift#L2436-L2455

The benefit of using `present` is that it's possible to dismiss the view controller to go back to the previous one. For example, when the user cancels the editing, we could just dismiss that view controller.

https://developer.apple.com/documentation/appkit/nsviewcontroller/1434431-present
https://stackoverflow.com/questions/28454291/transitioning-between-view-controller-os-x/28493373#28493373

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $60.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($60.00)

### Submitted pull Requests
- [#139 Use custom presentation animator](https://issuehunt.io/r/sindresorhus/gifski-app/pull/139)
- [#241 Use NSViewController.present() with custom animator #108](https://issuehunt.io/r/sindresorhus/gifski-app/pull/241)
---

### Tips

</details>

hazeycode:
While this still changes the contentViewController, presenting view controllers behave exactly like a stack. Works flawlessly (in my testing) with dismiss:

Fixes #108

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#108: Use `NSViewController.present()` with custom animator](https://issuehunt.io/repos/119822304/issues/108)
---
</details>

### DIFF
--- a/Gifski/ConversionCompletedViewController.swift
+++ b/Gifski/ConversionCompletedViewController.swift
@@ -70,6 +70,7 @@ final class ConversionCompletedViewController: NSViewController {
 				return
 			}
 
+			// Prevents a rare situation where the this closure is invoked after user has clicked back or dropped a new video file
 			guard self.view.window?.contentViewController == self else {
 				return
 			}
--- a/Gifski/ConversionCompletedViewController.swift
+++ b/Gifski/ConversionCompletedViewController.swift
@@ -70,6 +70,10 @@ final class ConversionCompletedViewController: NSViewController {
 				return
 			}
 
+			guard self.view.window?.contentViewController == self else {
+				return
+			}
+
 			self.tooltip.show(from: self.draggableFile, preferredEdge: .maxY)
 		}
 
@@ -192,10 +196,7 @@ final class ConversionCompletedViewController: NSViewController {
 
 	@IBAction
 	private func backButton(_ sender: NSButton) {
-		// It's safe to force-unwrap as there's no scenario where it will be nil.
-		let viewController = AppDelegate.shared.previousEditViewController!
-		viewController.isConverting = false
-		push(viewController: viewController)
+		presentingViewController?.presentingViewController?.dismiss(self)
 	}
 }
 
--- a/Gifski/ConversionCompletedViewController.swift
+++ b/Gifski/ConversionCompletedViewController.swift
@@ -70,7 +70,8 @@ final class ConversionCompletedViewController: NSViewController {
 				return
 			}
 
-			// Prevents a rare situation where the this closure is invoked after user has clicked back or dropped a new video file
+			// Prevents the following tooltip code from failing in a rare situation where
+			// this closure is invoked after user has clicked back or dropped a new video file
 			guard self.view.window?.contentViewController == self else {
 				return
 			}
--- a/Gifski/ConversionViewController.swift
+++ b/Gifski/ConversionViewController.swift
@@ -57,7 +57,9 @@ final class ConversionViewController: NSViewController {
 
 	/// Gets called when the Esc key is pressed.
 	override func cancelOperation(_ sender: Any?) {
-		cancelConversion()
+		if isRunning {
+			cancelConversion()
+		}
 	}
 
 	private func start(conversion: Gifski.Conversion) {
@@ -116,12 +118,8 @@ final class ConversionViewController: NSViewController {
 			progress?.cancel()
 		}
 
-		stopConversion { [weak self] in
-			// It's safe to force-unwrap as there's no scenario where it will be nil.
-			let viewController = AppDelegate.shared.previousEditViewController!
-			viewController.isConverting = false
-
-			self?.push(viewController: viewController)
+		stopConversion { [self] in
+			self.presentingViewController?.dismiss(self)
 		}
 	}
 
--- a/Gifski/ConversionViewController.swift
+++ b/Gifski/ConversionViewController.swift
@@ -118,8 +118,8 @@ final class ConversionViewController: NSViewController {
 			progress?.cancel()
 		}
 
-		stopConversion { [self] in
-			self.presentingViewController?.dismiss(self)
+		stopConversion { [weak self] in
+			self?.pop()
 		}
 	}
 
--- a/Gifski/EditVideoViewController.swift
+++ b/Gifski/EditVideoViewController.swift
@@ -90,8 +90,6 @@ final class EditVideoViewController: NSViewController {
 		self.inputUrl = inputUrl
 		self.asset = asset
 		self.videoMetadata = videoMetadata
-
-		AppDelegate.shared.previousEditViewController = self
 	}
 
 	var isConverting = false
@@ -108,9 +106,7 @@ final class EditVideoViewController: NSViewController {
 
 	@IBAction
 	private func cancel(_ sender: Any) {
-		let videoDropController = VideoDropViewController()
-		push(viewController: videoDropController)
-		AppDelegate.shared.previousEditViewController = nil
+		presentingViewController?.dismiss(self)
 	}
 
 	override func viewDidLoad() {
@@ -125,6 +121,10 @@ final class EditVideoViewController: NSViewController {
 		setUpLoopCountControls()
 		setUpDropView()
 		setUpTrimmingView()
+	}
+
+	override func viewWillAppear() {
+		isConverting = false
 	}
 
 	override func viewDidAppear() {
--- a/Gifski/EditVideoViewController.swift
+++ b/Gifski/EditVideoViewController.swift
@@ -106,7 +106,7 @@ final class EditVideoViewController: NSViewController {
 
 	@IBAction
 	private func cancel(_ sender: Any) {
-		presentingViewController?.dismiss(self)
+		pop()
 	}
 
 	override func viewDidLoad() {
--- a/Gifski/MainWindowController.swift
+++ b/Gifski/MainWindowController.swift
@@ -124,7 +124,7 @@ final class MainWindowController: NSWindowController {
 			return
 		}
 
-		rootVideoDropViewController.popAll {
+		rootVideoDropViewController.popToRootViewController(animated: false) {
 			let editController = EditVideoViewController(inputUrl: inputUrl, asset: asset, videoMetadata: videoMetadata)
 			self.rootVideoDropViewController.push(viewController: editController)
 		}
--- a/Gifski/MainWindowController.swift
+++ b/Gifski/MainWindowController.swift
@@ -2,6 +2,8 @@ import Cocoa
 
 final class MainWindowController: NSWindowController {
 	private let videoValidator = VideoValidator()
+	
+	public let rootVideoDropViewController = VideoDropViewController();
 
 	var isConverting: Bool {
 		window?.contentViewController is ConversionViewController
@@ -58,9 +60,10 @@ final class MainWindowController: NSWindowController {
 
 	convenience init() {
 		let window = NSWindow.centeredWindow(size: .zero)
-		window.contentViewController = VideoDropViewController()
 		window.centerNatural()
 		self.init(window: window)
+
+		window.contentViewController = rootVideoDropViewController
 
 		with(window) {
 			$0.delegate = self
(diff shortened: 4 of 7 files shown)


sindresorhus:
Thanks for working on this.

You have a lot of trailing whitespace. Make sure `swiftlint` passes without any warnings.

sindresorhus:
Cancelling a conversion no longer correctly works:

Screen Shot 2021-05-06 at 16 14 02

hazeycode:
Sorry about that @sindresorhus, I have fixed the cancellation state bug that I introduced and tested the conversion state cancelations and completions more thoroughly.

Gifski/ConversionViewController.swift
```diff
@@ -116,19 +118,15 @@ final class ConversionViewController: NSViewController {
...
-
-			self?.push(viewController: viewController)
+		stopConversion { [self] in
+			self.presentingViewController?.dismiss(self)
 		}
 	}
 
 	private func didComplete(conversion: Gifski.Conversion, gifUrl: URL) {
 		let conversionCompleted = ConversionCompletedViewController(conversion: conversion, gifUrl: gifUrl)
-		stopConversion { [weak self] in
```
sindresorhus, line 130: Why are you removing `weak`?

Gifski/EditVideoViewController.swift
```diff
@@ -108,9 +106,7 @@ final class EditVideoViewController: NSViewController {
 
 	@IBAction
 	private func cancel(_ sender: Any) {
-		let videoDropController = VideoDropViewController()
-		push(viewController: videoDropController)
-		AppDelegate.shared.previousEditViewController = nil
+		self.presentingViewController?.dismiss(self)
```
sindresorhus, line 109: You don't need `.self` here.

Gifski/Utilities.swift
```diff
@@ -3001,45 +3001,89 @@ extension URL {
...
+			window.contentViewController = fromViewController
+			fromViewController.view.animator().alphaValue = 1.0
 		})
 	}
+}
+
+extension NSViewController {
+	func push(viewController: NSViewController, completion: (() -> Void)? = nil) {
+		let animator = ViewControllerPresentationAnimator(completion: completion)
+		self.present(viewController, animator: animator)
```
sindresorhus, line 3085: You don't need `self`.

Gifski/ConversionViewController.swift
hazeycode: This was an incorrect fix for a bug where the delayed tooltip on ConversionCompletedViewController could throw an error under unlikely user behaviour. Steps to reproduce:

1. Start a conversion
2. Hit escape exactly during the transition from ConversionViewController to ConversionCompletedViewController (spamming escape as soon as transition begins) 
3. Click back button

I have found a correct fix for this and will update shortly

Gifski/ConversionCompletedViewController.swift
```diff
@@ -70,6 +70,10 @@ final class ConversionCompletedViewController: NSViewController {
 				return
 			}
 
+			guard self.view.window?.contentViewController == self else {
+				return
+			}
+
 			self.tooltip.show(from: self.draggableFile, preferredEdge: .maxY)
```
hazeycode, line 77: This is the fix for the tooltip crash. It's an odd one and quite difficult to reproduce.

Gifski/ConversionViewController.swift
```diff
@@ -116,12 +118,8 @@ final class ConversionViewController: NSViewController {
 			progress?.cancel()
 		}
 
-		stopConversion { [weak self] in
-			// It's safe to force-unwrap as there's no scenario where it will be nil.
-			let viewController = AppDelegate.shared.previousEditViewController!
-			viewController.isConverting = false
-
-			self?.push(viewController: viewController)
+		stopConversion { [self] in
```
sindresorhus, line 121: Same as https://github.com/sindresorhus/Gifski/pull/241#discussion_r627336411

Gifski/ConversionCompletedViewController.swift
```diff
@@ -192,10 +196,7 @@ final class ConversionCompletedViewController: NSViewController {
 
 	@IBAction
 	private func backButton(_ sender: NSButton) {
-		// It's safe to force-unwrap as there's no scenario where it will be nil.
-		let viewController = AppDelegate.shared.previousEditViewController!
-		viewController.isConverting = false
-		push(viewController: viewController)
+		self.presentingViewController?.presentingViewController?.dismiss(self)
```
sindresorhus, line 199: You don't need `self` here.

sindresorhus:
If you drag and drop a video, then drag and drop another video onto the editor, and then press "Cancel", you end up with this:

Screen Shot 2021-05-07 at 00 18 52

sindresorhus:
This may or may not be the same issue as the above: If you drag and drop a video, convert it, and on the success screen, drag and drop another video, and then press Cancel, you end up with:

Screen Shot 2021-05-07 at 00 22 44

sindresorhus:
Have you tested this change on macOS 10.15 or 10.14?

Gifski/ConversionCompletedViewController.swift
```diff
@@ -70,6 +70,10 @@ final class ConversionCompletedViewController: NSViewController {
 				return
 			}
 
+			guard self.view.window?.contentViewController == self else {
```
sindresorhus, line 73: I feel like this is just hiding a different issue. `self` is already weak at this point, so the only way this can happen is if this controller is retained while not being the current one.

sindresorhus (changes requested):

Gifski/ConversionCompletedViewController.swift
hazeycode: This is true. There seems to be some time during the transition that this can be invoked but before ARC has cleanup up the popped VC

sindresorhus: If there's really no better solution than this, it should at least have a code comment about this.

Gifski/ConversionViewController.swift
hazeycode: This one is a little different. The strong reference is required here so that the closure keeps ConversionViewController from being released during the transition.
There might be a better way that I haven't found yet to ensure this without requiring user code to keep a ref

Gifski/ConversionCompletedViewController.swift
hazeycode: I'm not sure why viewDidAppear is getting called at all on this view controller after clicking back. Will do some more digging

Gifski/ConversionViewController.swift
hazeycode: Must be the change in ownership model, before all view controllers ownership traced back to the edit view controller held by App

Gifski/MainWindowController.swift
```diff
@@ -121,8 +124,10 @@ final class MainWindowController: NSWindowController {
 			return
 		}
 
-		let editController = EditVideoViewController(inputUrl: inputUrl, asset: asset, videoMetadata: videoMetadata)
-		window?.contentViewController?.push(viewController: editController)
+		rootVideoDropViewController.popAll {
+			let editController = EditVideoViewController(inputUrl: inputUrl, asset: asset, videoMetadata: videoMetadata)
+			self.rootVideoDropViewController.push(viewController: editController)
+		}
```
hazeycode, line 130: Fix for https://github.com/sindresorhus/Gifski/pull/241#issuecomment-833701277

Gifski/VideoDropViewController.swift
```diff
@@ -56,7 +56,13 @@ final class VideoDropViewController: NSViewController {
...
-		let editController = EditVideoViewController(inputUrl: inputUrl, asset: asset, videoMetadata: videoMetadata)
-		push(viewController: editController)
+		guard let rootVideoDropViewController = (view.window?.windowController as? MainWindowController)?.rootVideoDropViewController else {
+			return
+		}
+
+		popAll {
+			let editController = EditVideoViewController(inputUrl: inputUrl, asset: asset, videoMetadata: videoMetadata)
+			rootVideoDropViewController.push(viewController: editController)
+		}
```
hazeycode, line 66: Fix for https://github.com/sindresorhus/Gifski/pull/241#issuecomment-833703219

hazeycode:
I only have macOS version 11.2.3 for testing at this time

Gifski/Utilities.swift
```diff
@@ -3001,44 +3001,78 @@ extension URL {
...
+	})
+}
+
+extension NSViewController {
+	func push(viewController: NSViewController, completion: (() -> Void)? = nil) {
+		let animator = ViewControllerPresentationAnimator(completion: completion)
+		present(viewController, animator: animator)
+	}
+
+	func popAll(completion: (() -> Void)? = nil) {
```
sindresorhus, line 3068: Can you follow the `UINavigationController` naming? `popToRootViewController(animated:)`

sindresorhus:
If I drag in a video, then drag another one into the editor, the editor animates slightly in size. This is new behavior. It should only fade between the view. Try out the App Store version to see the behavior difference.

---

You have some swiftlint warnings.

Gifski/ConversionCompletedViewController.swift
```diff
@@ -192,10 +198,7 @@ final class ConversionCompletedViewController: NSViewController {
 
 	@IBAction
 	private func backButton(_ sender: NSButton) {
-		// It's safe to force-unwrap as there's no scenario where it will be nil.
-		let viewController = AppDelegate.shared.previousEditViewController!
-		viewController.isConverting = false
-		push(viewController: viewController)
+		presentingViewController?.presentingViewController?.dismiss(self)
```
sindresorhus, line 201: Use `.pop()` here too for consistency?

Gifski/MainWindowController.swift
```diff
@@ -121,8 +124,10 @@ final class MainWindowController: NSWindowController {
 			return
 		}
 
-		let editController = EditVideoViewController(inputUrl: inputUrl, asset: asset, videoMetadata: videoMetadata)
-		window?.contentViewController?.push(viewController: editController)
+		rootVideoDropViewController.popToRootViewController(animated: false) {
+			let editController = EditVideoViewController(inputUrl: inputUrl, asset: asset, videoMetadata: videoMetadata)
+			self.rootVideoDropViewController.push(viewController: editController)
```
sindresorhus, line 129: Should `self` be weak here?

Gifski/Utilities.swift
```diff
@@ -3001,44 +3001,82 @@ extension URL {
...
+	
+	func pop() {
+		presentingViewController?.dismiss(self)
+	}
+
+	func popToRootViewController(animated: Bool = true, completion: (() -> Void)? = nil) {
+		guard let window = view.window else {
+			return
+		}
+		guard let rootVC = (window.windowController as? MainWindowController)?.rootVideoDropViewController else {
```
sindresorhus, line 3076: You are mixing reusable utilities with app-specific logic.

sindresorhus:
Can you use the Xcode memory graph to ensure there are no retain cycles?

https://medium.com/zendesk-engineering/ios-identifying-memory-leaks-using-the-xcode-memory-graph-debugger-e84f097b9d15

I'm a little bit paranoid that this change introduces subtle retain cycles. I spent years fixing retain cycles in Gifski.

hazeycode:
I think moving the transition code out of the NSViewController extension in Utilities into a RootViewController would resolve the issues here. RootViewController would take responsibilty for the lifetimes and transitions between view controllers and would be the only window contentViewController.

sindresorhus:
> I think moving the transition code out of the NSViewController extension in Utilities into a RootViewController would resolve the issues here.

Why would it have to be moved out the extension. We could still have a root view controller that manages everything, while having the reusable transition logic in the extension.

hazeycode:
> Why would it have to be moved out the extension? We could still have a root view controller that manages everything, while having the reusable transition logic in the extension.

I guess it doesn't need to be, so long as we have a way to tell which view controller is root, which should be straight forward if the window's contentViewController is always the root view controller. I'll have another look at this when I've got some more time

hazeycode:
> Why would it have to be moved out the extension? We could still have a root view controller that manages everything, while having the reusable transition logic in the extension.

I guess it doesn't have to be. The problem is relying on NSViewController and it's friends to keep things alive, so far my attempts have resulted in reference cycles or things getting released too early.

sindresorhus:
I'm closing this. This is taking forever and the current code works fine. This was just meant to clean up the code, but it's really not worth the risk of breaking things more. We'll eventually move the whole app to SwiftUI at some point anyway.

Issue #190, opened 2020-04-13
Trim empty frames from the start/end of videos when opening them


sindresorhus:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2454%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/gifski-app/issues/190)

Some videos contain one or more empty frames at the start.

Example: [simulator-recording.mp4.zip](https://github.com/sindresorhus/Gifski/files/4469007/simulator-recording.mp4.zip) (unzip and drag it into Gifski)

Gifski currently handles this gracefully when generating frames (https://github.com/sindresorhus/Gifski/commit/be7d3797027a297203f471af137978f9e2d8125e), but the empty frames still show up in the editor view as black frames. It would be nice to trim the video before showing it in the editor.

I think we could use `AVAssetReader` for this. Just read the frame buffers until they're no longer empty to get the trim range, trim the video, and then generate a new `AVMutableComposition` out of it.

More info: https://github.com/sindresorhus/Gifski/commit/be7d3797027a297203f471af137978f9e2d8125e#diff-75a4e71e51a82ac8607443b8f202430eR333

---

Note: This is not an easy issue. It's expected that you have advanced Swift knowledge and that you do a good job with this. The point of bounties is to save us time, not drag us down in a way too long-running pull request.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $54.00)

- [<img src='https://avatars1.githubusercontent.com/u/170270?v=4' alt='sindresorhus' width=24 height=24> sindresorhus](https://issuehunt.io/u/sindresorhus) ($54.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/gifski-app/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

PR #245, opened 2021-06-17
Trim empty frames in preview

codetheweb:
Closes https://github.com/sindresorhus/Gifski/issues/190.

Test file (has black frame in preview with current version of Gifski): 

https://user-images.githubusercontent.com/7410405/122309868-bcbfb100-ced4-11eb-9175-9b788dfe53f1.mp4

**Note**: this does not currently handle blank frames at the **end** of files. From my understanding, we would have to load all samples (`while let sampleBuffer = readerOutput.copyNextSampleBuffer()`) into memory to be able to iterate backwards through them. Happy to implement that as well if you think it's an acceptable trade-off.


Gifski/VideoValidator.swift
```diff
@@ -184,6 +184,47 @@ struct VideoValidator {
...
 
-		return .success(newAsset, newVideoMetadata)
+		// Find first non-empty frame
+		do {
+			let reader = try AVAssetReader(asset: newAsset)
+			let readerOutput = AVAssetReaderTrackOutput(track: newAsset.firstVideoTrack!, outputSettings: nil)
+			reader.add(readerOutput)
+			reader.startReading()
+
+			while let sampleBuffer = readerOutput.copyNextSampleBuffer() {
```
sindresorhus, line 194: Per docs, you need to check the `.status` property, and handle the different cases properly.

sindresorhus:
> From my understanding, we would have to load all samples (while let sampleBuffer = readerOutput.copyNextSampleBuffer()) into memory to be able to iterate backwards through them. Happy to implement that as well if you think it's an acceptable trade-off.

It's unclear to me whether or not this is a problem. From my understanding, the samples are not actually the decoded frames, but rather just metadata, so it should be pretty efficient. Needs to be tested though, but I would like to support trimming empty frames from the end too as that's more common than empty frames at the start.

sindresorhus (changes requested):

Gifski/VideoValidator.swift
```diff
@@ -184,6 +184,47 @@ struct VideoValidator {
...
+				// On first non-empty frame
+				if sampleBuffer.totalSampleSize != 0 {
```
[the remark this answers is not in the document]
codetheweb: The `.isValid` property does [not check if the frame is empty](https://developer.apple.com/documentation/coremedia/1489151-cmsamplebufferisvalid) (tested and returns true when `totalSampleSize == 0`.

codetheweb:
> It's unclear to me whether or not this is a problem. From my understanding, the samples are not actually the decoded frames, but rather just metadata, so it should be pretty efficient.

I thought about this some more and not sure what I was talking about before. We can step through the whole file without loading all samples into memory...

> Needs to be tested though, but I would like to support trimming empty frames from the end too as that's more common than empty frames at the start. And if we do trimming from the end, we could trim empty frames in the middle of the video too.

From what I've seen empty frames at the beginning of video files (i.e. simulator screen recordings) is caused by the start time metadata marker being negative / non zero (for example, it's `start_time=0.245000` for the example I posted (`ffprobe -i test.mp4 -show_format`)). Do you have an example of a file with blank frames at the end?

I apologize for the number of changes required; I know in the issue you requested that the PR be 99% done when opened.

sindresorhus:
I think this one has one or more blank frames at the end: [Gifski.mp4.zip](https://github.com/sindresorhus/Gifski/files/6681948/Gifski.mp4.zip)

codetheweb:
It looks fine to me in both QuickTime and the current version of Gifski:

Screen Shot 2021-06-20 at 1 30 58 PM

Screen Shot 2021-06-20 at 1 31 10 PM

sindresorhus:
If you play it in Quick Look, it's empty at the end, but I guess, it's just that the video track ends earlier than the actual video, which Gifski already handles by rewrapping the video track into a fresh asset.

I'm sure some users are having problems with empty frames at the end as I have seen it in encode failure errors on Crashlytics, but I don't think I actually have a fixture for this.

Would you be able to make a fixture for this? Maybe take the video with the empty leading frames and move those frames to the end or something.

codetheweb:
> Would you be able to make a fixture for this? Maybe take the video with the empty leading frames and move those frames to the end or something.

I would if I could, but as mentioned before as far as I can tell it's only possible to have blank / empty frames at the beginning of tracks, not the end (assuming the video track in a file gets moved into its own asset like Gifski does to prevent interference from audio tracks). It's easy to prepend / append a few frames of black to files with ffmpeg, but that's not the same thing as a blank frame.

sindresorhus:
I think you are right. Let's go with only trimming at the start then.

Gifski/VideoValidator.swift
```diff
@@ -184,6 +184,47 @@ struct VideoValidator {
...
+					let currentTimestamp = CMSampleBufferGetOutputPresentationTimeStamp(sampleBuffer)
```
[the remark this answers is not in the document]
codetheweb: Fixed in 9455603.

Gifski/VideoValidator.swift
```diff
@@ -184,6 +184,47 @@ struct VideoValidator {
...
+
+					// Add video track to composition and remove time range (from the begining up to current frame)
+					let composition = AVMutableComposition()
+					let videoCompTrack = composition.addMutableTrack(withMediaType: AVMediaType.video, preferredTrackID: CMPersistentTrackID())
+
+					try videoCompTrack?.insertTimeRange(CMTimeRangeMake(start: .zero, duration: newAsset.duration), of: newAsset.firstVideoTrack!, at: .zero)
+					videoCompTrack?.removeTimeRange(CMTimeRange(start: .zero, end: currentTimestamp))
+
+					reader.cancelReading()
```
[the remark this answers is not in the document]
codetheweb: Fixed in 9455603.

Gifski/VideoValidator.swift
```diff
@@ -184,6 +184,47 @@ struct VideoValidator {
 			return .failure
 		}
 
-		return .success(newAsset, newVideoMetadata)
+		// Find first non-empty frame
+		do {
+			let reader = try AVAssetReader(asset: newAsset)
+			let readerOutput = AVAssetReaderTrackOutput(track: newAsset.firstVideoTrack!, outputSettings: nil)
```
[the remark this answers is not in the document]
codetheweb: Fixed in 9455603.

Gifski/VideoValidator.swift
```diff
@@ -184,6 +184,47 @@ struct VideoValidator {
...
+			while let sampleBuffer = readerOutput.copyNextSampleBuffer() {
+				// On first non-empty frame
+				if sampleBuffer.totalSampleSize != 0 {
+					let currentTimestamp = CMSampleBufferGetOutputPresentationTimeStamp(sampleBuffer)
+
+					// Add video track to composition and remove time range (from the begining up to current frame)
+					let composition = AVMutableComposition()
+					let videoCompTrack = composition.addMutableTrack(withMediaType: AVMediaType.video, preferredTrackID: CMPersistentTrackID())
+
+					try videoCompTrack?.insertTimeRange(CMTimeRangeMake(start: .zero, duration: newAsset.duration), of: newAsset.firstVideoTrack!, at: .zero)
```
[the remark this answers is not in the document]
codetheweb: Fixed in 9455603.

Gifski/VideoValidator.swift
```diff
@@ -184,6 +184,47 @@ struct VideoValidator {
...
+			reader.startReading()
+
+			while let sampleBuffer = readerOutput.copyNextSampleBuffer() {
+				// On first non-empty frame
+				if sampleBuffer.totalSampleSize != 0 {
+					let currentTimestamp = CMSampleBufferGetOutputPresentationTimeStamp(sampleBuffer)
+
+					// Add video track to composition and remove time range (from the begining up to current frame)
+					let composition = AVMutableComposition()
+					let videoCompTrack = composition.addMutableTrack(withMediaType: AVMediaType.video, preferredTrackID: CMPersistentTrackID())
```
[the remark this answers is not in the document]
codetheweb: Fixed in 9455603.

Gifski/VideoValidator.swift
codetheweb: Fixed in 9455603.

Gifski/Utilities.swift
```diff
@@ -592,6 +592,54 @@ extension AVAsset {
 		imageGenerator.requestedTimeToleranceBefore = .zero
 		return imageGenerator.image(at: time)
 	}
+
+	func trimBlankFrames() -> AVAsset? {
+		guard let reader = try? AVAssetReader(asset: self) else {
+			return nil
```
codetheweb, line 598: I followed the existing pattern of returning optionals; but would we want to throw in this extension instead?

sindresorhus: Yes, the function should be throwing.

Gifski/Utilities.swift
```diff
@@ -592,6 +592,54 @@ extension AVAsset {
...
+		reader.startReading()
+
+		while let sampleBuffer = readerOutput.copyNextSampleBuffer() {
+			if reader.status != .reading {
+				reader.cancelReading()
+				return nil
+			}
+
+			// On first non-empty frame
+			if sampleBuffer.totalSampleSize != 0 {
```
sindresorhus, line 616: Use `guard`

Gifski/Utilities.swift
```diff
@@ -592,6 +592,54 @@ extension AVAsset {
...
+		guard let videoTrack = self.firstVideoTrack else {
+			return nil
+		}
+
+		let readerOutput = AVAssetReaderTrackOutput(track: videoTrack, outputSettings: nil)
+		reader.add(readerOutput)
+		reader.startReading()
+
+		while let sampleBuffer = readerOutput.copyNextSampleBuffer() {
+			if reader.status != .reading {
```
sindresorhus, line 610: Use `guard`

Gifski/Utilities.swift
```diff
@@ -592,6 +592,54 @@ extension AVAsset {
...
+
+				// Add video track to composition and remove time range (from the begining up to current frame)
+				let composition = AVMutableComposition()
+				guard let videoCompTrack = composition.addMutableTrack(withMediaType: .video, preferredTrackID: CMPersistentTrackID()) else {
+					reader.cancelReading()
+					return nil
+				}
+
+				do {
+					try videoCompTrack.insertTimeRange(CMTimeRangeMake(start: .zero, duration: self.duration), of: videoTrack, at: .zero)
```
sindresorhus, line 627: Use the modern initializer, not `CMTimeRangeMake`.

Gifski/Utilities.swift
```diff
@@ -592,6 +592,54 @@ extension AVAsset {
 		imageGenerator.requestedTimeToleranceBefore = .zero
 		return imageGenerator.image(at: time)
 	}
+
+	func trimBlankFrames() -> AVAsset? {
```
sindresorhus, line 596: I think there should be a method on `AVAssetTrack` (with the same name) that contains the main logic and this method should just use it.

sindresorhus, line 596: And this needs a doc comment.

Gifski/Utilities.swift
codetheweb: Fixed in a3a64ed.

Gifski/Utilities.swift
codetheweb: Fixed in a3a64ed.

Gifski/Utilities.swift
codetheweb: Fixed in a3a64ed.

Gifski/Utilities.swift
codetheweb: Fixed in a3a64ed.

codetheweb: Makes sense, added in a3a64ed.

Gifski/Utilities.swift
codetheweb: Fixed in a3a64ed.

Gifski/Utilities.swift
```diff
@@ -584,6 +584,16 @@ extension AVAssetImageGenerator {
 }
 
 
+enum VideoTrimmingError: Error {
```
codetheweb, line 587: Wasn't sure how specific / detailed you wanted the error types to be.

Gifski/Utilities.swift
```diff
@@ -758,6 +787,63 @@ extension AVAssetTrack {
...
+				throw VideoTrimmingError.assetReaderFailure
+			}
+
+			// On first non-empty frame
+			guard sampleBuffer.totalSampleSize == 0 else {
+				let currentTimestamp = sampleBuffer.outputPresentationTimeStamp
+
+				// Add video track to composition and remove time range (from the begining up to current frame)
+				let composition = AVMutableComposition()
+				guard let videoCompTrack = composition.addMutableTrack(withMediaType: .video, preferredTrackID: CMPersistentTrackID()) else {
```
sindresorhus, line 819: Not sure what the intention is with `CMPersistentTrackID()`? You could just write `0` if that's what you meant.

Gifski/Utilities.swift
```diff
@@ -758,6 +787,63 @@ extension AVAssetTrack {
...
+					throw error
+				}
+
+				videoCompTrack.removeTimeRange(CMTimeRange(start: .zero, end: currentTimestamp))
+
+				reader.cancelReading()
+
+				guard let trimmedTrack = AVPlayerItem(asset: composition).asset.firstVideoTrack else {
+					throw VideoTrimmingError.compositionIsMissingTrack
+				}
```
sindresorhus, line 837: I'm not sure what's going on here. You could just return:

```swift
composition.asset.firstVideoTrack
```

Gifski/Utilities.swift
```diff
@@ -758,6 +787,63 @@ extension AVAssetTrack {
 	}
 }
 
+extension AVAssetTrack {
```
sindresorhus, line 790: I think both these extension should live together in the file.

Gifski/Utilities.swift
```diff
@@ -758,6 +787,63 @@ extension AVAssetTrack {
...
+				do {
+					try videoCompTrack.insertTimeRange(.init(start: .zero, duration: asset.duration), of: self, at: .zero)
+				} catch {
+					reader.cancelReading()
+					throw error
+				}
+
+				videoCompTrack.removeTimeRange(CMTimeRange(start: .zero, end: currentTimestamp))
+
+				reader.cancelReading()
```
sindresorhus, line 833: I think you can just put this inside a `defer` at the top of the scope.

Gifski/Utilities.swift
```diff
@@ -592,6 +602,25 @@ extension AVAsset {
...
 		return imageGenerator.image(at: time)
 	}
+
+	/**
+	Removes blank frames from the begining of the asset.
+
+	This can be useful to trim blank frames from files produced by tools like the iOS simulator screen recorder.
+	*/
+	func trimBlankFrames() throws -> AVAsset {
+		guard let videoTrack = self.firstVideoTrack else {
```
sindresorhus, line 612:
```suggestion
		guard let videoTrack = firstVideoTrack else {
```

Gifski/Utilities.swift
```diff
@@ -758,6 +787,63 @@ extension AVAssetTrack {
...
 }
 
+extension AVAssetTrack {
+	/**
+	Removes blank frames from the begining of the track.
+
+	This can be useful to trim blank frames from files produced by tools like the iOS simulator screen recorder.
+	*/
+	func trimBlankFrames() throws -> AVAssetTrack {
+		guard let asset = self.asset else {
```
sindresorhus, line 797:
```suggestion
		guard let asset = asset else {
```

Gifski/VideoValidator.swift
```diff
@@ -184,6 +184,20 @@ struct VideoValidator {
...
-		return .success(newAsset, newVideoMetadata)
+		// Trim asset
+		do {
+			let trimmedAsset = try newAsset.trimBlankFrames()
+
+			return .success(trimmedAsset, newVideoMetadata)
+		} catch {
+			NSAlert.showModalAndReportToCrashlytics(
+				for: window,
+				title: "Could not trim empty frames from video.",
```
sindresorhus, line 195:
```suggestion
				title: "Could not trim empty leading frames from video.",
```

Gifski/VideoValidator.swift
```diff
@@ -184,6 +184,20 @@ struct VideoValidator {
 			return .failure
 		}
 
-		return .success(newAsset, newVideoMetadata)
+		// Trim asset
+		do {
+			let trimmedAsset = try newAsset.trimBlankFrames()
+
+			return .success(trimmedAsset, newVideoMetadata)
+		} catch {
```
sindresorhus, line 192: You should not ignore the error. It should be included in the alert.

sindresorhus (changes requested):

Gifski/VideoValidator.swift
codetheweb: Fixed in 50532fc.

Gifski/VideoValidator.swift
codetheweb: Fixed in 50532fc.

Gifski/Utilities.swift
codetheweb: Fixed in 50532fc.

Gifski/Utilities.swift
```diff
@@ -608,22 +598,87 @@ extension AVAsset {
...
-			throw VideoTrimmingError.assetIsMissingTrack
+			throw AVAssetTrack.VideoTrimmingError.assetIsMissingTrack
 		}
 
-		let trimmedTrack = try videoTrack.trimBlankFrames()
+		let trimmedTrack = try videoTrack.trimmingBlankFrames()
 
 		guard let trimmedAsset = trimmedTrack.asset else {
-			throw VideoTrimmingError.assetIsNil
+			throw AVAssetTrack.VideoTrimmingError.assetIsNil
```
sindresorhus, line 609: You can add a typealias for the error to `AVAsset`.

Gifski/Utilities.swift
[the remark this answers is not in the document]
codetheweb: Good suggestion, turned out much cleaner: 50532fc.

Gifski/Utilities.swift
```diff
@@ -758,6 +787,63 @@ extension AVAssetTrack {
...
+		let reader = try AVAssetReader(asset: asset)
+
+		let readerOutput = AVAssetReaderTrackOutput(track: self, outputSettings: nil)
+		reader.add(readerOutput)
+		reader.startReading()
+
+		while let sampleBuffer = readerOutput.copyNextSampleBuffer() {
+			guard reader.status == .reading else {
+				reader.cancelReading()
+				throw VideoTrimmingError.assetReaderFailure
```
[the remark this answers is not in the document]
codetheweb: Fixed in 50532fc.

Gifski/Utilities.swift
```diff
@@ -608,22 +598,87 @@ extension AVAsset {
...
+		throw VideoTrimmingError.assetIsEmpty
+	}
+}
+
+
+extension AVAssetTrack.VideoTrimmingError: LocalizedError {
+	public var errorDescription: String? {
+		self.rawValue
+	}
+}
```
sindresorhus: This is an abuse of `LocalizedError`. You need to actually do a switch where you write human-readable messages for each error.

Gifski/Utilities.swift
```diff
@@ -592,6 +592,90 @@ extension AVAsset {
...
+
+
+extension AVAssetTrack {
+	enum VideoTrimmingError: String, Error {
+		case assetIsNil
+		case assetReaderFailure
+		case assetIsEmpty
+		case assetIsMissingTrack
+		case compositionCouldNotBeCreated
+		case compositionIsMissingTrack
```
sindresorhus, line 624: This is unused.

Gifski/VideoValidator.swift
```diff
@@ -184,6 +184,20 @@ struct VideoValidator {
...
+		// Trim asset
+		do {
+			let trimmedAsset = try newAsset.trimmingBlankFrames()
+
+			return .success(trimmedAsset, newVideoMetadata)
+		} catch {
+			NSAlert.showModalAndReportToCrashlytics(
+				for: window,
+				title: "Could not trim empty leading frames from video.",
+				message: "Error: \(error.localizedDescription). This should not happen. Email sindresorhus@gmail.com and include this info:",
```
sindresorhus, line 196:
```suggestion
				message: "\(error.localizedDescription)\n\nThis should not happen. Email sindresorhus@gmail.com and include this info:",
```

Gifski/VideoValidator.swift
```diff
@@ -184,6 +184,20 @@ struct VideoValidator {
 			return .failure
 		}
 
-		return .success(newAsset, newVideoMetadata)
+		// Trim asset
+		do {
+			let trimmedAsset = try newAsset.trimmingBlankFrames()
+
```
sindresorhus, line 190:
```suggestion
```

Gifski/VideoValidator.swift
codetheweb: Example of how it looks now:
Screen Shot 2021-06-24 at 2 01 43 PM

Gifski/Utilities.swift
codetheweb: Fixed in 3a2c8a6.

Gifski/Utilities.swift
codetheweb: Fixed in 3a2c8a6.

Gifski/Utilities.swift
codetheweb: Added in 3a2c8a6.

codetheweb:
Thanks for all the detailed feedback; I really appreciate it.

Gifski/Utilities.swift
```diff
@@ -595,6 +595,103 @@ extension AVAsset {
...
+	*/
+	func trimmingBlankFrames() throws -> AVAsset {
+		guard let videoTrack = self.firstVideoTrack else {
+			throw VideoTrimmingError.assetIsMissingTrack
+		}
+
+		let trimmedTrack = try videoTrack.trimmingBlankFrames()
+
+		guard let trimmedAsset = trimmedTrack.asset else {
+			throw VideoTrimmingError.assetIsNil
```
sindresorhus, line 686: Can this actually happen? I don't think so. I think it would be better to use an `assertionFailure` and just fall back to an empty `AVMutableComposition()`.

sindresorhus:
https://github.com/sindresorhus/Gifski/pull/245/files#r657856913 was marked as resolved, but was not resolved.

Gifski/Utilities.swift
```diff
@@ -595,6 +595,103 @@ extension AVAsset {
...
+	Removes blank frames from the begining of the asset.
+
+	This can be useful to trim blank frames from files produced by tools like the iOS simulator screen recorder.
+	*/
+	func trimmingBlankFrames() throws -> AVAsset {
+		guard let videoTrack = self.firstVideoTrack else {
+			throw VideoTrimmingError.assetIsMissingTrack
+		}
+
+		let trimmedTrack = try videoTrack.trimmingBlankFrames()
```
sindresorhus, line 683: I just realized that this is kinda faulty. It works fine for Gifski, but it it drops any other video tracks and audio tracks. That is not expected from the method name and its docs.

Gifski/Utilities.swift
```diff
@@ -595,6 +595,102 @@ extension AVAsset {
...
+
+			// On first non-empty frame.
+			guard sampleBuffer.totalSampleSize == 0 else {
+				let currentTimestamp = sampleBuffer.outputPresentationTimeStamp
+				wrappedTrack.removeTimeRange(.init(start: .zero, end: currentTimestamp))
+				return wrappedTrack
+			}
+		}
+
+		throw VideoTrimmingError.assetIsEmpty
```
sindresorhus, line 647: This error is misnamed. It should be `videoTrackIsEmpty`.

Gifski/Utilities.swift
```diff
@@ -595,6 +595,102 @@ extension AVAsset {
 }
 
 
+extension AVAssetTrack {
+	enum VideoTrimmingError: Error {
+		case assetIsNil
+		case assetReaderFailure
+		case assetIsEmpty
+		case assetIsMissingTrack
```
sindresorhus, line 603:
```suggestion
		case assetIsMissingVideoTrack
```

sindresorhus (changes requested):

Gifski/Utilities.swift
codetheweb: It would be fairly straightforward to modify `AVAsset.trimmingBlankFrames()` so that it retained other tracks by creating a new composition; however video / audio might get out of sync then. Would you recommend just renaming it or modifying the logic?

sindresorhus: Just renaming it.

Gifski/Utilities.swift
```diff
@@ -595,6 +595,103 @@ extension AVAsset {
 }
 
 
+extension AVAssetTrack {
+	enum VideoTrimmingError: Error {
+		case assetIsNil
+		case assetReaderFailure
```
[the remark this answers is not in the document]
codetheweb: Fixed in 9a81320.

Gifski/Utilities.swift
codetheweb: Probably right, fixed in 9a81320.

Gifski/Utilities.swift
codetheweb: Updated in 9a81320.

Gifski/Utilities.swift
codetheweb: Fixed in 9a81320.

Gifski/Utilities.swift
codetheweb: Fixed in 9a81320.

sindresorhus:
Looks good now. Thanks :)

PR #248, opened 2021-07-14
Add support for changing the speed

### LINKED ISSUE #201: Change the speed of the GIF
A few people have requested this over email and some other GIF tools have this too.

---

**This issue requires you to have advanced Swift knowledge. It's not an easy beginner issue.**

### LINKED ISSUE #247: End frames dropped if they're the same
Hi,

I have a video that shows an animation that comes to rest in a final state and the video shows that final state for 2-3 seconds. However, when this video is converted to a .gif, the final state is only shown for a single frame. I'd like the gif to be as long as the video without it truncating duplicate frames at the end.

Or at least give me a checkbox to allow me to preserve all the frames if this behaviour is an optimisation in the app.

Here is the source and output showing what I mean.

https://user-images.githubusercontent.com/2559953/124855986-23744e00-dfed-11eb-8479-a275ef0da2e3.mp4
https://user-images.githubusercontent.com/2559953/124855971-1a837c80-dfed-11eb-8f56-ead0898ec449.gif

sindresorhus:
This was way more complicated than I had anticipated. Apparently, the API for playing videos only supports speeds of 0.5x to 2x (as usual, not documented) (https://github.com/feedback-assistant/reports/issues/229), so I had to employ a different tactic. Because of this, there's a small visual glitch when changing the speed because we're actually replacing the video with a copy. I have not found a way around that yet.

**Test build: [Gifski.app.zip](https://github.com/sindresorhus/Gifski/files/7003281/Gifski.app.zip)**

Fixes #201

---

This also fixes most of the problems with `generateCGImagesAsynchronously` (decode errors). Turns out the fix we had for https://github.com/sindresorhus/Gifski/blob/d8875c8aff216c34ac8380c0f627e4037c686575/Gifski/VideoValidator.swift#L174 was not actually used in the export as we read the video file in again when exporting: https://github.com/sindresorhus/Gifski/blob/d8875c8aff216c34ac8380c0f627e4037c686575/Gifski/Gifski.swift#L236 Duuh.

---

This also fixes #247 

https://github.com/sindresorhus/Gifski/pull/248/files#diff-c96c1241ad689fb7ba3363fb54bd27e6573310535d39eebdb597ce9838f3849dR340-R341

---

Screen Shot 2021-07-15 at 01 02 46

### DIFF
--- a/Gifski/EditVideoViewController.swift
+++ b/Gifski/EditVideoViewController.swift
@@ -3,6 +3,23 @@ import Combine
 import AVKit
 import Defaults
 
+struct SpeedView: View {
+	@Default(.outputSpeed) private var outputSpeed
+
+	var body: some View {
+		HStack(alignment: .firstTextBaseline) {
+			Text("Speed:")
+			Slider(value: $outputSpeed, in: 0.5...5, step: 0.5)
+			Text("\(outputSpeed.formatted)x")
+				.font(.system().monospacedDigit()) // TODO: Use `.monospacedDigit()` view modifier when targeting macOS 12.
+				.frame(width: 30, alignment: .leading)
+		}
+			.padding(.leading, 52)
+			.padding(.trailing, 20)
+			.padding(.bottom)
+	}
+}
+
 final class EditVideoViewController: NSViewController {
 	enum PredefinedSizeItem {
 		case custom
@@ -20,6 +37,7 @@ final class EditVideoViewController: NSViewController {
 	}
 
 	@IBOutlet private var estimatedSizeView: NSView!
+	@IBOutlet private var speedView: NSView!
 	@IBOutlet private var frameRateSlider: NSSlider!
 	@IBOutlet private var frameRateLabel: NSTextField!
 	@IBOutlet private var qualitySlider: NSSlider!
@@ -38,6 +56,7 @@ final class EditVideoViewController: NSViewController {
 
 	var inputUrl: URL!
 	var asset: AVAsset!
+	var modifiedAsset: AVAsset!
 	var videoMetadata: AVAsset.VideoMetadata!
 
 	private var resizableDimensions: ResizableDimensions!
@@ -95,6 +114,7 @@ final class EditVideoViewController: NSViewController {
 
 		self.inputUrl = inputUrl
 		self.asset = asset
+		self.modifiedAsset = asset
 		self.videoMetadata = videoMetadata
 
 		AppDelegate.shared.previousEditViewController = self
@@ -130,6 +150,7 @@ final class EditVideoViewController: NSViewController {
 		setUpDropView()
 		setUpTrimmingView()
 		setUpEstimatedFileSizeView()
+		setUpSpeedView()
 	}
 
 	override func viewDidAppear() {
@@ -139,6 +160,13 @@ final class EditVideoViewController: NSViewController {
 		setUpTabOrder()
 
 		tooltip.show(from: widthTextField, preferredEdge: .maxX)
+	}
+
+	private func setUpSpeedView() {
+		let view = SpeedView()
+		let hostingView = NSHostingView(rootView: view)
+		speedView.addSubview(hostingView)
+		hostingView.constrainEdgesToSuperview()
 	}
 
 	private func setUpEstimatedFileSizeView() {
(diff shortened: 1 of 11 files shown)


sindresorhus:
Please test it out and let me know.

What should the multiplier steps be? Currently, I just made it: `0.5`, `1`, `1.5`, `2`, `2.5`, `3`, `3.5`, `4`, `4.5`, `5`. They don't even have to be linear.

sindresorhus:
// @fawkeswei @janbrasna @codeclown @alexkaessner @loctran016

alexkaessner:
I’ve tested it (on Monterey) and it seems to work fine – awesome! 👌 

I personally think the factor makes more sense than percentages. It’s easier to wrap the head around. The steps should be good for some fast changes as well.

Gifski/EditVideoViewController.swift
```diff
@@ -3,6 +3,23 @@ import Combine
...
 import Defaults
 
+struct SpeedView: View {
+	@Default(.outputSpeed) private var outputSpeed
+
+	var body: some View {
+		HStack(alignment: .firstTextBaseline) {
+			Text("Speed:")
+			Slider(value: $outputSpeed, in: 0.5...5, step: 0.5)
+			Text("\(outputSpeed.formatted)x")
```
janbrasna, line 13:
```suggestion
			Text("\(outputSpeed.formatted)×")
```

janbrasna (commented):
Wow! Thanks for putting so much effort in the feature. 

I'll try out how bad the flicker is once I have access to a 10.15+ machine next week, however if it's there only when the user changes the speed, chances are their focus is on the actual slider, so no harm done if something re-renders while dragging around.

Factor and range choices both sound very reasonable.

For now only the usual Unicode/legibility for multiplication:

kornelski:
It's not critical, but ideally a change in speed should affect the FPS setting. If you start with a 20fps movie and run it at 0.5x, then max fps you can get is 10. Conversely, if you start with a 10fps movie and run it at 2x speed, then you can have a 20fps anim.

sindresorhus:
New build: https://dsc.cloud/sindresorhus/Gifski.app-1626442968 (a fix to trimming)

sindresorhus:
@kornelski I agree. I didn't think of that. I'll do that soon.

janbrasna:
@kornelski Would you like to alter the FPS proportionally always… or just when the speed change would hit the FPS limits of the asset, or the ability for smooth frame distribution?

kornelski:
I think always, with clamping.

```
actual_fps = min(50, fps_multiplier * speed_multiplier)
```

Display `actual_fps` to users, but make the slider change `fps_multiplier`.

janbrasna:
Love the feature! Did some slowing down and speeding up already, this is so handy!

@sindresorhus Now I see what you mean by the flicker:

Screen Recording 2021-07-27 at 23 33 28

Not sure what can be done about it. Maybe some masking technique to cover for the "realod" of the recreated asset? (Think Screen Sharing's dimming+spinner over the viewport when reconnecting… something alike, to suppress the perception of something changing after dropping the slider) — Or a duplicate of the component and hiding the old one with a delay once the new one is already loaded? I'm at a loss what trickery could be used here… :{

janbrasna:
PS: I've noticed some slight misalignments:
Screen Shot 2021-07-28 at 00 01 48
Screen Shot 2021-07-28 at 00 01 56

sindresorhus:
@kornelski Fixed in https://github.com/sindresorhus/Gifski/pull/248/commits/1e33e62b8c263ab4b46fcaa493c44c37a99d6adc

sindresorhus:
@janbrasna Alignment stuff fixed in https://github.com/sindresorhus/Gifski/pull/248/commits/64a66337483638638c7043d72655bf9f3a6816db

sindresorhus:
New build: [Gifski.app.zip](https://github.com/sindresorhus/Gifski/files/6994926/Gifski.app.zip) *(Please help test)*

kornelski:
Speed and fps have inverse relationship. If I set speed to 4x I get 5fps. If I set speed to 0.5x I get 20fps. It should be the opposite.

sindresorhus:
Ugh. I was thinking about it incorrectly. Fixed now.

[Gifski.app.zip](https://github.com/sindresorhus/Gifski/files/7003281/Gifski.app.zip)

PR #255, opened 2021-09-26
Add an "open" button to the conversion completed screen

sindresorhus:
To make it faster to convert another file, for people not using drag & drop. Clicking it, will present the system "open" panel.

I'm open to suggestions for a more fitting symbol (SF Symbols) than a `+`.

Screen Shot 2021-09-26 at 16 44 11


sindresorhus:
Some more alternatives. I still think the plus is the best one yet.
Screen Shot 2021-10-03 at 14 01 16
Screen Shot 2021-10-03 at 14 03 10
Screen Shot 2021-10-03 at 14 12 03

PR #256, opened 2021-10-03
Show the final GIF duration in the user-interface

### LINKED ISSUE #252: Duration of time
Hi @sindresorhus , in addition to knowing the estimated size, it would also be possible to know how long the video lasts after editing.

https://user-images.githubusercontent.com/20476002/133589612-3617b60d-c784-409d-9d9e-a068e836252a.mov

sindresorhus:
Fixes #252

Build: [Gifski.app.zip](https://github.com/sindresorhus/Gifski/files/7273878/Gifski.app.zip)
 
---

Screen Shot 2021-10-03 at 22 33 38


sindresorhus:
@Angelk90 Would you be able to try it out?

Angelk90:
@sindresorhus : Tried, it seems to work.
I only noticed that when a bar is already shifted, then consider that I took a few seconds away.
Each time I move the bars, the duration time takes the initial value of the video, until it is recalculated.
But no problem.

sindresorhus:
> Each time I move the bars, the duration time takes the initial value of the video, until it is recalculated.

Yeah, I forgot to mention that. I didn't find a good fix for that yesterday, but I managed to find a workaround for it today with a clearer head.

Issue #254, opened 2021-09-22
The trim leading video frames feature fails sometimes


sindresorhus:
@codetheweb I have noticed that there are a bunch of Crashlytics (automated) reports about this feature not working: "Could not trim empty leading frames from video.". Any ideas?

---

> AVAsset debug info ## Extension: mov Video codec: BGRA Audio codec: nil Duration: 25s Dimension: 376×805 Frame rate: 9.79 File size: 305.4 MB Is readable: true Is playable: true Is exportable: true Has protected content: false Track #1 ---- Type: Video Codec: nil Duration: 25s Dimensions: 376×805 Natural size: (376.0, 805.0) Frame rate: 9.79 Is playable: true Is decodable: true ----

> AVAsset debug info ## Extension: nil Video codec: v210 Audio codec: nil Duration: 3s Dimension: 900×700 Frame rate: 29.97 File size: 160 MB Is readable: true Is playable: true Is exportable: true Has protected content: false Track #1 ---- Type: Video Codec: nil Duration: 3s Dimensions: 900×700 Natural size: (900.0, 700.0) Frame rate: 29.97 Is playable: true Is decodable: true ----

> AVAsset debug info ## Extension: nil Video codec: v210 Audio codec: nil Duration: 5s Dimension: 900×700 Frame rate: 29.97 File size: 275.8 MB Is readable: true Is playable: true Is exportable: true Has protected content: false Track #1 ---- Type: Video Codec: nil Duration: 5s Dimensions: 900×700 Natural size: (900.0, 700.0) Frame rate: 29.97 Is playable: true Is decodable: true ----

> AVAsset debug info ## Extension: nil Video codec: v210 Audio codec: nil Duration: 2s Dimension: 900×700 Frame rate: 29.97 File size: 109 MB Is readable: true Is playable: true Is exportable: true Has protected content: false Track #1 ---- Type: Video Codec: nil Duration: 2s Dimensions: 900×700 Natural size: (900.0, 700.0) Frame rate: 29.97 Is playable: true Is decodable: true ----

codetheweb:
I was able to reproduce this by creating a v210 encoded .mov file:

https://user-images.githubusercontent.com/7410405/134365364-ab83c10e-b7ca-4510-8d92-a78209dba128.mov

I'll try and take a look later today.

codetheweb:
[...]

PR #257, opened 2021-10-03
Don't crash if trimming fails for some codecs

codetheweb:
Closes https://github.com/sindresorhus/Gifski/issues/254.

### DIFF
--- a/Gifski/Utilities.swift
+++ b/Gifski/Utilities.swift
@@ -604,6 +604,7 @@ extension AVAssetTrack {
 		case videoTrackIsEmpty
 		case assetIsMissingVideoTrack
 		case compositionCouldNotBeCreated
+		case codecNotSupported
 	}
 
 	/**
@@ -612,6 +613,16 @@ extension AVAssetTrack {
 	This can be useful to trim blank frames from files produced by tools like the iOS simulator screen recorder.
 	*/
 	func trimmingBlankFrames() throws -> AVAssetTrack {
+		// See https://github.com/sindresorhus/Gifski/issues/254 for context.
+		// TL;DR is that some codecs seem to always report a buffer size of 0
+		// when reading, breaking this function.
+		let buggyCodecs = ["v210", "BGRA"]
+		if let codecIdentifier = self.codecIdentifier {
+			if buggyCodecs.contains(codecIdentifier) {
+				throw VideoTrimmingError.codecNotSupported
+			}
+		}
+
 		// Create new composition
 		let composition = AVMutableComposition()
 		guard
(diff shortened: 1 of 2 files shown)


Gifski/VideoValidator.swift
```diff
@@ -188,6 +188,15 @@ struct VideoValidator {
 		do {
 			let trimmedAsset = try newAsset.trimmingBlankFramesFromFirstVideoTrack()
 			return .success(trimmedAsset, newVideoMetadata)
+		} catch AVAssetTrack.VideoTrimmingError.codecNotSupported {
+			NSAlert.showModal(
+				for: window,
+				title: "Could not trim empty leading frames from video.",
+				message: "We can't trim empty leading frames from this codec, sorry. You can still create a GIF."
+			)
```
sindresorhus, line 196: Just thinking out loud, do we even need to show this modal? The user cannot do anything about it anyway and it's likely they won't notice a difference, and they will also most likely not understand what it means. Maybe we should just skip the modal?

Gifski/Utilities.swift
```diff
@@ -612,6 +613,16 @@ extension AVAssetTrack {
 	This can be useful to trim blank frames from files produced by tools like the iOS simulator screen recorder.
 	*/
 	func trimmingBlankFrames() throws -> AVAssetTrack {
+		// See https://github.com/sindresorhus/Gifski/issues/254 for context.
+		// TL;DR is that some codecs seem to always report a buffer size of 0
+		// when reading, breaking this function.
+		let buggyCodecs = ["v210", "BGRA"]
+		if let codecIdentifier = self.codecIdentifier {
+			if buggyCodecs.contains(codecIdentifier) {
```
sindresorhus, line 621:
```suggestion
		if
			let codecIdentifier = codecIdentifier,
			buggyCodecs.contains(codecIdentifier)
		{
```

Gifski/Utilities.swift
```diff
@@ -612,6 +613,16 @@ extension AVAssetTrack {
 	This can be useful to trim blank frames from files produced by tools like the iOS simulator screen recorder.
 	*/
 	func trimmingBlankFrames() throws -> AVAssetTrack {
+		// See https://github.com/sindresorhus/Gifski/issues/254 for context.
+		// TL;DR is that some codecs seem to always report a buffer size of 0
+		// when reading, breaking this function.
```
sindresorhus, line 618:
```suggestion
		// In short: Some codecs seem to always report a sample buffer size of 0 when reading, breaking this function. (macOS 11.6)
```

sindresorhus (changes requested):

Gifski/VideoValidator.swift
codetheweb: Sounds good to me. 👍 

Removed in 0f15040.

sindresorhus:
Thanks for fixing :)

PR #262, opened 2022-01-10
Skip incorrect out-of-range frames

kornelski:
When using trimming the last frame has incorrect timestamp.

### DIFF
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -396,6 +396,9 @@ final class Gifski {
 				)
 			}
 			
+			// TODO: This is just a workaround. Look into the cause of this.
+			// https://github.com/sindresorhus/Gifski/pull/262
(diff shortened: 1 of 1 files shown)


sindresorhus:
> When using trimming the last frame has incorrect timestamp.

I may be misunderstanding this, but your fix ensures that a frame is not before the first time code. Did you mean to say "first frame", not "last"?

How do I reproduce this? Do I trim from the start or end?

kornelski:
I can reproduce it every time by using the trim feature - just trim any fragment of a video. The *last* frame that is being decoded and fed to gifski has timestamp that is before the first frame. There's some wraparound or buggy timestamping happening.
Try changing this to an `assert` and see when it fails.

sindresorhus:
> There's some wraparound or buggy timestamping happening.

Very strange. I'll look into the core issue more when I have some time.

PR #278, opened 2022-08-07
Fix green line sometimes occurring on the GIF

sindresorhus:
To reproduce, take the [this video](https://user-images.githubusercontent.com/170270/183291184-b53aa3b0-b4b4-4e58-89dd-97755e35dc4f.mp4) and set the size to 112 in the existing Gifski version. It will produce a GIF that is has the dimensions 111x112 (incorrect dimensions) and it also has a green line on the right side:  
Heart Spin 1080

I have confirmed that the green line happens before passing it to the gifski library.

However, if we disable `generator.maximumSize` as done in this PR, the GIF comes out with the correct dimensions and no artifacts:
Heart Spin 1080

The GIF also comes out correct when the dimensions are 108 instead of 112, so it makes me think it has something to do with it not handling sizes well that are not multiples on the original size.

### DIFF
--- a/Gifski/Gifski.swift
+++ b/Gifski/Gifski.swift
@@ -295,10 +295,12 @@ final class Gifski {
 		generator.requestedTimeToleranceBefore = .zero
 		generator.requestedTimeToleranceAfter = .zero
 
+		// Disabled as it can create a green line on the side if the size is not a multiple of the original size. For example, when the orignal is 1080 and you want the size 112. However 108 does not create a green line. (macOS 12.5)
+		// https://github.com/sindresorhus/Gifski/pull/278
 		// This improves the performance a little bit.
-		if let dimensions = conversion.dimensions {
-			generator.maximumSize = CGSize(widthHeight: dimensions.longestSide)
-		}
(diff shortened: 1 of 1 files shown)


sindresorhus:
@kornelski Do you have any idea about the cause of this? If it's really the multiple thing, we could maybe only set the maximum size when it's a multiple.

kornelski:
You get green when all components of YCbCr color are zero, so it's an uninitialized area or a read out of bounds. 

Codecs typically internally can only have a size that is a multiple of 4/8/16 (with potentially buggy tricks used to hide this) and chroma subsampling also often adds a requirement that dimensions must be even. 

So these types of bugs are easy to make in a video decoder. Entirely unsurprising that it happens when scaling.

kornelski:
So I suggest decoding at video's native resolution and raising frames later in RGB(A).

If you tell the codec to resize, only require smaller sizes and prefer integer scaling factors (divide by 2 or 3)

sindresorhus:
Thanks for explaining :)

sindresorhus:
> So I suggest decoding at video's native resolution and resizing frames later in RGB(A).

👍

PR #280, opened 2022-08-12
add brew install instructions

amilajack:


amilajack:
points to the wrong gifski, closing

PR #283, opened 2022-12-01
Add m1-friendly swiftlint script to build phase.

### LINKED ISSUE #284: Missing default support for swiftlint installation on M1 mac
The homebrew install location for swiftlint differs on M1 mac vs x86 arch. This project should check the m1 location, and then fall back gracefully in the build phase script.

PR Pending.

ben-p-commits:
Add m1-friendly swiftlint script to build phase. Homebrew on mac silicon has a different installation path (opt/homebreow), so this catches both m1 and x86 scenarios.

This script came from [an issue](https://github.com/realm/SwiftLint/issues/2992#issuecomment-771001634) on the swiftlint repository.

image

Fixes #284

### DIFF
--- a/Gifski.xcodeproj/project.pbxproj
+++ b/Gifski.xcodeproj/project.pbxproj
@@ -438,8 +438,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "swiftlint\n";
-			showEnvVarsInLog = 0;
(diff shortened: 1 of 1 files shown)


sindresorhus:
I think you can just add to the path without testing it first. It makes no difference. And we don't really need the handholding.

This should be enough

```
PATH="/opt/homebrew/bin/:${PATH}"
swiftlint
```

ben-p-commits:
@sindresorhus , updated per your comment.

DaniSpringer:
I spent quite a while trying to fix this error. Was there some way I should have known about this? Otherwise perhaps we can improve the docs.

PR #286, opened 2022-12-10
Update gifski to 1.8.1

kornelski:
There are 3 new C functions that you could hook up to the UI if you like:

* gifski_set_motion_quality — controls denoising/motion smoothing. Lower quality reduces motion in the video, may add smears, but reduce file sizes.
* gifski_set_lossy_quality — controls lossy gif encoder. Lower quality adds grain/noise, but reduces file sizes.
* gifski_set_extra_effort — makes encoding slower, but might increase quality slightly.

### DIFF
--- a/gifski-api/Cargo.lock
+++ b/gifski-api/Cargo.lock
@@ -70,15 +70,15 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bytemuck"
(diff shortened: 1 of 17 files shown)


sindresorhus:
Thanks 👍

I don't plan a new release for a while. The current release haven't had any reported issues, so I don't want to jinx it.

As for the new settings. I don't want to do any changes to do the current editor as it's really cumbersome. I'll consider whether it makes sense to add those settings when I eventually do the SwiftUI rewrite.

PR #292, opened 2023-01-25
Update to gif.ski 1.10.0

kornelski:
Performance improvements. Smaller files at lowest qualities.

PR #293, opened 2023-02-02
Update gif.ski 1.10.1

kornelski:
#292


sindresorhus:
Still crashing even with this. I may be something Gifski is doing wrong. I'm honestly not sure. See the `Crashed: c-write` thread: https://gist.github.com/sindresorhus/812c18015684809c42917d64b7da320e

kornelski:
Hmm, it's crashing here: https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/thread.rs#L265

which is a very low-level thing. If pthread_join failed, something must have gone really wrong at the OS level?

Can you capture stderr of the process? Panics print an error message to stderr.

PR #295, opened 2023-02-16
Update gif.ski to 1.10.2

kornelski:
I've added `gifski_set_error_message_callback` that works similarly to `gifski_set_write_callback`, and gives a NUL-terminated string with fatal errors. These are never printed on success, so you can log them and report them the same way as crashes.


kornelski:
It also catches all panics in `gifski_finish` and turns them into `GifskiError::THREAD_LOST` error code. If you get that error, it's a bug or some other dead-end situation, perhaps out of memory error.

PR #321, opened 2024-02-09
Update gif.ski to 1.14.4

kornelski:
* Improved and bugfixed trimming of frames with low motion.
* Optimized GIF encoder and a few dependencies. The new binary is even a bit smaller.


sindresorhus:
Thanks 👍

Just FYI. I don't plan to release this anytime soon. The current Gifski version on the App Store is stable and has had no reported issues for a long time, so I prefer to keep it that way for now. And I'm working on a rewrite of the app. I will put up a TestFlight for it in a while when I have ironed out some issues.

kornelski:
@sindresorhus in the Gifski rewrite, are you only changing internals, or redesigning the UI too?

sindresorhus:
Changing internals was the main priority, but I have modernized the UI a little bit too. You can try it out here if you want: [Gifski - new.zip](https://github.com/sindresorhus/Gifski/files/14791787/Gifski.-.new.zip)


It's mostly working, except the width/height text fields are a bit buggy. I still need to do some polish on the UI (more interesting landing screen, etc). I will do a public TestFlight when I manage to find time to fix the text fields.

PR #323, opened 2024-06-26
[Suggestion] Support translations

LenhartStephan:
I'v added a string catalog (Localizable.xcstrings) and a german translation for the auto generated strings. This is not covering all user facing strings (e.g. the dialog for "Animated GIF Preview Limitation" is currently not included) but should be a good start for l10n.
Let me know what you think.


sindresorhus:
Thanks for the offer, but as [documented](https://github.com/sindresorhus/Gifski?tab=readme-ov-file#can-i-contribute-localizations), we don't plan to localize the app.

Issue #136, opened 2019-09-19
GIF Preview feature


arggh:
I think previewing the gif in the same screen where I tweak the conversion settings is a superior user experience compared to going back and forth between the settings-view and the save-view. Especially, since there is already a video player present in the settings-view.

For example, there could be a button <kbd>Render preview</kbd> or a toggle: <kbd>Original</kbd> - <kbd>Preview</kbd>

If you select <kbd>Preview</kbd>, the original video would be replaced with the gif and a progress indicator, that would let me know whenever gifski is busy converting a new preview. Tweaking the settings would start the preview conversion from scratch.

This would make it effortless to tweak the quality and framerate settings.

Another idea to speed up the preview process: how about rendering just the first frame of the gif with the settings applied? That should be plenty fast, and you would get some perspective into what the final result will look like, given the resolution and quality. Maybe even let the user scrub-select frames and render them as needed?

sindresorhus:
I agree, it's a bit cumbersome at the moment if you need to tweak settings.

> or a toggle: Original - Preview

This sounds like a great idea indeed.

> Another idea to speed up the preview process: how about rendering just the first frame of the gif with the settings applied?

Or maybe 5 frames evenly spread out over the video time range. That way you can quickly get a feel of the image quality of the whole video.

> Maybe even let the user scrub-select frames and render them as needed?

I think the optimal rendering steps would be:

1. Render first frame.
2. Render the next 4 frames, evenly spread out.
3. Start rendering the whole thing for real.
4. Also allow scrubbing to render one frame at the scrubbed position.

If anyone wants to work on this, even just step `1` would be a good improvement.

kornelski:
Because gifski tries to encode differences between frames, non-consecutive frames will look a bit different than in the full anim.

How about starting encoding from the start and allowing viewing everything that has been encoded so far?

hazeycode:
Is anyone already working on this? I can take a shot at it if not

mmulet:
Solved by pull request #329

Issue #104, opened 2019-08-04
Animate the initial view


sindresorhus:
See https://github.com/sindresorhus/Gifski/pull/88

This issue is blocked us to being able to adopt SwiftUI, which will make such animations much simpler to implement.

sindresorhus:
This can now be done. The SwiftUI stuff can just be behind an availability check.

Issue #11, opened 2018-02-26
Master crashes


sindresorhus:
@kornelski When I "archive" master and run the app, it crashes if I don't have `gcc` installed (try with `brew uninstall gcc`).

```
Process:               Gifski [11129]
Path:                  /Applications/Gifski.app/Contents/MacOS/Gifski
Identifier:            com.sindresorhus.Gifski
Version:               1.1.0 (2)
App Item ID:           0
App External ID:       0
Code Type:             X86-64 (Native)
Parent Process:        ??? [1]
Responsible:           Gifski [11129]
User ID:               502

Date/Time:             2018-02-25 13:12:05.364 -0800
OS Version:            Mac OS X 10.13.2 (17C89)
Report Version:        12
Anonymous UUID:        23501669-5784-1A67-4D38-0583AC23BEC6

Sleep/Wake UUID:       7463A5D6-0414-4382-9426-082B4CE27497

Time Awake Since Boot: 630000 seconds
Time Since Wake:       280 seconds

System Integrity Protection: enabled

Crashed Thread:        0

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Reason:    DYLD, [0x1] Library missing

Application Specific Information:
dyld: launch, loading dependent libraries

Dyld Error Message:
  Library not loaded: /usr/local/opt/gcc/lib/gcc/7/libgomp.1.dylib
  Referenced from: /Applications/Gifski.app/Contents/MacOS/Gifski
  Reason: image not found

Binary Images:
       0x104d7d000 -        0x104ec3ff7 +com.sindresorhus.Gifski (1.1.0 - 2) <F81B6CB9-1DAC-37D0-9569-408389989AD8> /Applications/Gifski.app/Contents/MacOS/Gifski
       0x104f0e000 -        0x104f2eff7 +ProgressKit (1.0 - ???) <5845D54C-644D-309D-857B-F5F266CCE798> /Applications/Gifski.app/Contents/Frameworks/ProgressKit.framework/Versions/A/ProgressKit
       0x10decf000 -        0x10df1998f  dyld (519.2.2) <6695F30B-4E88-3C0B-9867-7D738C44A3E6> /usr/lib/dyld
[...]

Issue #10, opened 2018-02-25
Layout issues in the save panel accessory view


LarsJK:
Noticed some layout issues when running master:

<img width="714" alt="image" src="https://user-images.githubusercontent.com/1528255/36647439-0490cc8c-1a86-11e8-80bb-ce66c044c6cb.png">

PR #329, opened 2025-03-22
implemented GIF preview feature issue #136

mmulet:
Hi! 
I implemented the gif preview features as requested in issue #136
Please watch this short video for a demonstration:

https://github.com/user-attachments/assets/0b02212b-29f7-43d9-91f3-ec74a0ab9e05

Here is what the feature does:

- [x] adds the Show Preview Texture toggle
- [x] Renders the current frame right away
- [x] Starts rendering the whole thing for real.
- [x] Once it is all rendered you can preview the render with the play button
- [x] Allows for scrubbing to render one frame at the scrubbed position.
- [x] Implemented in SwiftUI and Combine
Looking at the issue #136 this should fulfill all of your major requirements. 

I will submit this to issueHunt for the ongoing bounty, but any extra tips (you

### DIFF
--- a/Gifski/PreviewGenerator.swift
+++ b/Gifski/PreviewGenerator.swift
@@ -1,0 +1,379 @@
+//
+//  PreviewGenerator.swift
+//  Gifski
+//
+//  Created by Michael Mulet on 3/22/25.
+
+import Foundation
+import SwiftUI
+import AVFoundation
+/**
+	This class handles the generation of all previews.
+    It has two types, singleFrame and entireAnimation
+    but both are very similar. and work the same way.
+    Here is how:
+
+	When settings update, or video time change we add a new
+	preview to a queue which runs on a background thread.
+
+	The queue has a depth of 1, so only the most recent
+	preview will be generated. If a new preview is added
+	before the current preview is generated, the current
+	preview will be overwritten by the new preview.
+
+	Once it is done generating the preview, it updates
+	the previewImage property which is observed @Published
+
+	Please note that when you generate a single frame
+	preview, it will generate a preview of the closet
+	frame in the animation, which may not be at the
+	exact time you scrubbed to. This is because the
+	frame rate of the animation is usually very low
+	and the fame you scrubbed may not exist in the
+	output GIF.
+*/
+final class PreviewGenerator: ObservableObject {
+	@MainActor
+	@Published var previewImage: PreviewImage?
+	@MainActor
+	@Published var imageBeingGeneratedNow = false
+
+	enum GeneratorType {
+		case singleFrame
+		case entireAnimation
+	}
+
+	var type: GeneratorType
+
+	init(type: GeneratorType) {
+		self.type = type
+	}
+
+
+
+	enum PreviewGeneration {
+		case oneFrame(atTime: Double)
+		case entireAnimation
+	}
+	struct PreviewImage  {
+		var image: NSImage
+		var whatWasGenerated: PreviewGeneration
+		var settingsAtGenerateTime: GIFGenerator.Conversion
+	}
+
+	private enum ImageDataType {
+		case stillImage
+		case animatedGIF
+	}
+
+	/// NSImages are not Sendable so we will
+	/// just send the Data to the main thread
+	/// and create the image there
+	private struct PreviewImageData: Sendable {
+		var imageData: Data
+		var type: ImageDataType
+		var whatWasGenerated: PreviewGeneration
+		var settingsAtGenerateTime: GIFGenerator.Conversion
+
+		func toPreviewImage() -> PreviewImage?{
+			let image: NSImage
+			switch type {
+			case .animatedGIF:
+				image = NSImage(data: imageData) ?? NSImage()
+			case .stillImage:
+				guard let imageSource = CGImageSourceCreateWithData(imageData as CFData, nil),
+						  let cgImage = CGImageSourceCreateImageAtIndex(imageSource, 0, nil) else {
+						return nil
+				}
+				image = NSImage(cgImage: cgImage, size: .zero)
+			}
+
+			return PreviewImage(
+				image: image,
+				whatWasGenerated: self.whatWasGenerated,
+				settingsAtGenerateTime: self.settingsAtGenerateTime
+			)
+		}
+	}
+
+
+	private enum CreatePreviewState {
+		case idle
+		/// If the state is running we may queue the next
+		/// time to preview, which will render after
+		/// we are done. Increase the serial number
+		/// to indicate that a new preview must be drawn
+		case running(whatToGenerate: PreviewGeneration,
+					 settingsAtGenerateTime: GIFGenerator.Conversion,
+					 previewSerialNumber: Int
+		)
+	}
+
+	private var createPreviewState: CreatePreviewState = .idle {
+		didSet {
+			let isRunning: Bool
+			switch createPreviewState {
+			case .idle:
+				isRunning = false
+			case .running:
+				isRunning = true
+			}
+			Task {@MainActor in
+				self.imageBeingGeneratedNow = isRunning
+			}
+		}
+	}
+
+	private enum PreviewResult{
+		case new_image(image: PreviewImageData)
+		case no_change
+	}
+
+	/// This function spawns the createPreviewThread if none exists.
+	/// If one does exist, it will add the most recent time to a queue to
+	/// be generated later. It will overwrite the most recent queued item
+	/// so only the most recent time will generate a preview.
+	@MainActor
+	func onCurrentTimeDidChange(currentTime: Double, conversionSettings: GIFGenerator.Conversion)  {
+		switch type {
+		case .entireAnimation:
+			/// Only create new frames
+			/// on scrub if you are a
+			/// single frames type
+			return
+		case .singleFrame:
+			break
+		}
+
+		let previewSerialNumber: Int
+		let wasIdleBeforeThisFunctionCall: Bool
+		switch createPreviewState {
+		case .idle:
+			wasIdleBeforeThisFunctionCall = true
+			previewSerialNumber = 0
+
+		case .running( _, _, let oldPreviewSerialNumber):
+			wasIdleBeforeThisFunctionCall = false
+			previewSerialNumber = oldPreviewSerialNumber + 1
+		}
+
+		createPreviewState = .running(
+			whatToGenerate: .oneFrame(atTime: currentTime),
+			settingsAtGenerateTime: conversionSettings,
+			previewSerialNumber: previewSerialNumber
+		)
+		if wasIdleBeforeThisFunctionCall {
+			Task(priority: .utility) {
+				await self.createPreviewBackgroundThreadLoop()
+			}
+		}
+	}
+
+	@MainActor
+	func onSettingsDidChangeGeneratePreview(conversionSettings: GIFGenerator.Conversion){
+		let previewSerialNumber: Int
+		let wasIdleBeforeThisFunctionCall: Bool
+		switch createPreviewState {
+		case .idle:
+			wasIdleBeforeThisFunctionCall = true
+			previewSerialNumber = 0
+
+		case .running( _, _, let oldPreviewSerialNumber):
+			wasIdleBeforeThisFunctionCall = false
+			previewSerialNumber = oldPreviewSerialNumber + 1
+		}
+
+		let whatToGenerate: PreviewGeneration
+		switch type {
+		case .entireAnimation:
+			whatToGenerate = .entireAnimation
+		case .singleFrame:
+			/// We don't have a preview, so don't
+			/// bother updating nothing
+			guard let previewImage else {
+				return
+			}
+			whatToGenerate = previewImage.whatWasGenerated
+		}
+		createPreviewState = .running(
+			whatToGenerate: whatToGenerate,
+			settingsAtGenerateTime: conversionSettings,
+			previewSerialNumber: previewSerialNumber
+		)
+
+		if wasIdleBeforeThisFunctionCall {
+			Task(priority: .utility) {
+				await self.createPreviewBackgroundThreadLoop()
+			}
+		}
+	}
+
+
+	/// This will use GiFGenerate To generate a new preview image at a particular
+	///  time or of the entire animation
+	private func generatePreviewImage(
+		whatToGenerate: PreviewGeneration,
+		settingsAtGenerateTime: GIFGenerator.Conversion,
+		oldPreviewImage: PreviewImage?
+	) async -> PreviewResult? {
+		switch whatToGenerate {
+		case .entireAnimation:
+			if case .entireAnimation = oldPreviewImage?.whatWasGenerated,
+			   oldPreviewImage?.settingsAtGenerateTime == settingsAtGenerateTime
+			{
+				return .no_change
+			}
+			let data = try? await GIFGenerator.run(settingsAtGenerateTime) { _ in
+				///no-op
+			}
+
+			guard let data
+				   else {
+				return nil
+			}
+			return .new_image(image: .init(
+				imageData: data,
+				type: .animatedGIF,
+				whatWasGenerated: whatToGenerate,
+				settingsAtGenerateTime: settingsAtGenerateTime
+			))
+		case .oneFrame(let previewTimeToGenerate):
+			guard let frameRange = settingsAtGenerateTime.timeRange else {
+				/// Don't have a frame range at all
+				///  don't produce a preview GIF
+				return nil
+			}
+			guard let frameRate = settingsAtGenerateTime.frameRate else {
+				return nil
+			}
+			/// We want to generate 1 frame
+			/// we have currentFrameSettings.frameRate frames/second
+			/// or 1/currentFrameSettings.frameRate seconds/frame
+			/// then multiply by 1 frame to the duration of one frame
+			let duration_of_one_frame: Double = 1 / (frameRate.toDouble)
+			/// Line up the current time to a frame that will
+			/// be generated the generator
+			let frame_number = ((previewTimeToGenerate - frameRange.lowerBound) / duration_of_one_frame).rounded(.down)
+
+			let start = frame_number * duration_of_one_frame + frameRange.lowerBound
+
+			if let oldPreviewImage,
+			   oldPreviewImage.settingsAtGenerateTime == settingsAtGenerateTime,
+			   case let .oneFrame(oldPreviewImageTimestamp) = oldPreviewImage.whatWasGenerated,
+			   oldPreviewImageTimestamp == start
+			{
+				return .no_change
+			}
+			var currentFrameSettings = settingsAtGenerateTime
+			/// Set the frame Rate artificially high
+			/// because the GifGenerator may fail
+			/// to run if near the end and less than
+			/// one frame is pushed through
+			currentFrameSettings.frameRate = 18
+			/// Generate an average of 2.5 frames
+			/// GIFGenerator fails if generating only 1
+			/// frame. So le'ts make sure at least 2 frames
+			/// generate
+			currentFrameSettings.timeRange = start...(start + 2.5 / 18.0)
+
+			let data = try? await GIFGenerator.run(currentFrameSettings) { _ in
+				///no-op
+			}
+
+			guard let data else {
+				return nil
+			}
+
+
+			return .new_image(
+				image: .init(
+					imageData: data,
+					type: .stillImage,
+					whatWasGenerated: .oneFrame(atTime: start),
+					settingsAtGenerateTime: settingsAtGenerateTime
+				)
+			)
+		}
+	}
+
+	/// This function runs in a loop to create previews.
+	///  It is intended to run this on a background thread
+	///  using Task(priority: .utility)
+	private func createPreviewBackgroundThreadLoop() async {
+		///
+		/// Loop Explanation:
+		/// Generate the top frame in the queue.
+		/// Then check for additional frames, if there are any
+		/// generate the next frame in the queue, in an infinite
+		/// loop.
+		while true {
+			let createPreviewStateAtStartOfLoop: CreatePreviewState = self.createPreviewState
+			switch createPreviewStateAtStartOfLoop {
+			case .idle:
+				/// No more work. End the loop.
+				return
+			case .running(
+				let whatToGenerate,
+				let settingsAtGenerateTime,
+				let serialNumberOfPreview
+			):
+				let data = await generatePreviewImage(
+					whatToGenerate: whatToGenerate,
+					settingsAtGenerateTime: settingsAtGenerateTime,
+					oldPreviewImage: self.previewImage
+				)
+				///This may be different than the start of the loop as we may
+				///have received requests during the time it took to generate
+				let createPreviewStateAfterGeneratingPreview: CreatePreviewState = self.createPreviewState
+
+				switch createPreviewStateAfterGeneratingPreview {
+				case .idle:
+					/// This case should never happen
+					/// because the only time we set the createPreviewState
+					/// to .idle is in the case below:
+					assertionFailure()
+					switch data {
+					case .no_change:
+						return
+					case nil:
+						Task {
+							@MainActor in
+							self.previewImage = nil
+						}
+					case .new_image(let newImage):
+						Task {
+							@MainActor in
+							previewImage = newImage.toPreviewImage()
+						}
+					}
+				case .running(_, _, let serialNumberAfterPreview):
+					let previewIsFinished = serialNumberOfPreview == serialNumberAfterPreview
+						switch data {
+						case .no_change:
+							break
+						case nil:
+							Task {
+								@MainActor in
+								self.previewImage = nil
+							}
+						case .new_image(let newImage):
+							Task {
+								@MainActor in
+								previewImage = newImage.toPreviewImage()
+							}
+						}
+
+					guard previewIsFinished else {
+						///Preview is not finished
+						///resume the loop and process
+						///the next frame
+						continue
+					}
+					/// Preview is finished. End the createPreviewFunction
+					self.createPreviewState = .idle
+					return
+				}
+			}
+		}
+	}
+}
--- a/Gifski/PreviewView.swift
+++ b/Gifski/PreviewView.swift
@@ -1,0 +1,30 @@
+//
+//  PreviewView.swift
+//  Gifski
+//
+//  Created by Michael Mulet on 3/21/25.
+//
+
+import SwiftUI
+
+@MainActor
+final class PreviewViewState: ObservableObject {
+	@Published var previewImage: NSImage?
+	init(previewImage: NSImage? = nil) {
+		self.previewImage = previewImage
+	}
+}
+
+struct PreviewView: View {
+	@ObservedObject var previewViewState: PreviewViewState // swiftlint:disable:this swiftui_state_private
+    var body: some View {
+		ZStack {
+			CheckerboardView()
+			VStack {
+				ImageView(image: previewViewState.previewImage ?? NSImage())
+					.scaledToFit()
+				Text("Preview")
+			}
+		}
+	}
+}
--- a/Gifski/TrimmingAVPlayer.swift
+++ b/Gifski/TrimmingAVPlayer.swift
@@ -9,12 +9,21 @@ struct TrimmingAVPlayer: NSViewControllerRepresentable {
 	var loopPlayback = false
 	var bouncePlayback = false
 	var speed = 1.0
+	// TrimmingAVPlayer + PreviewGenerator
+	var showPreview = false
+	var currentTimeDidChange: ((Double) -> Void)?
+	var previewImage: NSImage?
+	var previewAnimation: NSImage?
+	var animationBeingGeneratedNow = false
+	// end TrimmingAVPlayer + PreviewGenerator
 	var timeRangeDidChange: ((ClosedRange<Double>) -> Void)?
+
 
 	func makeNSViewController(context: Context) -> NSViewControllerType {
 		.init(
 			playerItem: .init(asset: asset),
 			controlsStyle: controlsStyle,
+			currentTimeDidChange: currentTimeDidChange,
 			timeRangeDidChange: timeRangeDidChange
 		)
 	}
@@ -27,7 +36,16 @@ struct TrimmingAVPlayer: NSViewControllerRepresentable {
 		nsViewController.loopPlayback = loopPlayback
 		nsViewController.bouncePlayback = bouncePlayback
 		nsViewController.player.defaultRate = Float(speed)
-		nsViewController.player.rate = nsViewController.player.rate > 0 ? Float(speed) : -Float(speed)
+		nsViewController.showPreview = showPreview
+		nsViewController.animationBeingGeneratedNow = animationBeingGeneratedNow
+		nsViewController.previewAnimation = previewAnimation
+		nsViewController.previewImage = previewImage
+
+		///  Feel free to take out this if statement.
+		///  The constant rewinding on setting change just really annoyed me. :)  -mmulet
+		if nsViewController.player.rate != 0 {
+			nsViewController.player.rate = nsViewController.player.rate > 0 ? Float(speed) : -Float(speed)
+		}
 	}
 }
 
@@ -42,6 +60,79 @@ final class TrimmingAVPlayerViewController: NSViewController {
 	private let controlsStyle: AVPlayerViewControlsStyle
 	private let timeRangeDidChange: ((ClosedRange<Double>) -> Void)?
 	private var cancellables = Set<AnyCancellable>()
+	private var previewView: NSHostingView<PreviewView>!
+
+	// TrimmingAVPlayerViewController+PreviewGenerator
+	private let currentTimeDidChange: ((Double) -> Void)?
+	private var periodicTimeObserver: Any?
+	private var rateObserver: NSKeyValueObservation?
+	@MainActor
+	var previewViewState = PreviewViewState(previewImage: nil)
+
+	fileprivate var previewAnimation: NSImage? {
+		didSet {
+			/// Only update the previewImage if we are
+			/// playing
+			guard player.rate != 0 else {
+				return
+			}
+			Task{
+				@MainActor in
+				self.previewViewState.previewImage = self.previewAnimation
+			}
+		}
+	}
+
+	fileprivate var previewImage: NSImage? {
+		didSet {
+			/// Only assign to the image if we are paused
+			guard player.rate == 0 else {
+				return
+			}
+			Task {
+				@MainActor in
+				self.previewViewState.previewImage = self.previewImage
+			}
+		}
+	}
+
+	/// On Showing or hiding the preview we
+	/// also have to hide the trim buttons
+	/// because we hide the play button if
+	/// the animation preview is not
+	/// available
+	fileprivate var showPreview = false {
+		didSet {
+			playerView.showPreview = self.showPreview
+			guard oldValue != self.showPreview else {
+				return
+			}
+			if self.showPreview {
+				playerView.contentOverlayView?.addSubview(previewView)
+				playerView.hideTrimButtons()
+			   previewView.constrainEdgesToSuperview() // Ensure it fills the parent view
+			} else {
+				playerView.hideTrimButtons()
+				previewView.removeFromSuperview()
+			}
+		}
+	}
+
+	///Have to update the player views' AnimationBeingGeneratedNow
+	///because it will have to hide the play button if the animation
+	///is unavailable
+	fileprivate var animationBeingGeneratedNow = false {
+		didSet {
+			playerView.animationBeingGeneratedNow = self.animationBeingGeneratedNow
+			guard oldValue != self.animationBeingGeneratedNow else {
+				return
+			}
+			playerView.hideTrimButtons()
+		}
+	}
+
+	// end TrimmingAVPlayerViewController+PreviewGenerator
+
 
 	var playerView: TrimmingAVPlayerView { view as! TrimmingAVPlayerView }
 
@@ -68,6 +159,7 @@ final class TrimmingAVPlayerViewController: NSViewController {
 		}
 	}
 
+
 	/**
 	Get or set the current player item.
 
@@ -93,17 +185,90 @@ final class TrimmingAVPlayerViewController: NSViewController {
 	init(
 		playerItem: AVPlayerItem,
 		controlsStyle: AVPlayerViewControlsStyle = .inline,
+		currentTimeDidChange: ((Double) -> Void)? = nil,
 		timeRangeDidChange: ((ClosedRange<Double>) -> Void)? = nil
 	) {
 		self.playerItem = playerItem
 		self.player = LoopingPlayer(playerItem: playerItem)
 		self.controlsStyle = controlsStyle
 		self.timeRangeDidChange = timeRangeDidChange
+		self.currentTimeDidChange = currentTimeDidChange
+
 		super.init(nibName: nil, bundle: nil)
+
+		/**
+		 We have to observe the rate to see whether we should be
+		 showing a still preview at the time you scrub to, or
+		 if we should play a preview of the entire animation.
+		 */
+		rateObserver = player.observe(\.rate, options: [.new, .old]) { [weak self] player, change in
+			guard let self else {
+				return
+			}
+			guard showPreview else {
+				return
+			}
+			/// If Animation is being generated, then
+			/// just show the preview image
+			if animationBeingGeneratedNow {
+				/// Avoid the infinite loop!
+				///  just stop it player.rate is already 0
+				guard change.newValue != 0 else {
+					return
+				}
+				player.rate = 0
+				Task{
+					@MainActor in
+					self.previewViewState.previewImage = self.previewImage
+				}
+
+				return
+			}
+
+			let should_show_animation = change.newValue != 0
+			Task {
+				@MainActor in
+				if should_show_animation {
+					self.previewViewState.previewImage = self.previewAnimation
+				} else {
+					self.previewViewState.previewImage = self.previewImage
+				}
+			}
+		}
+
+		/// https://developer.apple.com/documentation/avfoundation/monitoring-playback-progress-in-your-app
+		///	 This is where we keep track of how the user scrubbed on the timeline
+		///
+		///  Notice that the interval it set to a large value
+		///  This means that it will provide a callback in 2 cases.
+		///  1. On loop, in which case it will show the first frame (or last
+		///  	if you have bounce enabled)
+		///  2. When manually scrubbing. In which case it will show
+		///  	the frame you click on
+		///
+		///  You can set this to a lower interval (like 0.1) for a live preview
+		///  but performance will suffer
+		self.periodicTimeObserver = self.player.addPeriodicTimeObserver(
+			forInterval: CMTime(seconds: 1_000_000.0, preferredTimescale: 600),
+			queue: .main
+		)
+		{ [weak self] time in
+		   guard let self else {
+			   return
+		   }
+			guard let currentTimeDidChangeCallback = self.currentTimeDidChange else {
+				return
+			}
+			currentTimeDidChangeCallback(time.seconds)
+		}
 	}
 
 	deinit {
 		print("TrimmingAVPlayerViewController - DEINIT")
+		if let periodicTimeObserver {
+			player.removeTimeObserver(periodicTimeObserver)
+		}
+		rateObserver?.invalidate()
 	}
 
 	@available(*, unavailable)
@@ -112,6 +277,7 @@ final class TrimmingAVPlayerViewController: NSViewController {
 	}
 
 	override func loadView() {
+		previewView = NSHostingView(rootView: PreviewView(previewViewState: self.previewViewState))
 		let playerView = TrimmingAVPlayerView()
 		playerView.allowsVideoFrameAnalysis = false
 		playerView.controlsStyle = controlsStyle
@@ -156,6 +322,13 @@ final class TrimmingAVPlayerViewController: NSViewController {
 final class TrimmingAVPlayerView: AVPlayerView {
 	private var timeRangeCancellable: AnyCancellable?
 	private var trimmingCancellable: AnyCancellable?
+
+	/// TrimmingAVPlayerView + PreviewGenerator
+	/// These are needed to hide the play button
+	/// when an animation is not yet available
+	/// this is done in hideTrimButtons
+	fileprivate var showPreview = false
+	fileprivate var animationBeingGeneratedNow = false
 
 	/**
 	The minimum duration the trimmer can be set to.
@@ -243,6 +416,32 @@ final class TrimmingAVPlayerView: AVPlayerView {
 			.forEach {
 				$0.isHidden = true
 			}
+
+		if self.showPreview && animationBeingGeneratedNow {
+			/// Hide the play button when the animation is being generated
+			/// Only show it when you can play the animation
+			superview.subviews
+				.first { $0 != avTrimView }?
+				.subviews
+				.forEach {
+					$0.isHidden = true
+				}
+		} else {
+			/// Need to potentially unhide the play button
+			superview.subviews
+				.first { $0 != avTrimView }?
+				.subviews
+				.forEach {
+					$0.isHidden = false
+				}
+			superview.subviews
+				.first { $0 != avTrimView }?
+				.subviews
+				.filter { ($0 as? NSButton)?.image == nil }
+				.forEach {
+					$0.isHidden = true
+				}
+		}
 	}
 
 	fileprivate func addCheckerboardView() {
--- a/Gifski.xcodeproj/project.pbxproj
+++ b/Gifski.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0C12A7E42D8E4E6000EC7A25 /* PreviewView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0C12A7E32D8E4E6000EC7A25 /* PreviewView.swift */; };
+		0C12A7EA2D8E8AD300EC7A25 /* PreviewGenerator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0C12A7E92D8E8AD300EC7A25 /* PreviewGenerator.swift */; };
 		0E7925202329BDBE00058B94 /* ShareController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0E79251F2329BDBE00058B94 /* ShareController.swift */; };
 		0E7925282329BDBE00058B94 /* Share Extension.appex in Embed Foundation Extensions */ = {isa = PBXBuildFile; fileRef = 0E79251B2329BDBE00058B94 /* Share Extension.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
 		5FF0DFFB278BA5DB00A80F09 /* libgifski_static.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 5F6ABD84278BA5A20040DDF0 /* libgifski_static.a */; };
@@ -80,6 +82,8 @@
 /* End PBXCopyFilesBuildPhase section */
 
 /* Begin PBXFileReference section */
+		0C12A7E32D8E4E6000EC7A25 /* PreviewView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PreviewView.swift; sourceTree = "<group>"; };
+		0C12A7E92D8E8AD300EC7A25 /* PreviewGenerator.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PreviewGenerator.swift; sourceTree = "<group>"; };
 		0E79251B2329BDBE00058B94 /* Share Extension.appex */ = {isa = PBXFileReference; explicitFileType = "wrapper.app-extension"; includeInIndex = 0; path = "Share Extension.appex"; sourceTree = BUILT_PRODUCTS_DIR; };
 		0E79251F2329BDBE00058B94 /* ShareController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; lineEnding = 0; path = ShareController.swift; sourceTree = "<group>"; usesTabs = 1; };
 		0E7925242329BDBE00058B94 /* Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; path = Info.plist; sourceTree = "<group>"; };
@@ -216,11 +220,13 @@
 			children = (
 				E3AE62861E5CD2F300035A2F /* App.swift */,
 				E3A6BD102245345C00F62256 /* Constants.swift */,
+				0C12A7E32D8E4E6000EC7A25 /* PreviewView.swift */,
 				E33552F02ACAC3280023AAE9 /* AppState.swift */,
 				E3FC365B2377FA0000CF7C59 /* Shared.swift */,
 				E33552EE2ACAC3190023AAE9 /* MainScreen.swift */,
 				E33552F22ACAC5D80023AAE9 /* StartScreen.swift */,
 				E37F68E32ACADA40007F1A7F /* EditScreen.swift */,
+				0C12A7E92D8E8AD300EC7A25 /* PreviewGenerator.swift */,
 				E37F68E12ACAD9F1007F1A7F /* ConversionScreen.swift */,
 				E37F68DF2ACAD9D1007F1A7F /* CompletedScreen.swift */,
 				E3908B7326754568000723A7 /* EstimatedFileSize.swift */,
@@ -410,6 +416,7 @@
 			files = (
 				E3D08F6E1E5D7BFD00F465DF /* Utilities.swift in Sources */,
 				E339F011203820ED003B78FB /* GIFGenerator.swift in Sources */,
+				0C12A7E42D8E4E6000EC7A25 /* PreviewView.swift in Sources */,
 				E33552F12ACAC3280023AAE9 /* AppState.swift in Sources */,
 				E3A6BD112245345C00F62256 /* Constants.swift in Sources */,
 				C2040B8920435871004EE259 /* GifskiWrapper.swift in Sources */,
@@ -421,6 +428,7 @@
 				85BF910922F3279300AD3FF6 /* TrimmingAVPlayer.swift in Sources */,
 				8548806522B78E8300E97401 /* IntTextField.swift in Sources */,
 				E3961F802AC9F2A700708EB7 /* Intents.swift in Sources */,
+				0C12A7EA2D8E8AD300EC7A25 /* PreviewGenerator.swift in Sources */,
 				E30C8EEF29387E7A002E053F /* Gifski.swift in Sources */,
 				E3FC365C2377FA0000CF7C59 /* Shared.swift in Sources */,
 				E37F68E02ACAD9D1007F1A7F /* CompletedScreen.swift in Sources */,
--- a/Gifski/EditScreen.swift
+++ b/Gifski/EditScreen.swift
@@ -1,6 +1,8 @@
 import SwiftUI
 import AVFoundation
 
+
+
 struct EditScreen: View {
 	@Environment(AppState.self) private var appState
 	@Default(.outputQuality) private var outputQuality
@@ -20,6 +22,17 @@ struct EditScreen: View {
 	@State private var resizableDimensions = Dimensions.percent(1, originalSize: .init(widthHeight: 100))
 	@State private var shouldShow = false
 
+
+	/// EditScreen + PreviewGenerator
+	@State private var currentScrubbedTime: Double = .zero
+	@State private var showPreview = false
+	@StateObject private var singleFramePreviewGenerator = PreviewGenerator(
+		type: .singleFrame
+	)
+	@StateObject private var fullAnimationPreviewGenerator = PreviewGenerator(
+		type: .entireAnimation
+	)
+
 	init(
 		url: URL,
 		asset: AVAsset,
@@ -31,13 +44,20 @@ struct EditScreen: View {
 		self._metadata = .init(wrappedValue: metadata)
 	}
 
+
 	var body: some View {
 		VStack {
 			// TODO: Move the trimmer outside the video view.
 			TrimmingAVPlayer(
 				asset: modifiedAsset,
 				loopPlayback: loopGIF,
-				bouncePlayback: bounceGIF
+				bouncePlayback: bounceGIF,
+				showPreview: showPreview,
+				// EditScreen + PreviewGenerator
+				currentTimeDidChange: self.onCurrentTimeDidChange,
+				previewImage: singleFramePreviewGenerator.previewImage?.image,
+				previewAnimation: fullAnimationPreviewGenerator.previewImage?.image,
+				animationBeingGeneratedNow: fullAnimationPreviewGenerator.imageBeingGeneratedNow
 			) { timeRange in
 				DispatchQueue.main.async {
 					self.timeRange = timeRange
@@ -61,19 +81,24 @@ struct EditScreen: View {
 		.onChange(of: outputQuality, initial: true) {
 			estimatedFileSizeModel.duration = metadata.duration
 			estimatedFileSizeModel.updateEstimate()
+			onSettingsDidChange()
 		}
 		// TODO: Make these a single call when tuples are equatable.
 		.onChange(of: resizableDimensions) {
 			estimatedFileSizeModel.updateEstimate()
+			onSettingsDidChange()
 		}
 		.onChange(of: timeRange) {
 			estimatedFileSizeModel.updateEstimate()
+			onSettingsDidChange()
 		}
 		.onChange(of: bounceGIF) {
 			estimatedFileSizeModel.updateEstimate()
+			onSettingsDidChange()
 		}
 		.onChange(of: frameRate) {
 			estimatedFileSizeModel.updateEstimate()
+			onSettingsDidChange()
 		}
 		.onChange(of: bounceGIF) {
 			guard bounceGIF else {
@@ -81,6 +106,13 @@ struct EditScreen: View {
 			}
 
 			showKeyframeRateWarningIfNeeded()
+		}
+		.onChange(of: showPreview) {
+			guard showPreview else {
+				return
+			}
+			/// Force a redraw of the preview on
+			onCurrentTimeDidChange(currentTime: self.currentScrubbedTime)
 		}
 		.alert2(
 			"Reverse Playback Preview Limitation",
@@ -148,6 +180,11 @@ struct EditScreen: View {
 	private var bottomBar: some View {
 		HStack {
 			Spacer()
+			if showPreview && (singleFramePreviewGenerator.imageBeingGeneratedNow || fullAnimationPreviewGenerator.imageBeingGeneratedNow) {
+				ProgressView()
+					.frame(width: 10, height: 10)
+			}
+			Toggle("Show Preview", isOn: $showPreview)
 			Button("Convert") {
 				appState.navigationPath.append(.conversion(conversionSettings))
 			}
@@ -621,5 +658,32 @@ private struct LoopSetting: View {
 		SSApp.runOnce(identifier: "gifLoopCountWarning") {
 			isGifLoopCountWarningPresented = true
 		}
+	}
+}
+
+
+///  EditScreen + PreviewGenerator
+extension EditScreen {
+	private func onSettingsDidChange() {
+		guard showPreview else {
+			return
+		}
+		singleFramePreviewGenerator.onSettingsDidChangeGeneratePreview(conversionSettings: self.conversionSettings)
+		fullAnimationPreviewGenerator.onSettingsDidChangeGeneratePreview(conversionSettings: self.conversionSettings)
+	}
+
+	private func onCurrentTimeDidChange(currentTime: Double) {
+		self.currentScrubbedTime = currentTime
+		guard showPreview else {
+			return
+		}
+		self.singleFramePreviewGenerator.onCurrentTimeDidChange(
+			currentTime: currentTime,
+			conversionSettings: conversionSettings
+		)
+		self.fullAnimationPreviewGenerator.onCurrentTimeDidChange(
+			currentTime: currentTime,
+			conversionSettings: conversionSettings
+		)
 	}
 }


sindresorhus:
- The spinner besides the preview is too big.
- Don't need the "Preview" text on the video. The checkbox makes it clear enough.
- The "Show PReview" checkbox doesn't look that good.

sindresorhus:
I ended up with empty preview many times:

Screen Shot 2025-04-09 at 03 48 30
Screen Shot 2025-04-09 at 03 34 10


---

Test video:

https://github.com/user-attachments/assets/b372bbeb-4554-4537-9343-dec8cfd25349

sindresorhus (changes requested):

mmulet:
## Changes

As usually here is a new demo video:

https://github.com/user-attachments/assets/eda28e2b-52f0-4cb9-a74d-556206e6ba19




- [x] The spinner besides the preview is too big.
- [x] Don't need the "Preview" text on the video. The checkbox makes it clear enough.
- [x] Improve Visual for the "Show Preview" checkbox
- [x] FIX: I ended up with empty preview many times:
  - fixed
- [x] Don't hide PlayButton, just disable 
- [x] Implement reusable getter/setter to hide/show the PlayButton on Trimmer
- [x] Preview Logic OUtside of trimming AVPlayer
  - All the logic has been moved out of trimmingAVplayer and moved into the PreviewViewState
- [x] Don't hard-wrap, playback rate
- [x] Instead of PeriodicTimeObserver, Add a AVPlayer extension that wraps addPeriodicTimeObserver in a asyncstream and use that instead. It would make it easier to use. It could even accept a Duration instead, for improved usability.
  - Moved the peridicTimerObserver to a new stream focued on when the trimer scrubs to a new time
- [x] /// Instead of /**
- [x] Use Combine. for playerRateObserer (m)
- [x] PreviewGenerator to Observable
- [x] Cleanup PreviewGenerator 
  - Much better now, switched form my old-timey thread loop to an AsyncStream. It has been completely reworked

I found some more things that needed to be fixed while I was at it:

- [x] Fix not having enough frames for a preview
- [x] Fix trimmed video callback happening on loopback, bounce, or play button press
- [x] Fix video playback not syncing up with preview
- [x] Fix scrubbing preview not pausing the gif (have to pause the gif because there are no native gif pausing)

One last thing:

While testing the preview, I noticed a bug where it would flash the wrong frame. I confirmed that this is not a problem with preview, it is a problem with the export itself, if you convert the video it has the same results, the same bug, you can see it here:

https://github.com/user-attachments/assets/3bc49c05-fd2f-47c0-abf3-6e1508a3725f

Gifski/Utilities.swift
```diff
@@ -3060,21 +3065,76 @@ final class LoopingPlayer: AVPlayer {
...
 					playImmediately(atRate: defaultRate)
+					DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+						self.timeChangeDueToLoopBounceOrPlayButtonPress = false
+					}
+				}
+			}
+	}
+}
+
+extension LoopingPlayer {
```
sindresorhus, line 3106: Should be AVPlayer extension

Gifski/Utilities.swift
```diff
@@ -3060,21 +3065,76 @@ final class LoopingPlayer: AVPlayer {
...
 					seekToEnd()
 					playImmediately(atRate: -defaultRate)
+					DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+						self.timeChangeDueToLoopBounceOrPlayButtonPress = false
+					}
 				} else if loopPlayback {
+					timeChangeDueToLoopBounceOrPlayButtonPress = true
 					seekToStart()
 					playImmediately(atRate: defaultRate)
+					DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
```
sindresorhus, line 3098:
```suggestion
					delay(.seconds(0.1)) {
```

Gifski/Utilities.swift
```diff
@@ -3060,21 +3065,76 @@ final class LoopingPlayer: AVPlayer {
...
+					/**
+					 Not having this pause causes a glitch on the loop
+					 during normal playback.
+
+					 But having the pause causes a glitch on the loop
+					 of preview playback because it will set the rate
+					 to zero causing it to enter pause mode but
+					 will immediatly unpause causing a flash of unwanted
+					 content
+					 */
```
sindresorhus, line 3078: Incorrect indentation.

Gifski/TrimmingAVPlayer.swift
```diff
@@ -210,7 +284,51 @@ final class TrimmingAVPlayerView: AVPlayerView {
 		.toCancellable
 	}
 
-	fileprivate func hideTrimButtons() {
+	var playPauseButtonTarget: AnyObject?
+
+	fileprivate func enableOrDisablePlayButton(enable: Bool) {
```
sindresorhus, line 289:
```suggestion
	fileprivate func setPlayButtonEnabled(_ enabled: Bool) {
```

Gifski/TrimmingAVPlayer.swift
```diff
@@ -150,9 +206,27 @@ final class TrimmingAVPlayerViewController: NSViewController {
...
+			.receive(on: DispatchQueue.main)
+			.sink { [weak self] newRate in
+				guard let self else {
+					return
+				}
+				rateDidChange?(player, newRate)
+			}
+			.store(in: &cancellables)
+		Task {
+			for await time in  self.player.scrubTimeStream() {
```
sindresorhus, line 221:
```suggestion
			for await time in self.player.scrubTimeStream() {
```

Gifski/TrimmingAVPlayer.swift
```diff
@@ -150,9 +206,27 @@ final class TrimmingAVPlayerViewController: NSViewController {
...
+			.publisher(for: \.rate)
+			.receive(on: DispatchQueue.main)
+			.sink { [weak self] newRate in
+				guard let self else {
+					return
+				}
+				rateDidChange?(player, newRate)
+			}
+			.store(in: &cancellables)
+		Task {
```
sindresorhus, line 220:
```suggestion

		Task {
```

Gifski/TrimmingAVPlayer.swift
```diff
@@ -42,6 +62,34 @@ final class TrimmingAVPlayerViewController: NSViewController {
 	private let controlsStyle: AVPlayerViewControlsStyle
 	private let timeRangeDidChange: ((ClosedRange<Double>) -> Void)?
 	private var cancellables = Set<AnyCancellable>()
+	private let onScrubToNewTime: ((AVPlayer, Double) -> Void)?
+	private let rateDidChange: ((AVPlayer, Float) -> Void)?
+
+
+	fileprivate var enablePlayButton = true {
```
sindresorhus, line 69:
```suggestion
	fileprivate var isPlayButtonEnabled = true {
```

Gifski/TrimmingAVPlayer.swift
```diff
@@ -9,12 +9,22 @@ struct TrimmingAVPlayer: NSViewControllerRepresentable {
 	var loopPlayback = false
 	var bouncePlayback = false
 	var speed = 1.0
+
+
+	var onScrubToNewTime: ((AVPlayer, Double) -> Void)?
+	var rateDidChange: ((AVPlayer, Float) -> Void)?
```
sindresorhus, line 15: Should not pass in AVPlayer to these.

Gifski/PreviewView.swift
```diff
@@ -0,0 +1,189 @@
...
+			 */
+			guard newRate != 0 else {
+				return
+			}
+			player.rate = 0
+			return
+		}
+		shouldShowAnimation = newRate != 0
+
+		if newRate > 0 && previousRate == 0 {
```
sindresorhus, line 132:
```suggestion
		if newRate > 0, previousRate == 0 {
```

Gifski/PreviewView.swift
```diff
@@ -0,0 +1,189 @@
...
+		}
+
+		guard showPreview else {
+			return
+		}
+		if fullAnimationPreviewGenerator.previewImage == nil {
+			shouldShowAnimation = false
+			/**
+			 Avoid infinite loop: stop the player if rate is already 0
+			 */
```
sindresorhus, line 123:
```suggestion
			Avoid infinite loop: stop the player if rate is already 0.
			*/
```

Gifski/PreviewGenerator.swift
```diff
@@ -0,0 +1,220 @@
...
+		}
+	}
+
+	/**
+	 An async Stream that will yield when a new item is added,
+	 but keeps an up to date sequence of items so you can
+	 know if you are processing the latest item inserted into
+	 the stream.
+	 */
+	private struct LatestCommandAsyncStream: AsyncSequence {
```
sindresorhus, line 193: I don't see why this is an asyncsequence struct. could just be a function that returns an asyncstream.

Gifski/PreviewGenerator.swift
```diff
@@ -0,0 +1,220 @@
...
+				guard commandStream.commandIsLatest(command: item) else {
+					continue
+				}
+				if previewImageCommand == item.command {
+					/**
+					 Don't regenerate if it's the same
+					 */
+					continue
+				}
+				Task {
```
sindresorhus, line 39: Task is not synchronous, so you are introducing race issues.

Use `await MainActor.run {}` instead.

Gifski/TrimmingAVPlayer.swift
```diff
@@ -210,7 +284,51 @@ final class TrimmingAVPlayerView: AVPlayerView {
...
+				return true
+			} as? NSButton) else {
+				return
+			}
+		if playPauseButton.target !== self {
+			playPauseButtonTarget = playPauseButton.target
+			playPauseButton.target = self
+		}
+
+		playPauseButton.isEnabled = enable
```
sindresorhus, line 313: This should be a AVPlayerView extension called `trimViewPlayPauseButton`. Then you could simplify `fileprivate var enablePlayButton = true {` to be a getter/setter that proxies to `playPauseButton.isEnabled` directly.

And `playPauseButtonTarget = playPauseButton.target` can be done in `init`.

Gifski/TrimmingAVPlayer.swift
```diff
@@ -210,7 +284,51 @@ final class TrimmingAVPlayerView: AVPlayerView {
...
+	}
+
+	@objc func playPauseButtonPressed(_ sender: Any) {
+		if let loopingPlayer = player as? LoopingPlayer,
+		/**
+		 only prevent time change notifications on play button press,
+		 not pause button press
+		 */
```
sindresorhus, line 321: Better formatting and the comment should explain why it's needed.

sindresorhus:
- In preview, if I pause in the middle and resume, it starts from the beginning, not the paused position.
- The "Show preview" button should be a toggle with a SF Symbol icon, just like with crop.
	- It should also be in the menu bar menu, like with crop.
- When the drag the position handle (the red line), I often see the huge progress indicator show, then hide, then show. It flickers.
- When it has converted the whole video to a GIF and I play and then pause, it still shows the huge progress indicator momentarily. It should not show.
- If I click "Show preview", let it convert the whole video to GIF, then press "Hide preview", and then "Show preview" again, it re-renders the whole GIF. Nothing changed, so it should be able to reuse the existing GIF. A user may want to be able to quickly toggle back and forth to compare quality with the original.
- The code still needs a lot of minor cleanup. Don't hard-wrap code comments.
- Would be great if you could go over the code and try to simplify it. It's becoming quite complex.

sindresorhus (changes requested):

sindresorhus:
> While testing the preview, I noticed a bug where it would flash the wrong frame. I confirmed that this is not a problem with preview, it is a problem with the export itself, if you convert the video it has the same results, the same bug, you can see it here:

Seems like a bug in the "Bounce" logic.

Gifski/Utilities.swift
```diff
@@ -3060,21 +3065,76 @@ final class LoopingPlayer: AVPlayer {
...
+					DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+						self.timeChangeDueToLoopBounceOrPlayButtonPress = false
+					}
+				}
+			}
+	}
+}
+
+extension LoopingPlayer {
+	func scrubTimeStream(forInterval duration: Duration = .seconds(1_000_000.0), queue: DispatchQueue = .global(qos: .utility)) -> AsyncStream<Duration> {
```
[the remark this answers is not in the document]
mmulet: I renamed it to scrubTimeStream because I reorganized it to a stream that emits events only when the user finishes scrubbing.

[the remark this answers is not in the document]
mmulet: Similar to what I said above, the 1_000_000.0 seconds is for the purpose of only emitting events when the user scrubs, not during normal playback.

Gifski/Utilities.swift
```diff
@@ -3060,21 +3065,76 @@ final class LoopingPlayer: AVPlayer {
...
+			 but performance will suffer
+			 */
+			let periodicTimeObserver = addPeriodicTimeObserver(
+				forInterval: CMTime(seconds: duration.toTimeInterval, preferredTimescale: CMTimeScale(NSEC_PER_SEC)),
+				queue: queue
+			)
+			{ time in
+				guard !self.timeChangeDueToLoopBounceOrPlayButtonPress else {
+					return
 				}
```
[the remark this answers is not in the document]
mmulet: Not a general stream anymore, this stream just reflects scrubTime event. So, this is necessary to reject invalid events (not caused by a user scrub)

mmulet:
> * In preview, if I pause in the middle and resume, it starts from the beginning, not the paused position.
- That is because it's the preview outputs a GIF, and there are no GIF playback controls. All I can do is start at beginning or stop. The only solution to this is to make a custom GIF SwiftUI player, perhaps with `CGAnimateImageDataWithBlock(_:_:_:)`. I will probably take this approach.

> * When it has converted the whole video to a GIF and I play and then pause, it still shows the huge progress indicator momentarily. It should not show.
- When you Pause it takes a moment to generate the preview image at the current pause time. If I switch to a custom GIF player, this shouldn't be a problem anymore.

Gifski/TrimmingAVPlayer.swift
mmulet: Switching to a custom GIF player should also fix this.

Gifski/Utilities.swift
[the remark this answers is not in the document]
sindresorhus: That should be a separate stream that `map` this one. Separation of concerns.

sindresorhus:
> The preview outputs a GIF, which is directly passed to a NSImage and there are no GIF playback controls. All I can do is start at beginning or stop. The only solution to this is to make a custom GIF SwiftUI player, perhaps with CGAnimateImageDataWithBlock(_:_:_:). I will probably take this approach.

Wouldn't it be easier to make a video out of the GIF frames? Then you could reuse everything we already have since it's just a video.

mmulet:
> > The preview outputs a GIF, which is directly passed to a NSImage and there are no GIF playback controls. All I can do is start at beginning or stop. The only solution to this is to make a custom GIF SwiftUI player, perhaps with CGAnimateImageDataWithBlock(_:_:_:). I will probably take this approach.
> 
> Wouldn't it be easier to make a video out of the GIF frames? Then you could reuse everything we already have since it's just a video.

I actually did try that earlier but I ran into 2 problems:
1. Re-encoding the gif increased the time to preview by at least 2x.
2. If I used a lossless format to preserve the preview exactly, the memory usage ballooned.
 - If I used a lossy format, how does the user know if a gif preview looks bad due to compression artifacts from the lossy preview or due to the my gif quality?

Gifski/PreviewGenerator.swift
mmulet: It's an AsyncSequence because it keeps track of the latest item pushed to the queue.

Generating a full preview can take 5-15 seconds, so imagine that a user scrubs the Quality meter a lot (let's say 5 times). If we had a stream it would generate a preview for each of the intermediate points and it would be (5-15) x 5 = 25 - 75 seconds until the correct preview appears. Using the sequence we can skip all the intermediates, and just generate the first and last preview. So now it will be only 10-30 seconds until the preview appears.

sindresorhus: > It's an AsyncSequence because it keeps track of the latest item pushed to the queue.

And you can still do that with a function. You just store it as a local variable inside the function and then return the stream.

mmulet: Sorry for the trouble, but I'm a bit confused on this one point. 

- Yes, I know we can keep the item in a closure
But 
- how do we refer to the item later to check if it is latest
- How do we add items to the stream?

Like this pseudocode?:

Return a tuple?

```swift
func getStream() -> (AsyncStream<...>, IsLatestFunc, AddItemFunc) {
  var asyncContinuation: ...
  let stream = AsyncStream { continuation in 
    asyncContinuation = continuation
  }

  var latest:....
  func addItem(...) {
    latest = ...
    asyncContinuation.yield(...)
  }
  func isLatest(item:....) -> Bool {
    latest == item
  }
  return (stream, isLatest, addItem)

}
```

Or is there something built in to AsyncStream that I'm missing?

sindresorhus: > You know what, we can just sidestep this entire issue if we just make the preview GIF conversion cancellable, then when it is converting and another event comes in, we can just cancel the current version and only emit when its done.

That would be even better. Things should generally be cancellable when possible.

mmulet:
I uploaded the commit with the video player, I would say it's in progress, but not finished. 

Overall, I have mixed feelings about it.

On one hand, it does simplify things quite a bit, I was able to simplify most of the code into pure functions and I moved almost everything out of TrimmingAVPlayer, and there is no "PreviewState", "PreviewGenerator", or "LatestItemStream". 

On the other, AVPlayerView is generally glitchy and hard to work with. I've made some good progress and squashed the major bugs, but I'm sure you will have no problem finding more.

For example:
- When the preview generates, and we switch the AVAssets to the TrimmingAVPlayer, it causes a flash of (unrelated) video content to the screen. For the sake of toggling on and off the preview, I got around this by moving the preview view to a second track in a AVMutableComposition, and creating two separate AVMutableVideoComposition that you can switch at runtime. So that you can toggle the preview on/off, the asset stays the same, it's just changes videoComposition. 
 
I had hoped I could do something similar for the EditScreen overall. Instead of swapping in and out AVAssets to the TrimmingAVPlayer, I could have one AVMutableComposition and I could swap in and out the tracks as you update the preview. But, I couldn't get this to work reliably, I had assumed you could just 

```swift
track.removeTimeRange(...)
track.insertTimeRange(updatedPreview)
```
but I just end up with blank screen. And everything else I tried all needed an updated AVPlayerItem, which just got me back to where I started.

-----
Last but not least, when you export a preview of a smaller size, it surrounds the entire rest of the video with a black background. This background, as far as I can tell, is baked into the AVPlayer. Nothing I have tried has gotten rid of it. There are only two solutions that I can think of:

1. Create a video of the checkerboard pattern, calculated to the exact height and width needed to blend into the background and add another track below all the previous tracks, and place it there.
2. Give up having the preview and the regular video in one AVMutableComposition, but then we have the AVAsset swapping problem mentioned above. (we would also have to introduce emptyRanges into the track so the trimmer view would match up between preview and normal mode, and I'm not sure if this would also cause a black box to appear over the background)

mmulet:
Took longer than expected, but I got it working such that the preview is incredibly smooth.

## Overview of changes
First, here is what the code does:
1. As soon as you enter edit screen you can press the preview button in the toolbar to instantly show a preview of the current frame.
2. This works in pause and play mode, but in play mode it will preform slowly**, until..
3. The complete video (of just the trim range) is converted to GIF, henceforth I will call this a FullPreview.
4. Once the video is converted to GIF, it will play smoothly in the preview.
5. At all times, you can toggle the preview on and off to compare the original video with the GIF.
6. After you change quality you can see the change reflected in the preview (speed is limited by Gifski at the moment), and it will generate a new FullPreview.
7. This works with all settings such as speed, trim range, frame rate, and size.

** I'm debating if I should disable the play button like I used to. One one hand it might be jarring to see a choppy preview at first, on the other it's nice to have a fast limited frame rate preview. One of your the original ideas for this was to render about 5 frames spread out to give a good idea of the video, and I think this implementation does that.

## Implementation
The new code is conceptually very simple, there are just two main parts 
1. The preview is implemented by an [avvideocompositing](https://developer.apple.com/documentation/avfoundation/avvideocompositing), `PreviewVideoCompositor` this subclass is responsible for rendering the preview as well as the original video. A custom compositor may seem like overkill, but it is the only way to get smooth playback of the GIF, and to handle some corner cases with the AVPlayer. When a FullPreview is available, it will use that, otherwise it will convert the video to a GIF using Gifski on the fly.
2. The generation of new FullPreview is handled by `createPreviewStream`. Which returns a function you can call to start a new full preview and an AsyncStream which emits events when the preview on progress and when it is finished.


Code flow:
- `EditScreen` starts the stream with `createFullPreviewStream`, and setups of the compositor with `PreviewableComposition`. Previewable compositor is a AVMutableComposition with 2 tracks. One for the original video from the asset, and the second for the full preview. `PreviewableComposition` also has a custom compositor, `PreviewVideoCompositor`, which is responsible for rendering the preview.
- The `PreviewVideoCompositor` now works, and in the absence of a FullPreview will render each video frame on the fly. I added a simple function to `GIFGenerator`, `convertOneFrame`, to do this on the fly rendering. Depending on the video size and quality, this can be slow.  So we still need the FullPreview.
- On any setting change, the `EditScreen` will `requestFullNewPreview` to request a new FullPreview.For each new setting it checks if it needs to re-render the GIF, and if so it will convert the video to a GIF, then convert that GIF to a video as an AVAsset.
- Once created, the stream will emit an event containing the new FullPreview, at which point the `EditScreen` will will update the `PreviewableComposition` by inserting the new FullPreview `AVAsset` into the second track.
- Rinse and repeat.

## Misc.
- I put a lot of effort into making the code as simple and clean as possible, but it did balloon up a bit. This is mostly because AVPlayer has a lot of quirks that I had to fix to give the best possible user experience. So if you see something that looks odd, there is probably a reason for it. I tried to document most things I would consider odd, with the exact reason why I did it that way.
- When compositing a preview smaller than the source video, I just blur the original and put it behind the preview. This is easy to change, so we can make it whatever we want.
- I made sure not to hard wrap any comments this time :)

sindresorhus:
Great to see you managed to find a solution. From some quick testing, it seems to work.

---

I am a little worried about all the code though. The logic has gotten significantly more complicated now, which increases the chances of crashes and bugs.

---

I noticed some color difference now in non-preview mode.

This PR:

Screen Shot 2025-05-01 at 23 12 13

Main branch:


Screen Shot 2025-05-01 at 23 13 04

mmulet:
> Great to see you managed to find a solution. From some quick testing, it seems to work.


> I am a little worried about all the code though. The logic has gotten significantly more complicated now, which increases the chances of crashes and bugs.

I feel that, as I mentioned above, AVPlayer had many subtle bugs that had to workaround to get something that worked well. I simplified as much as possible, but not simpler. Although it is a lot as code, I tried as much as possible to keep everything modeled as pure functions (ie modular) so that we don't have a big pile of state to debug and maintain. I'm trying to keep each part as self-contained as possible.

> I noticed some color difference now in non-preview mode.

Good catch. It looks like a color space issue. I literally memcpy the pixel data, so it must be a buffer attribute. I will look into this.

mmulet:
The problem was exactly what I thought it was. An easy 3 line fix.

mmulet:
For this commit I followed the same cleanup rules from the crop video branch:

- [x] Asked ChatGPT for bug fixes/ code simplification for all changes
- [x] Looked at the diff from main to find any leftovers / style issues
- [x] Removed unnecessary returns from switch statements

Gifski/Preview/LockedCVPixelBuffer.swift
```diff
@@ -0,0 +1,139 @@
...
+
+import Foundation
+import CoreVideo
+
+/**
+ A CVPixelBuffer that has had it's base address locked via [CVPixelBufferLockBaseAddress](CVPixelBufferLockBaseAddress)
+ */
+protocol LockedCVPixelBuffer {
+	var buf: CVPixelBuffer {get}
+}
```
sindresorhus, line 16: I wonder if it would be simpler with just a helper method:

```swift
extension CVPixelBuffer {
 func withLocked<T>(flags: CVPixelBufferLockFlags = [], _ body: (UnsafeMutableRawPointer, Int, Int) throws -> T) rethrows -> T? {
  guard CVPixelBufferLockBaseAddress(self, flags) == kCVReturnSuccess,
        let base = CVPixelBufferGetBaseAddress(self) else { return nil }
  defer { CVPixelBufferUnlockBaseAddress(self, flags) }
  return try body(base,
                  CVPixelBufferGetBytesPerRow(self),
                  CVPixelBufferGetHeight(self))
 }
}
```

Gifski/Preview/LockedCVPixelBuffer.swift
```diff
@@ -0,0 +1,139 @@
...
+			return nil
+		}
+		return out
+	}
+}
+
+/**
+ Helper class to make [LockedCVPixelBuffer.copy](LockedCVPixelBuffer.copy(to:)) more elegant. It handles copying for planar and non planer frames in the same way
+ */
+fileprivate struct PixelBufferByteCopier<B: LockedCVPixelBuffer> {
```
sindresorhus, line 79: This would likely be more efficient as a free function:

```swift
func copy(_ src: CVPixelBuffer, to dst: CVPixelBuffer) -> Bool {
```

sindresorhus:
I would also feed ChatGPT all the preview code and ask it to find memory issues. It's easy to get things wrong with low-level code like pixel buffers.

sindresorhus (changes requested):

mmulet:
- [x] Merge the main branch (with the crop video changes)
- [x] use `@Bindable` instead of `binding(for:)`
- [x] Feed ChatGPT all the preview code and ask it to find memory issues. It's easy to get things wrong with low-level code like pixel buffers
 - ChatGPT said the code is robust.
- [x] use checkerboard background during preview
- [x] Use `originalFrame.buf.propagateAttachments(to: outputFrame.buf)` to copy attachments
- [x] Remove inouts from readWritable and so on.
- [x] `let planes = planeCount == 0 ? [nil] : Array(0..<planeCount).map { Optional($0) }` -> `for planeIndex in 0..<max(1, planeCount)`
- [x] Instead of `LockedPixelBuffer` classes refactor to use closures
- [x] `PixelBufferByteCopier` to free function

## Preview and Crop together
- [ ] allow preview of cropped video
I haven't implemented this yet, because I have a couple of different ideas for how to do it. Here are the two main ones I thought of:

1.  Allow preview and crop at the same time, but don't show a preview of the crop. In other words render out the entire frame of the video, only cropping it when you actually export it. This way you can adjust the crop while doing a preview of the video.

### Upsides
- Can see things like quality, speed, and framerate, while cropping. 

### Downsides
- You don't see a preview of the crop
- May take longer to render the preview if doing a long video

------

2. Hide the crop controls while previewing, but show the cropped video in the preview.
### Upsides
- Can see the cropped video while previewing
- faster

### Downsides
- This hides the crop controls while previewing, so if you turn on the crop while you preview, crop might appear to not work. (this could be fixed by showing a tooltip saying "Crop is disabled while previewing" or something like that)

------

3. Let the user choose between the two options, by replacing the preview and crop toolbar items with a menu that changes the current mode.
- Normal mode
- Crop mode
- Preview mode
- Edit crop and Preview mod

I know you will have an opinion on this, so I wanted to get your thoughts before I implement it.

Gifski/Preview/createAVAssetFromGIF.swift
```diff
@@ -0,0 +1,157 @@
...
+			pixelBufferAdaptor.append(pixelBuffer, withPresentationTime: presentationTime)
+		}
+		if frameIndex >= numberOfImages {
+			break
+		}
+	}
+
+	writerInput.markAsFinished()
+	await withCheckedContinuation { continuation in
+		assetWriter.finishWriting {
```
sindresorhus, line 105: `assetWriter.finishWriting()` is already async

Gifski/Preview/createAVAssetFromGIF.swift
```diff
@@ -0,0 +1,157 @@
...
+				progressThreshold = progress + 0.05
+			}
+			defer {
+				frameIndex += 1
+			}
+			guard let cgImage = CGImageSourceCreateImageAtIndex(imageSource, frameIndex, nil) else {
+				continue
+			}
+			guard let pixelBufferPool = pixelBufferAdaptor.pixelBufferPool,
+				  let pixelBuffer = createPixelBuffer(from: cgImage, using: pixelBufferPool)
```
sindresorhus, line 87: Should it really silently skip frames?

Gifski/Preview/createAVAssetFromGIF.swift
```diff
@@ -0,0 +1,157 @@
...
+		return CMTimeScale(30.0)
+	}()
+
+
+
+	let dataReadyStream = AsyncStream { continuation in
+		writerInput.requestMediaDataWhenReady(on: dispatchQueue) {
+			continuation.yield()
+		}
+	}
```
sindresorhus, line 70: Not sure I see the point of making this an asyncstream.

Gifski/Preview/createAVAssetFromGIF.swift
```diff
@@ -0,0 +1,157 @@
...
+				  let pixelBuffer = createPixelBuffer(from: cgImage, using: pixelBufferPool)
+			else {
+				continue
+			}
+
+			let presentationTime = CMTime(
+				value: CMTimeValue(frameIndex),
+				timescale: frameRate
+			)
+			pixelBufferAdaptor.append(pixelBuffer, withPresentationTime: presentationTime)
```
sindresorhus, line 96: Should maybe check the return value.

Gifski/Preview/compositePreview.metal
```diff
@@ -0,0 +1,138 @@
+//
+//  compositePreview.metal
+//  Gifski
+//
+//  Created by Michael Mulet on 4/22/25.
+//
+
```
sindresorhus, line 7: I have no experience with Metal, but maybe some of the ChatGPT feedback can be useful: https://chatgpt.com/share/682df120-5d28-8010-933c-5eee3f41fcca

sindresorhus:
Do we actually need Metal for this? I asked ChatGPT and it said we could solve it with CoreImage, and it seems to work ok: https://github.com/sindresorhus/Gifski/commit/5a56a7cc3bbc426807f303c72aa75ff63dbdb75a

sindresorhus (changes requested):

mmulet:
> > Do we actually need Metal for this? I asked ChatGPT and it said we could solve it with CoreImage, and it seems to work ok: [5a56a7c](https://github.com/sindresorhus/Gifski/commit/5a56a7cc3bbc426807f303c72aa75ff63dbdb75a)
> 
> I had a solution using CoreImage that I tried before moving to metal, but it really suffered from performance issues. But, maybe I was doing something wrong! I will compare this commit to my old branch and see if there is a performant way to use CoreImage to do this.

I just tested it, and Metal is still about 4-5x less CPU usage than the CoreImage code. But, the important thing is that they are both realtime (no frame drops) for fullHD. So, if you wanted to make the codebase simpler I could see switching to pure CoreImage.

Gifski/Preview/createAVAssetFromGIF.swift
mmulet: I believe I made it an `AsyncStream` because `imageSource`, `writerInput` , and `pixelBufferAdaptor` are non-sendable so I couldn't pass them directly to `requestMediaDataWhenReady` 's callback block. I could bypass this entire thing by using the publisher for `isReadyForMoreMediaData` rather than using `requestMediaDataWhenReady`:

```swift
var cancelable: AnyCancellable?
defer {
 cancelable?.cancel()
}
try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Void, Error>) in
cancelable = writerInput.publisher(for: \.isReadyForMoreMediaData).sink { isReady in
	guard isReady else {
		return
	}
	var progressThreshold = 0.05
	while writerInput.isReadyForMoreMediaData && frameIndex < numberOfImages {
		guard !Task.isCancelled else {
			continuation.resume(throwing: CancellationError())
			return
		}
		let progress = Double(frameIndex) / Double(numberOfImages)
		if progress > progressThreshold {
			onProgress(progress)
			progressThreshold = progress + 0.05
		}
		defer {
			frameIndex += 1
		}
		guard let cgImage = CGImageSourceCreateImageAtIndex(imageSource, frameIndex, nil) else {
			continue
		}
		guard let pixelBufferPool = pixelBufferAdaptor.pixelBufferPool,
					let pixelBuffer = createPixelBuffer(from: cgImage, using: pixelBufferPool)
		else {
			continue
		}

		let presentationTime = CMTime(
			value: CMTimeValue(frameIndex),
			timescale: frameRate
		)
		guard pixelBufferAdaptor.append(pixelBuffer, withPresentationTime: presentationTime) else {
			continuation.resume(throwing: CreateAVAssetError.cannotAppendNewImage)
			return
		}
	}
	if frameIndex >= numberOfImages {
		continuation.resume(returning: Void())
	}
}
}
```

mmulet:
> I don't remember why we didn't do it initially, but maybe we could make the crop persistent, so even if the user exits the crop edit mode (by clicking the crop icon again), the crop is still applied and we still show the black overlay indicating where the crop is, just not the white crop rect handles? The user can simple go back to crop mode and reset the crop not crop the video. I think this may be more ideal behavior.
> 
> If we do the above, the situation with preview would be easier, because if the user clicks the preview button, we would just exit crop edit mode.

Now that crop edit mode and preview mode are mutually exclusive, should the toolbar items in the top right corner be a Menu or a Picker instead of two separate buttons?

#### Pros
- It would be immediately obvious that you can only enter one of those two modes at a time
- No surprises, like being surprised that the crop editor goes away when you hit preview button and vice-versa

#### Cons
- Might make discoverability a little harder, as you don't immediately see the other two buttons

sindresorhus:
> So, if you wanted to make the codebase simpler I could see switching to pure CoreImage.

I would prefer to use CoreImage if it has no other downsides than being less efficient.

Gifski/Preview/createAVAssetFromGIF.swift
sindresorhus: I think you can use `assumeIsolated`.

https://forums.swift.org/t/how-can-i-safely-use-a-non-sendable-object-in-an-async-context/76520/12

Alternatively, a Sendable box: 

```swift
final class SendableWrapper<T>: @unchecked Sendable {
    private var unsafeValue: T

    private let lock = NSLock()

    var value: T {
        get {
            lock.withLock { unsafeValue }
        }
        set {
            lock.withLock { unsafeValue = newValue }
        }
    }

    init(_ value: T) {
        unsafeValue = value
    }
}
```

mmulet: Hmmm, I tried to go the `actor` route, like in the forum thread you linked, and it did work, but I couldn't get around creating a new `Task` inside `isReadyForMoreMediaData` which would mean the adding frame task would not be cancelable anymore. So, I went with the Sendable box approach.

I did end up asking ChatGPT, and it just told me to use an `AsyncStream`

mmulet:
- [x] move to CoreImage
- [x] make crop persistent, and make it work with the preview
- [x] add and use `CVPixelBuffer` helper extension
- [x] `CVPixelBuffer.copy(to:)` errors instead of `Bool` return value
- [x] `LockedPixelBufferPlane` -> `CVPixelBuffer.LockedPlane`
- [x] `makeANewBufferThatThisCanCopyTo` -> `makeCompatibleBuffer`
- [x] `withLocked` -> `withLockedBaseAddress` + `withLockedPlanes`
- [x] use async version of `assetWriter.finishWriting()`
- [x] Throw on error in `createAVAssetFromGIF` instead of skipping framess
- [x] `AsyncStream` to checked continuation in createAVAssetFromGIF
- [x] Check `pixelBufferAdaptor.append` return value
- [x] `createAVAssetFromGIF` `AsyncStream` to `Sendable boxed` values

## Other bugs
- [x] Fixed a bug where if you rapidly press the crop edit button multiple times, sometimes it would hide the trimmer view when it shouldn't.

## outputCropRect
I need to change the preview when the `outputCropRect` changes, Simple enough, right? I tried to do this with the following code:
```swift
.onChange(of: outputCropRect) {
  //This is literally left blank
}
```
Then the the crop edit controls no longer work, they more a tiny bit then get "stuck".

see video:

https://github.com/user-attachments/assets/d02fffae-94a0-4d13-96e0-f3dde16b80e8


I discovered what is happening is that whenever the body of the view is re-evaluated it must end up changing the `NSHostingView` of the `CropOverlayView` causing the `dragGesture` to stop working. The problem is that if you observe outputCropRect in any way in the view hierarchy, it causes swiftUI to re-evaluate, and we lose all drag control.  The `NSHostingView` needs to be stable during the drag gesture, otherwise it will stop working. In other words, the `NSHostingView` needs to be lifted up above the hierarchy in order to be stable  I achieve this stability via two steps:

1.  By moving from passing an `AnyView` to the `TrimmingAVPlayer` to passing a `NSView` directly. This way I can directly compare the `NSView` and see if it is the same as the previous one, so I won't have to remove it from its superview (breaking the drag gesture)
2. I need the overlay to be stable (not change on re-evaluation of the view hierarchy). I do this by lifting up the `EditScreen` view (and storing the actual view in `_EditScreen`). This way when the inner state (`_EditScreen`) changes SwiftUI will not re-evaluate the outer view, and thus the `NSHostingView` will not change.

mmulet:
- [x] After switching to `CoreImage` there was a Colorspace issue (preview was slightly brighter than it should be). This adjusts the colorspace so that it is correct.

sindresorhus:
I also encountered a crash. I cannot consistently reproduce, but I enabled the preview, changed dimensions to 480 or something, and then let it run for a couple of loops, and then pause/play a couple of times, and it crashed:

```
#0	0x00000001008cb894 in std::__1::__hash_table<std::__1::__hash_value_type<long, qos_info_t>, std::__1::__unordered_map_hasher<long, std::__1::__hash_value_type<long, qos_info_t>, std::__1::hash<long>, std::__1::equal_to<long>, true>, std::__1::__unordered_map_equal<long, std::__1::__hash_value_type<long, qos_info_t>, std::__1::equal_to<long>, std::__1::hash<long>, true>, std::__1::allocator<std::__1::__hash_value_type<long, qos_info_t>>>::__emplace_unique_key_args<long, std::__1::piecewise_construct_t const&, std::__1::tuple<long const&>, std::__1::tuple<>> ()
#1	0x00000001008cb42c in qosWaiterSignallerInvariantCheck ()
#2	0x00000001008c5a7c in interposed_dispatch_group_wait ()
#3	0x000000019a92dee4 in invocation function for block in CI::Context::recursive_render(CI::TileTask*, CI::roiKey const&, CI::Node*, bool) ()
#4	0x0000000100ef8514 in _dispatch_call_block_and_release ()
#5	0x0000000100f152dc in _dispatch_client_callout ()
#6	0x0000000100f01594 in _dispatch_lane_serial_drain ()
#7	0x0000000100f0246c in _dispatch_lane_invoke ()
#8	0x0000000100f0ffbc in _dispatch_root_queue_drain_deferred_wlh ()
#9	0x0000000100f0f414 in _dispatch_workloop_worker_thread ()
#10	0x00000001006137a4 in _pthread_wqthread ()
Enqueued from com.apple.root.user-initiated-qos.cooperative (Thread 255) Queue : com.apple.root.user-initiated-qos.cooperative (serial)
#0	0x0000000100efe5ac in dispatch_async ()
#1	0x000000019a92d820 in CI::Context::recursive_render ()
#2	0x000000019a71def0 in CI::Context::render ()
#3	0x000000019a988ccc in invocation function for block in CI::image_render_to_surface(CI::Context*, CI::Image*, CGRect, __IOSurface*, CI::RenderDestination const*) ()
#4	0x000000019a98c888 in CI::recursive_tile ()
#5	0x000000019a9872c4 in CI::tile_node_graph ()
#6	0x000000019a9882bc in CI::image_render_to_surface ()
#7	0x000000019a71807c in -[CIContext(CIRenderDestination) _startTaskToRender:toDestination:forPrepareRender:forClear:error:] ()
#8	0x000000019a717554 in -[CIContext(CIRenderDestination) startTaskToRender:fromRect:toDestination:atPoint:error:] ()
#9	0x000000019a784040 in -[CIContext render:toCVPixelBuffer:bounds:colorSpace:] ()
#10	0x0000000101c5c404 in static PreviewRenderer.renderPreview(previewImage:outputFrame:previewCheckerboardParams:) at /Users/sindresorhus/dev/oss/Gifski/Gifski/Preview/PreviewRenderer.swift:62
#11	0x000000027734f454 in swift::runJobInEstablishedExecutorContext ()
#12	0x00000002773509b4 in swift_job_runImpl ()
#13	0x0000000100f0e30c in _dispatch_root_queue_drain ()
#14	0x0000000100f0ee2c in _dispatch_worker_thread2 ()
#15	0x0000000100613768 in _pthread_wqthread ()
#16	0x000000010061af48 in start_wqthread ()
```

---

Fixture video:

https://github.com/user-attachments/assets/37b826ad-5d57-49f2-92d0-16f4eab358d7

sindresorhus:
I also encountered a crash while it was just playing (size 128). It crashed at: https://github.com/sindresorhus/Gifski/pull/329/files#diff-8d1d6a777c6075916ce9e67873409a0d7dd75e4eb51f1e062e1a89ac2908ec7bR62

```
Thread 9 Queue : com.apple.root.user-initiated-qos.cooperative (concurrent)
#0	0x0000000102ab6f34 in std::__1::__hash_table<std::__1::__hash_value_type<long, qos_info_t>, std::__1::__unordered_map_hasher<long, std::__1::__hash_value_type<long, qos_info_t>, std::__1::hash<long>, std::__1::equal_to<long>, true>, std::__1::__unordered_map_equal<long, std::__1::__hash_value_type<long, qos_info_t>, std::__1::equal_to<long>, std::__1::hash<long>, true>, std::__1::allocator<std::__1::__hash_value_type<long, qos_info_t>>>::find<long> ()
#1	0x0000000102aba880 in findPrimitiveInfoNoAssert ()
#2	0x0000000102ab5a34 in interposed_dispatch_group_wait ()
#3	0x000000019a995150 in CI::RenderTask::waitUntilCompleted ()
#4	0x000000019a723710 in -[CIRenderTask waitUntilCompletedAndReturnError:] ()
#5	0x000000019a78405c in -[CIContext render:toCVPixelBuffer:bounds:colorSpace:] ()
#6	0x0000000103e4c404 in static PreviewRenderer.renderPreview(previewImage:outputFrame:previewCheckerboardParams:) at /Users/sindresorhus/dev/oss/Gifski/Gifski/Preview/PreviewRenderer.swift:62
#7	0x0000000103e4c690 in static PreviewRenderer.renderPreview(previewFrame:outputFrame:previewCheckerboardParams:) at /Users/sindresorhus/dev/oss/Gifski/Gifski/Preview/PreviewRenderer.swift:32
#8	0x0000000103ea993c in PreviewVideoCompositor.render(fullPreviewFrame:originalFrame:outputFrame:compositionTime:) at /Users/sindresorhus/dev/oss/Gifski/Gifski/Preview/PreviewVideoCompositor.swift:97
#9	0x0000000103ea798c in closure #1 in PreviewVideoCompositor.startRequest(_:) at /Users/sindresorhus/dev/oss/Gifski/Gifski/Preview/PreviewVideoCompositor.swift:38
#10	0x0000000103eae8dc in partial apply for closure #1 in PreviewVideoCompositor.startRequest(_:) ()
#11	0x0000000103d848e4 in thunk for @escaping @isolated(any) @callee_guaranteed @async () -> (@out A) ()
#12	0x0000000103e1e558 in partial apply for thunk for @escaping @isolated(any) @callee_guaranteed @async () -> (@out A) ()
```

Gifski/Preview/PreviewRenderer.swift
```diff
@@ -0,0 +1,97 @@
...
+			bounds: outputRect,
+			colorSpace: outputFrame.colorSpace ?? CGColorSpace(name: CGColorSpace.sRGB)
+		)
+	}
+
+	private static func createCheckerboard(
+		outputRect: CGRect,
+		uniforms: PreviewCheckerboardParameters
+	) -> CIImage {
+		guard let filter = CIFilter(name: "CICheckerboardGenerator") else {
```
sindresorhus, line 74: Use `CIFilter.checkerboardGenerator()`. You must import `import CoreImage.CIFilterBuiltins` instead then.

Gifski/Preview/createFullPreviewStream.swift
```diff
@@ -0,0 +1,193 @@
...
+ }
+ ...
+
+ //whenever your settings change and you need to generate a new fullPreview
+ requestNewFullPreview(SettingsForFullPreview(...)
+ ```
+ ## Internals
+ internally the fullPreviewStream works like this: It sets up its own AsyncStream that waits for calls to( RequestNewFullPreview)[RequestNewFullPreview], once there it cancels any existing requests for generation, then spawns a new task to generate the fullPreview.
+ */
+func createFullPreviewStream() -> (AsyncStream<FullPreviewGenerationEvent>, RequestNewFullPreview) {
```
sindresorhus, line 43: All these free function should be under a namespace.

mmulet:
> I'm seeing some preview frames being lighter than others. Notice the grass. Also some quick flicker when showing the bird. This is not visible if I generate the GIF and look at it.
> 
>  Screen.Shot.2025-05-24.at.00.42.51.mov

Looks like the brighter one is incorrect, the switch is exactly when it switches from using prebaked to reading off of the preview `AVAsset` track, so the issue must be with the colorspace of the `AVAsset` track. I verified that it doesn't look bright when I look at the `AVAsset` directly (the `TemporaryURLAVAsset`), so something goes wrong during compositing. I'll track it down.


> I also encountered a crash while it was just playing (size 128). It crashed at: https://github.com/sindresorhus/Gifski/pull/329/files#diff-8d1d6a777c6075916ce9e67873409a0d7dd75e4eb51f1e062e1a89ac2908ec7bR62
> 
> Crash:
> 
> > Thread 9: EXC_BAD_ACCESS (code=1, address=0xf06c8f32f4d8)
> > A bad access to memory terminated the process.
> 
> ```
> Thread 9 Queue : com.apple.root.user-initiated-qos.cooperative (concurrent)
> #0	0x0000000102ab6f34 in std::__1::__hash_table<std::__1::__hash_value_type<long, qos_info_t>, std::__1::__unordered_map_hasher<long, std::__1::__hash_value_type<long, qos_info_t>, std::__1::hash<long>, std::__1::equal_to<long>, true>, std::__1::__unordered_map_equal<long, std::__1::__hash_value_type<long, qos_info_t>, std::__1::equal_to<long>, std::__1::hash<long>, true>, std::__1::allocator<std::__1::__hash_value_type<long, qos_info_t>>>::find<long> ()
> #1	0x0000000102aba880 in findPrimitiveInfoNoAssert ()
> #2	0x0000000102ab5a34 in interposed_dispatch_group_wait ()
> #3	0x000000019a995150 in CI::RenderTask::waitUntilCompleted ()
> #4	0x000000019a723710 in -[CIRenderTask waitUntilCompletedAndReturnError:] ()
> #5	0x000000019a78405c in -[CIContext render:toCVPixelBuffer:bounds:colorSpace:] ()
> #6	0x0000000103e4c404 in static PreviewRenderer.renderPreview(previewImage:outputFrame:previewCheckerboardParams:) at /Users/sindresorhus/dev/oss/Gifski/Gifski/Preview/PreviewRenderer.swift:62
> #7	0x0000000103e4c690 in static PreviewRenderer.renderPreview(previewFrame:outputFrame:previewCheckerboardParams:) at /Users/sindresorhus/dev/oss/Gifski/Gifski/Preview/PreviewRenderer.swift:32
> #8	0x0000000103ea993c in PreviewVideoCompositor.render(fullPreviewFrame:originalFrame:outputFrame:compositionTime:) at /Users/sindresorhus/dev/oss/Gifski/Gifski/Preview/PreviewVideoCompositor.swift:97
> #9	0x0000000103ea798c in closure #1 in PreviewVideoCompositor.startRequest(_:) at /Users/sindresorhus/dev/oss/Gifski/Gifski/Preview/PreviewVideoCompositor.swift:38
> #10	0x0000000103eae8dc in partial apply for closure #1 in PreviewVideoCompositor.startRequest(_:) ()
> #11	0x0000000103d848e4 in thunk for @escaping @isolated(any) @callee_guaranteed @async () -> (@out A) ()
> #12	0x0000000103e1e558 in partial apply for thunk for @escaping @isolated(any) @callee_guaranteed @async () -> (@out A) ()
> ```
> 
> The top frame is:
> 
> Screen Shot 2025-05-24 at 00 55 23

I was able to reproduce this (just once haven't been able to do it again). I will look into it. If it's non-deterministic it's probably a threading issue. I mostly just used your ChatGPT code, but I will read the docs on everything and find what is/isn't thread-safe., and fix.

sindresorhus:
ChatGPT review: https://chatgpt.com/share/6830f704-1a00-8010-a8cc-00f345a26a5c (haven't validated all of these, but worth looking at)

sindresorhus:
Not all good ideas, but using an `actor` may be one: https://chatgpt.com/share/6830f853-a350-8010-8ec0-ad6167bbaaf7

sindresorhus (changes requested):

sindresorhus:
- The crop rect should not be draggable when not in "crop edit mode".

sindresorhus:
I haven't been able to make it crash either.

mmulet:
> The preview can be a bit flaky when I change the size:

From what I can see from the video, the "flakiness" is caused by the preview entering into the regeneration state after the size change. (Ie. it's converting all the GIF frames "on-the-fly" from the video again because the full preview is not available).
 
I'm not sure I can do anything to "fix" this because smooth video playback isn't possible until the full preview is available. 

## Some possible alternatives
- I could add a "loading" indicator watermark, so the user knows that the preview is still generating.
or 
- I could generate a few frames interspersed throughout the video, so the user can see a preview of the video while the full preview is generating. (This might interfere with the ability to click anywhere in the trimmer to seek to that exact position in the preview while the preview is generating. On the other hand, I could add an observer to the player rate and produce different behavior based on whether the player is playing or paused, so it might work out)

mmulet:
> When entering "edit crop mode", the focus is on the toolbar back button, which is not ideal as it could cause the user to accidentally undo their work if they press Space.

I can't reproduce this, it seems to me like when I enter/exit "edit crop mode" the focus remains unchanged. Here I have focus on the trimmer and when I press `space` it just pauses/plays ( tried both clicking on the button on the toolbar, and using the command shortcut)


https://github.com/user-attachments/assets/7eb20e99-5596-459c-9c86-142877ea581c

And here, I have focus on the dimensions, and when I enter/exit "edit crop mode" the focus remains on that field:



https://github.com/user-attachments/assets/3bd4eb58-4b3a-4469-a0a8-26445b0ca876

sindresorhus:
> I can't reproduce this, it seems to me like when I enter/exit "edit crop mode" the focus remains unchanged. Here I have focus on the trimmer and when I press space it just pauses/plays ( tried both clicking on the button on the toolbar, and using the command shortcut)

You need to enable keyboard navigation focus in System Settings => Keyboard

mmulet:
- [x] Stop regen of preview when the crop edit mode is on.
- [x] `LockedCVPixelBuffer` to `Utilities`
- [x] remove outdated comment in `CVPixelBuffer`
- [x] Throw on failure in `LockedCVPixelBuffer`
- [x] With LockedBasedAddress `T?` -> `T`
- [x] throw if buffer.baseAddress is nil in `LockedCVPixelBuffer`
- [x] Comment CreateAVAssetFromGIF settings
- [x] When entering "edit crop mode", the focus is on the toolbar back button, which is not ideal as it could cause the user to accidentally undo their work if they press Space.
 - I should have remembered to `enable keyboard navigation` you told me that before, sorry.

## other issues
- [x] Fix: Preview command wasn't showing up in the top menu.

> Had another crash just now. All I did was enable preview, let it play, and change the dimensions a few times with maybe 10 seconds between.

I will look into these crashes. The second crash is interesting because I've never seen it crash at `static GIFGenerator.convertOneFrame(frame:dimensions:quality:fast:)`, only at `-[CIContext render:toCVPixelBuffer:bounds:colorSpace:] ()`.

sindresorhus:
I got the actor commit to crash too:

```
Thread 39: EXC_BAD_ACCESS (code=1, address=0xf000000001fc6b27)

Thread 39 Queue : com.apple.root.user-initiated-qos.cooperative (concurrent)
#0	0x0000000104bb73ac in std::__1::__hash_table<std::__1::__hash_value_type<long, qos_info_t>, std::__1::__unordered_map_hasher<long, std::__1::__hash_value_type<long, qos_info_t>, std::__1::hash<long>, std::__1::equal_to<long>, true>, std::__1::__unordered_map_equal<long, std::__1::__hash_value_type<long, qos_info_t>, std::__1::equal_to<long>, std::__1::hash<long>, true>, std::__1::allocator<std::__1::__hash_value_type<long, qos_info_t>>>::__do_rehash<true> ()
#1	0x0000000104bbb96c in std::__1::__hash_table<std::__1::__hash_value_type<long, qos_info_t>, std::__1::__unordered_map_hasher<long, std::__1::__hash_value_type<long, qos_info_t>, std::__1::hash<long>, std::__1::equal_to<long>, true>, std::__1::__unordered_map_equal<long, std::__1::__hash_value_type<long, qos_info_t>, std::__1::equal_to<long>, std::__1::hash<long>, true>, std::__1::allocator<std::__1::__hash_value_type<long, qos_info_t>>>::__emplace_unique_key_args<long, std::__1::piecewise_construct_t const&, std::__1::tuple<long const&>, std::__1::tuple<>> ()
#2	0x0000000104bba788 in createPrimitiveEntry ()
#3	0x0000000104bb5644 in interposed_dispatch_semaphore_create ()
#4	0x00000001b034ca4c in IOGPUNotificationQueueCreate ()
#5	0x00000001b0356f0c in IOGPUCommandQueueCreateWithQoS ()
#6	0x00000001b034367c in -[IOGPUMetalCommandQueue initWithDevice:descriptor:] ()
#7	0x0000000117ddc81c in -[AGXG16XFamilyCommandQueue initWithDevice:descriptor:] ()
#8	0x0000000117e0bae0 in -[AGXG16XFamilyDevice newCommandQueue] ()
#9	0x000000018eea9b9c in -[MTLDebugDevice newCommandQueue] ()
#10	0x000000019a70c3d4 in CIMetalCommandQueueCreate ()
#11	0x000000019a70c358 in +[CIContext(Internal) internalContextWithMTLDevice:options:] ()
#12	0x000000019a70c02c in -[CIContext initWithOptions:] ()
#13	0x0000000105fa1560 in @nonobjc CIContext.init() ()
#14	0x0000000105fa0ce4 in CIContext.__allocating_init() ()
#15	0x0000000105fa0534 in PreviewRenderer.renderPreview(previewImage:outputFrame:previewCheckerboardParams:) at /Users/sindresorhus/dev/oss/Gifski/Gifski/Preview/PreviewRenderer.swift:74
#16	0x0000000105f9fd44 in PreviewRenderer.renderPreview(previewFrame:outputFrame:previewCheckerboardParams:) at /Users/sindresorhus/dev/oss/Gifski/Gifski/Preview/PreviewRenderer.swift:50
#17	0x0000000105f9f764 in static PreviewRenderer.renderPreview(previewFrame:outputFrame:previewCheckerboardParams:) at /Users/sindresorhus/dev/oss/Gifski/Gifski/Preview/PreviewRenderer.swift:25
#18	0x0000000105fd1468 in PreviewVideoCompositor.RenderPreview.render(state:fullPreviewFrame:originalFrame:outputFrame:compositionTime:) at /Users/sindresorhus/dev/oss/Gifski/Gifski/Preview/PreviewVideoCompositor.swift:138
#19	0x0000000105fcf248 in closure #1 in PreviewVideoCompositor.startRequest(_:) at /Users/sindresorhus/dev/oss/Gifski/Gifski/Preview/PreviewVideoCompositor.swift:51
#20	0x0000000105fd77ac in partial apply for closure #1 in PreviewVideoCompositor.startRequest(_:) ()
#21	0x0000000105ea89f0 in thunk for @escaping @isolated(any) @callee_guaranteed @async () -> (@out A) ()
#22	0x0000000105f6fc30 in partial apply for thunk for @escaping @isolated(any) @callee_guaranteed @async () -> (@out A) ()
```

Gifski/Preview/createAVAssetFromGIF.swift
```diff
@@ -0,0 +1,175 @@
...
+			} catch {
+				sendableWriterInput.value.markAsFinished()
+				continuation.resume(throwing: error)
+			}
+		}
+	}
+	await assetWriter.finishWriting()
+	guard assetWriter.status != .failed else {
+		throw CreateAVAssetError.failedToWrite
+	}
```
sindresorhus, line 122: Shouldn't this be:

```suggestion
	guard assetWriter.status == .completed else {
		throw (assetWriter.status.error ?? CreateAVAssetError.failedToWrite)
	}
```

currently it ignores `.cancelled` and `.unknown`.

Gifski/Components/TrimmingAVPlayer.swift
```diff
@@ -46,20 +71,23 @@ final class TrimmingAVPlayerViewController: NSViewController {
...
+			guard oldValue != overlay else {
+				return
+			}
+			if let oldValue {
+				oldValue.removeFromSuperview()
 			}
 
 			guard let overlay else {
-				underTrimOverlayView = nil
+//				underTrimOverlayView = nil
```
sindresorhus, line 85: Leftover?

Gifski/Preview/PreBakedFrames.swift
```diff
@@ -0,0 +1,86 @@
...
+		let numberOfFramesToPrecomputeAtTheEnd = min(
+			numberOfFrames,
+			Int(ceil(offsets.end.seconds * frameRate * speed)) + 1
+		)
+
+		let beginningRange = Set(0..<numberOfFramesToPrecomputeAtTheBeginning)
+		let endRange = Set(numberOfFrames - numberOfFramesToPrecomputeAtTheEnd..<numberOfFrames)
+		let allFrames = beginningRange.union(endRange)
+
+		let dict = try await allFrames.asyncMap { frameIndex in
```
sindresorhus, line 41: There is no benefit of using asyncMap here. It's not concurrent, it just allows waiting for asynchronous operations. Also, concurrency may be problematic as I'm not sure if CGImageSource is thread-safe.

```suggestion
		let dict = try await allFrames.map { frameIndex in
```

Gifski/Preview/PreBakedFrames.swift
```diff
@@ -0,0 +1,86 @@
...
+			guard let image = CGImageSourceCreateImageAtIndex(imageSource, frameIndex, nil) else {
```
sindresorhus, line 42: Use `imageSource.createImage`

Gifski/Preview/PreBakedFrames.swift
sindresorhus: Could maybe also use some cancellation support:

```swift
try Task.checkCancellation()
```

sindresorhus:
- I feel like the preview feature is still very complicated. I would love to see it refactored/simplified a bit. If there is a bug in the future, I'm honestly not sure where to start debugging. There are so many moving pieces. And I do realize it's a complicated task, but I think it's worth a shot.
- Make sure all the Metal features work on all devices and do fallbacks if not: `MTLDevice.supportsFeatureSet(_:)` For example, I believe `setMeshBytes` does not work on Intel Macs.
- Feed the files in the "Preview" folder to ChatGPT with this prompt (if you have it available, I would recommend the o4-mini-high model): 
  > Find real bugs and meaningful improvements. Ignore style and APIs you are not aware of. No nitpicking. Focus on logic flaws, missed edge cases, performance issues, unnecessary complexity, or cleaner structure. Be exhaustive.

  - For example: https://chatgpt.com/share/6839c199-39ec-8010-a4cb-5a97ff134bf7
It can also help to use the same prompt with multiple AIs, like Grok or Gemini: https://grok.com/share/bGVnYWN5_de61b1ee-4110-43ca-a980-a82d7d456845
And when you're done with the list, you can just ask "give me more".

sindresorhus (changes requested):

mmulet:
- [x] Radical simplification, see below
- [x] Set metal shader language version to 2.0, the old version supported by intel Macs. I don't have an Intel mac to test this on, however. I do use one advanced feature `astc textures`, but I put that behind a feature flag () with a fallback for older machines. (more about that in the simplification section)
- [x] Passes Swift 6 concurrency checks.
> Find real bugs and meaningful improvements. Ignore style and APIs you are not aware of. No nitpicking. Focus on logic flaws, missed edge cases, performance issues, unnecessary complexity, or cleaner structure. Be exhaustive.
- [x] I have github copilot, so it has github copilot versions of the models, but I did this with:
  - openai gpt-4o
  - Gemini 2.5 Pro
  - Claude Sonnet 4
- [x] Pause preview when generating and try to get that single frame as fast as possible. 
  - The only thing I can do to make it faster now would be to render at a smaller resolution, but that might not reflect the quality of the video.
  - Although if you want, I could do something like [progressive image rendering](https://jmperezperez.com/cssconfau16/assets/progressive-jpeg-vs-baseline-jpg.jpg) where I very quickly convert a low res preview and then progressively increase the resolution as the frames are generated.


## Radical simplification
The gist is that I removed the second track from the `PreviewableComposition`, and instead I pre-render the gif as textures. I didn't do this at first because I thought the texture size would be too big to be practicable. It turns out, that the the macs have a built in runtime support and encoder for [astc textures](https://www.khronos.org/opengl/wiki/ASTC_Texture_Compression), which allows compression down to about 2.00 bits per pixel. This makes the approach feasible. By using these compressed textures, I can avoid almost all of the weird `AVPlayer` quirks. This simplifies the code a lot.

Then, I went through each and every line of code and removed unnecessary parts, moving everything that could, made everything as reusable and generic as possible etc. Each file (other than Utilities) is now about 150 lines or less, and it is meant to be easy to read and understand. 

Let's go through a summary of the major changes:

1. `createFullPreviewStream` -> `FullPreviewStream` which handles creation of the `FullPreview`, the entire stream is now an `actor` which makes the state changes easy to reason about. Same as before only one generation will be in progress at a time, and each generation will be canceled if a new one is requested. 

The actual conversion of the video to a `FullPreview` is simple, first convert to GIF, `GIFGenerator.runProgressable` then convert the animated GIF data to the textures one per frame (`createAVAssetFromGIF` is gone). Both tasks are done using the new `ProgressableTask` which is simply a `Task` that can report progress, along with a way to compose two together as just one task with one progress stream.

2. `FullPreviewGenerationEvent` is the event emitted by the `FullPreviewStream`. It is now even simpler, and avoids unnecessary state like `FullPreviewStatus` because it is now directly `Sendable`. 

Just as before, `EditScreen` subscribes the `FullPreviewStream` and updates the `PreviewableComposition` with the new `FullPreview` when it is available, now the logic is even simpler:
```swift
.task {
	for await event in fullPreviewStream.eventStream {
			fullPreviewState = event
	}
}
```

This new state hides all its internals, but it still can be used to display the proper UI (`canShowPreview`, `progress`, `isGenerating`, etc.) and to get frames for drawing with `getPreviewFrame`

3. This brings us to the `PreviewableComposition`. It has been heavily simplified:
- removes the preview track
- no more need to `updateFullPreviewTrack` and no (`AVPlayer` quirks)

It just uses the standard code for adding a video compositor.

4. After we have made our `PreviewComposition` we input into the `TrimmingAVPlayer` as before, but with handy helper functions like `forceAVPlayerToRedraw` and `updatePreviewState`, which do exactly what they say.

5. Now we go to `PreviewVideoCompositor` which has been heavily simplified as well. Gone are the `OutputCache`, the `PreBakedFrames`. Render logic is now simple: 
- Show preview y/n
- if no, just render the original video frame
- if yes, render the preview frame, get it from the state

6. The functionality of `PreviewRenderer` is mostly the same (draw the preview, with checkerboard pattern in the background). I did move to using Metal 2.0, the old version still compatible with Intel Macs. Plus, I moved it to an actor to future proof it for `Swift 6`. Despite it's similar function, I did completely refactor it to make the purpose of each part clear.
7. 
Still, here is a rundown of the `PreviewRenderer`:

- Everything used to initialize the graphics context, is all put in `PreviewRendererContext`, this handles the initialization of all shared resource that will be used to render each and every frame.
- Everything related to creating the textures is in `SendableTexture`, everything there is isolated to the `PreviewRenderer` actor and compiles with `Swift 6` concurrency checks.
 - This handle conversion from `CGImage` and `Data` to `MTLTexture` as well as compression to `astc` textures.
- `PreviewRenderer` itself is only concerned about the actual rendering of the frames. 

Since you said you are unfamiliar with metal, I commented almost every line of code in  `PreviewRenderer.swift`, `PreviewRendererContext.swift`, and `compositePreview.metal`. This should really help make it maintainable in the future.

## That's it!

mmulet:
## TODO
- [ ] bug: synchronize preview with video after speed changes.

sindresorhus:
Thanks for working more on this. The new version looks really good! It seems to run great and I was not able to find any bugs or crashes.

sindresorhus:
> Set metal shader language version to 2.0, the old version supported by intel Macs.

I think Intel Macs support the latest Metal version, just not all the features, so better to set it to latest and just feature detect for unsupported features.

mmulet:
## Changes
- [x] feature check metal features
 - `MTLDevice.supportsFeatureSet` is deprecated, but we do have `MTLDevice.supportFamily`, and I do use that to check. I went through [Metal Feature Set Tables ](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) manually, and I asked A.I., and we both agree that checks are correct. Everything we do except the `astc` compression is supported by the `common1` family. Then, we also check for `astc` support, which is only available on `apple2` and later, with a fallback for when it is not available.
- [x] `fieldFont` func -> `fieldFont` getter
- [x] move `CropSettings` out of `Utilities`
- [x] `MTLPixelASTCCompressionError` -> `ASTCCompressionError`
- [x] doc comment `astc_ldr`
- [x] `SendableTexture` don't capture self
 - The strong reference to self was necessary for actor isolation. The only way I could get around it was to not isolate the `init` for `SendableTexture`, but I made it `fileprivate` so we can manually check that each call to `init` is isolated.
- [x] `CVPixelBufferError`, `CGImageSourceError` -> `Error`, 
- [x] `CGImageSource.create`: `CFDictionary` -> `Dictionary`
- [x] `CGImageDestination.add(image:...` -> `CGImageDestination.addImage(...`
- [x] `Data.convertToTexture` -> `Data.convertToTexture()`
- [x] doc `Int.aligned`
 - Actually removed `Int.aligned` because if you calculated the `bytesPerRow` of the ASTC image using `blocksPerRow` you don't have to align it because it already a multiple of `ASTCBlockSize`.
- [x] Add specific error codes for each `CVReturn` error in `CVMetalTextureCache`
- [x] `CGPoint.clamped(within:)` -> `CGPoint.nearestPointInsideRectBounds(_:)`
- [x] `MTLCommandBuffer.commit` with weak self
- [x] `CVPixelBuffer.convertToGIF(...)` out of `Utilities.swift`
- [x] `GifGenerator` extension in `Utilities.swift` to `GIFGenerator.swift`
- [x] reconfigure `ASTCImage.bytesPerRow`
- [x] `renderEncoder.endEncoding()` -> `defer { renderEncoder.endEncoding() }`

sindresorhus:
Sorry about the delay. Been traveling.

mmulet:
## Changes
- [x] `PreviewRenderer.shared` will now try to create a new `PreviewRenderer` if it has failed before.
- [x] Typed throws wherever possible
 - Honestly, mixed feelings about this. It makes some of the code more verbose, and most of the time we interact with apple provided APIs that throw untyped errors. So the entire functions error are still untyped. Furthermore, swift has no way to do union of errors other than the completely type erased `any Error` or creating an `enum` of the type (unless I missed something at WWDC last week), so we end up with typed errors like `BaseAddressLockedError<LockedPlanesError<BaseAddressLockedError<LockedPlanesError<CopyError>>>>`.
 
 

> Sorry about the delay. Been traveling.

No worries. 

------
### UI Changes
This might still be a bit too early for this, but I was watching some of the WWDC videos last week, and as I'm sure you know, they are moving to the new "liquid glass" look. Do you think there is anything in GIFski preview? Beta 1 is usually a bit too buggy, so I haven't tried it out yet. But from the video, it looks like they are revamping [the navigation bar](https://developer.apple.com/videos/play/wwdc2025/310/?time=95) among other changes.

Gifski/Preview/FullPreviewGenerationEvent.swift
```diff
@@ -154,9 +158,6 @@ struct FullPreviewGenerationEvent: Equatable, Sendable, PreviewComparable {
 			reportedCompositionTime.seconds * settings.speed
 		}
 
-		enum Error: Swift.Error {
```
sindresorhus: It can still be called `Error`. It's scoping.

Gifski/Preview/FullPreviewGenerationEvent.swift
```diff
@@ -130,13 +130,17 @@ struct FullPreviewGenerationEvent: Equatable, Sendable, PreviewComparable {
 		let settings: SettingsForFullPreview
 		let gifData: [SendableTexture?]
 
-		func getGIF(at compositionTime: CMTime) throws -> SendableTexture {
+		func getGIF(at compositionTime: CMTime) throws(GetGIFError) -> SendableTexture {
 			guard let image = gifData[getCurrentGIFIndex(at: compositionTime)] else {
-				throw Error.failedToGetGIFFrame
+				throw GetGIFError.failedToGetGIFFrame
```
sindresorhus, line 135:
```
throw .failedToGetGIFFrame
```

Gifski/Preview/PreviewRenderer.swift
```diff
@@ -25,12 +28,14 @@ actor PreviewRenderer {
 		previewFrame: SendableTexture,
 		outputFrame: SendableCVPixelBuffer,
 		fragmentUniforms: CompositePreviewFragmentUniforms
-	) async throws {
+	) async throws(MTLCommandQueueRenderError<PreviewRenderer.RenderPreviewError>) {
```
sindresorhus, line 31: My fault, I should have been clearer. I meant to only use typed throws when the method throws a single typed error. It doesn't make sense when there are multiple error and wrapping is required.

It's not a huge benefit with typed throw, but for small methods that throw a single error, the typed-ness can be used as documentation and it also lets us use the `throw .errorName` syntax.

Gifski/Preview/PreviewRendererContext.swift
```diff
@@ -92,36 +101,48 @@ struct PreviewRendererContext {
 		samplerDescriptor.tAddressMode = .clampToEdge
 
 		guard let samplerState = metalDevice.makeSamplerState(descriptor: samplerDescriptor) else {
-			throw PreviewRenderer.Error.failedToMakeSampler
+			throw SetupSamplerStateError.failedToMakeSampler
```
sindresorhus, line 104: In cases like this, it was better to use `PreviewRenderer.Error.failedToMakeSampler`

sindresorhus (changes requested):

mmulet:
- [x] Remove wrapped Errors
 - [x] Use `Error.error` -> `.error` when appropriate

-----
I removed the complicated typed error handling, I like it much better now 👍

sindresorhus:
> This might still be a bit too early for this, but I was watching some of the WWDC videos last week, and as I'm sure you know, they are moving to the new "liquid glass" look. Do you think there is anything in GIFski preview? Beta 1 is usually a bit too buggy, so I haven't tried it out yet. But from the video, it looks like they are revamping [the navigation bar](https://developer.apple.com/videos/play/wwdc2025/310/?time=95) among other changes.

I haven't tried it yet, but it's likely some UI tweaks are needed. Although we get a lot for free by using SwiftUI. Let's continue in: https://github.com/sindresorhus/Gifski/issues/336

sindresorhus:
If you're interested in tackling other issues, I'm happy to add bounties to them. No pressure though, I totally get it if you're feeling burnt out. Cropping and preview ended up being way more challenging than I expected.

---

In case you are interested, I think the priority now is:
- https://github.com/sindresorhus/Gifski/issues/332
- https://github.com/sindresorhus/Gifski/issues/337
- https://github.com/sindresorhus/Gifski/issues/336
- https://github.com/sindresorhus/Gifski/issues/104

mmulet:
> If you're interested in tackling other issues, I'm happy to add bounties to them.

Yeah, let's do it! Obviously, I'm not going to get rich doing this (I probably made less than $1/hr on these 2 issues), but bounties really help motivate me to see the issues through completion (even if they take 3 months longer than anticipated 😅). 
(Also, if you ever feel like bumping up the size of the bounties on the longer ones, I wouldn’t complain).

> No pressure though, I totally get it if you're feeling burnt out. Cropping and preview ended up being way more challenging than I expected.

I'm not feeling burnt out at all. Let's go!

> In case you are interested, I think the priority now is:

Definitely interested.

Honestly, the hardest part of bug/issue bounty hunting is just finding issues that people are willing to fund and actually respond to. So many PRs end up ignored or forgotten. You’ve been great to work with so far, so keep them coming.

If you’ve got issues on other repos you’re funding, I’m happy to check those out too.

sindresorhus:
Note that I may be slower at replying as I have more limited time for work in July (visiting parents).

mmulet:
> Note that I may be slower at replying as I have more limited time for work in July (visiting parents).

No worries.



> There is also some crash happening on macOS 26 likely introduced with this pull request. Probably not worth looking into yet, but could be an issue if it still happens in later betas.
> 
> [2025-06-30_23-05-11.1807_+0400-6e0d7e6585177d4fe79fb2e1a2ab10e39987becf.crash.zip](https://github.com/user-attachments/files/20996554/2025-06-30_23-05-11.1807_%2B0400-6e0d7e6585177d4fe79fb2e1a2ab10e39987becf.crash.zip)

Looks like it crashes on this line in `CropOverlayView.swift`

```swift
	window?.isMovableByWindowBackground = windowIsMovable
```
Perhaps in the new OS, the old window becomes a null reference as soon as the newWindow is available? If this is still a problem in later betas, worse case scenario we can always workaround it by changing it back to 

```swift
.onAppear {
  let windowIsMovable = SSApp.swiftUIMainWindow?.isMovableByWindowBackground
  SSApp.swiftUIMainWindow?.isMovableByWindowBackground = false
}
```
But hopefully there will be something we can do to keep `.bindHostingWindow` so we can keep everything pure SwiftUI.

Issue #180, opened 2020-02-09
Quick Action


sindresorhus:
Since macOS 10.14, we can now [make Quick Actions that show up in Finder](https://developer.apple.com/documentation/macos_release_notes/macos_mojave_10_14_release_notes/appkit_release_notes_for_macos_10_14#3044962). It would be cool to have a GIF quick action that just quickly converts a GIF without any UI, just with some sensible defaults the user could change in a preferences window. This would also allow for batch conversion.

mmulet:
Solved by pull request #330

PR #330, opened 2025-03-24
Implemented the Quick Action feature from issue  #180

mmulet:
Hi!

I implemented the Quick Action feature from issue #180
This time I have 2 videos for you!
This one shows how you can use the quick action:
1. Just select the video, select Quick action -> Quick Convert to GIF with Gifski
2. Then it quickly converts the GIF!
3. The gif is saved in to Downloads folder (the easiest place to save from an App Sandbox perspective), and the Quick Action opens Finder to the file after it creates it.

https://github.com/user-attachments/assets/e5c27662-30a9-440b-975a-02cc48482d27

The second video shows how you can change the sensible defaults for quick conversion.
Just open settings from the menu bar or Cmd+, and then edit the settings using
the same UI as you use in the normal app.
https://github.com/user-attachments/assets/e84b0ba4-2206-48d1-8951-5fe5955e2ba1

I believe this should be exactly what you wanted!

Once again I will submit this to issueHunt for the ongoing bounty, but any extra tips (you


mmulet:
> I've been thinking more about this, and I'm not sure it actually makes sense to have a single set of settings that always apply. I know I originally suggested that in the issue, but I hadn't fully thought it through.
> 
> The optimal GIF settings really depend on the kind of video you're converting, and there can be many different use-cases. Showing a small settings window each time could help, but then users might also expect a preview, and at that point, we’re basically back at the main app UI.
> 
> Now that we support Shortcuts, I’m wondering if a better approach is to let users create their own shortcuts with their preferred settings. These can also show up in the Quick Actions menu, which keeps things both flexible and efficient.
> 
> Thoughts?

Yeah, I like this new idea better too. The shortcuts app is basically a microcosm of the settings anyways.
I gave it a try and it works well:

https://github.com/user-attachments/assets/10ff2b5f-984b-4f23-a637-4189358e7a65

sindresorhus:
Open a new pull request with a short tip about this and I will award you the Quick Action bounty. You can add a "Tips" section above https://github.com/sindresorhus/Gifski?tab=readme-ov-file#faq

Issue #64, opened 2019-03-14
Green line artifact on right side of gif


brumm:
[<img alt="Issuehunt badges" src="https://issuehunt.io/r/sindresorhus/gifski-app/issues/64/badge.svg" />](https://issuehunt.io/r/sindresorhus/gifski-app/issues/64)

# The bounty for this issue now applies to https://github.com/sindresorhus/Gifski/issues/183 instead.

---

Hey,

I'm seeing green artifacts on the right side of the resulting gif. The input video file is taken straight from the Mojave screen recording tool.

Here's a sample:

![Screen Recording 2019-03-14 at 15 53 36](https://user-images.githubusercontent.com/170500/54367573-e1cc7600-4672-11e9-8d97-e3eabfd512e2.gif)

---

**Note:** This is not an easy issue and requires good Swift and macOS/iOS experience.

> There is a $60.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/gifski-app/issues/64).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/gifski-app/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

sindresorhus:
Could you share the original recording? (Just ZIP it and attach it here) So I can reproduce it locally.

brumm:
Sure thing!

[Screen Recording 2019-03-14 at 15.53.36.mov.zip](https://github.com/sindresorhus/gifski-app/files/3002645/Screen.Recording.2019-03-14.at.15.53.36.mov.zip)

sindresorhus:
I can reproduce the problem when converting the video to PNG frames using ffmpeg too:

```
$ ffmpeg -i "Screen Recording 2019-03-14 at 15.53.36.mov" frame%04d.png
```

This is what I end up with:

![frame0001](https://user-images.githubusercontent.com/170270/54913098-993f7300-4f24-11e9-9006-1841184e6938.png)

sindresorhus:
I think it's because the dimensions of the video is 384x529, which is a problem, as both sides have to be divisible by 4.

@kornelski Any thoughts what we could do about this?

kornelski:
If AVFoundation and ffmpeg both agree that there's a green line there, there's a green line there. 

[...]

PR #331, opened 2025-03-25
Implemented Crop video issue #219

### LINKED ISSUE #219: Crop video
[Issuehunt badges](https://github.com/sindresorhus/Gifski/issues/64)

**The bounty refers to a different issue, but it applies here.**

It could sometimes be useful to crop the video.

If we do this, it should be fully made in SwiftUI.

---

This issue requires advanced Swift and SwiftUI knowledge. It's not an easy issue.

mmulet:
Hi again!
I thought I might as well go for 3/3, so I solved another issue ( #219 to be exact)!

Fixes #219 

As usual, watch this video to get an overview of the new feature:

https://github.com/user-attachments/assets/8e239920-3a01-46cc-a6c7-cb872c8e36dc

Here is an overview of the changes:
- [x] To the dimensions settings adds a `Crop Toggle` which turns on/off output cropping.
- [x] Adds the Edit Crop button, which if you dare press it, brings you to the edit crop screen which lets you, edit the crop.
- [x] As the video loops in the background, you can move the crop around, resize it, rescale it, or just use some predefined presets.
- [x] Naturally, the new edit screen is done in all SwiftUI, just as you like it.
- [x] When the crop is active you can view it in main editing screen (so you can easily see that it will be cropped)
- [x] Finally, you can actually export the cropped video.

This should be everything you asked for in the issue!

By now this should come as no surprise, but I will submit this to issueHunt for the ongoing bounty, but any extra tips (you

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#64: Green line artifact on right side of gif](https://oss.issuehunt.io/repos/119822304/issues/64)
---
</details>


sindresorhus:
Thanks for working on this 🙏 I think it would have been smart to discuss how it should work before implementing it though.

I think the best way to implement cropping is to do it right in the editor screen, not a separate screen. For example, the user may need to scrub the video position while deciding how to crop it. The only problem I see with doing it this way is that the trimmer control (the yellow control) is above the video, but maybe you could find a way to make the video not go under the trimmer.

It would be more natural to put a button for enabling crop editing in the top-right toolbar. And it doesn't need both a switch and a "edit crop" button. It's enough with just a toggle to enter crop mode (which shows the cropper), and maybe also puts a reset crop button in the toolbar, when in crop mode.

We don't need crop presets, so you can drop that. What we may need is a way to lock the cropper to certain ratios. For example, the user may want exactly 1:1 ratio or 16:9 ratio.

The tips can simply be a one time popover.

The cropper itself is a bit buggy:
- If I drag the top-right handle to the top-right, then continue dragging the mouse out, and then drag it back, it should not move the cropper until the mouse enters the video again.
- The drag handles are a bit too big.
- The cropper needs some subtle shadow.
- It should not be possible to drag the cropper to be zero size. There should be a minimum, like 20x20.
- The cropper should look more native. Take a look at the cropper in the markup tools if you Quick Look an image in Finder.

Gifski/Constants.swift
```diff
@@ -1,25 +1,43 @@
...
+typealias CropRect = CGRect
+
 extension Defaults.Keys {
 	static let outputQuality = Key<Double>("outputQuality", default: 1)
 	static let outputSpeed = Key<Double>("outputSpeed", default: 1)
 	static let outputFPS = Key<Int>("outputFPS", default: 10)
 	static let loopGIF = Key<Bool>("loopGif", default: true)
 	static let bounceGIF = Key<Bool>("bounceGif", default: false)
+	static let outputCrop = Key<Bool>("outputCrop", default: false)
+	static let outputCropRect: Key<CropRect> = .init("outputCropRect", default: .initialCropRect)
```
sindresorhus, line 32: This should just be state and not saved to userdefaults.

Gifski/EditScreen.swift
```diff
@@ -391,11 +420,11 @@ private struct DimensionsSetting: View {
...
 			return CGSize(width: width, height: height).rounded()
 		}
-		.filter { $0.width <= videoDimensions.width && $0.height <= videoDimensions.height }
+			.filter { $0.width <= videoDimensions.width && $0.height <= videoDimensions.height }
 
 		let predefinedPixelDimensions = pixelDimensions
-			// TODO
-//			.filter { resizableDimensions.validate(newSize: $0) }
+		// TODO
+		//			.filter { resizableDimensions.validate(newSize: $0) }
```
sindresorhus, line 427: Don't make unrelated changes.

sindresorhus (changes requested):

mmulet:
## Requested Changes

Here is what I've got, let me know if I am missing anything or there is something else you want to add.

- [ ]  cropping right in the editor screen
- [ ] make video not go under the trimmer
- [ ] button for enabling crop editing in the top-right toolbar
- [ ] Lock crop to ratios (like 1:1 or 16:9) instead of presets
- [ ] Fix:  drag the top-right handle to the top-right, then continue dragging the mouse out, and then drag it back, it should not move the cropper until the mouse enters the video again.
- [ ] The drag handles are a bit too big. (this is a one line fix)
- [ ] Add shadow to cropper
- [ ] Min size for the cropper
- [ ] Style after the cropper in the markup tools if you Quick Look an image in Finder.
- [ ] Remove crop from defaults (This should just be state)
- [ ] Fix style changes

Gifski/EditScreen.swift
mmulet: Sorry about that, I must have accidentally done some formatting and forgot to revert. Should be really easy to fix, I'll fix that real quick and push the fixed version. Should make the pull request easier to review.

mmulet: Found the issue, vscode was hiding the formatting diffs because of this setting

settings.json:
```json
{
  "diffEditor.ignoreTrimWhitespace": true
}
```
So no future commits won't have this problem.

mmulet:
Alright, I fixed the things!:

https://github.com/user-attachments/assets/28d07cf9-f941-40e3-a288-a8c7340cbfea

The full list of changes are in my last comment, but here is an overview:

Everything is now in the main screen, just click on the crop button in the tool bar to bring up the crop editor. Speaking of which, you'll notice that it now mimics the cropper in the markup tools of Quick Look in Finder (not shown in the video, but it also mimics the behavior of shift drag and option drag from Quick Look). Then, see how you can still edit the time with the trimmer while cropping, but if the trimmer gets in the way, you can toggle it on or off.
Lastly, you can quickly choose an aspect ratio from the menu.

Let's discuss the aspect ratio:
I experimented with snapping the crop to specific aspect ratios, but it really got in the way when adjusting the crop. For example, when you edit the crop you typically only edit one side. When it would snap to the aspect ratio, the side that you were not dragging would also change. This isn't what I would expect when dragging just one side. Then, if this is not what I wanted, I would have to adjust all four sides of the crop again.

If you really want an interactive aspect ratio snap, maybe we could make it another mode? Perhaps in that mode the crop is always centered in the center of the video, and the corner of the crop would follow the mouse, snapping to the nearest known aspect ratio. Do you think this is necessary or would the menu suffice?

Gifski/Constants.swift
```diff
@@ -1,12 +1,27 @@
...
+}
+/// Codable Rect for Cropping
+/// The coordinates are as follows:
+/// - The origin in the upper left hand corner
+///  - x is in [0,1]
+///  - y is in [0,1]
+///  It's just like a texCoords in a UV texture
+///  in common graphics subroutines.
+typealias CropRect = CGRect
+
```
sindresorhus, line 24: This is moot now.

Gifski/CropOverlayView.swift
```diff
@@ -0,0 +1,838 @@
...
+			let windowIsMovable = SSApp.swiftUIMainWindow?.isMovableByWindowBackground
+			SSApp.swiftUIMainWindow?.isMovableByWindowBackground = false
+			Task {
+				@MainActor in
+				self.windowIsMovable = windowIsMovable
+			}
+			if let flagsMonitor {
+				NSEvent.removeMonitor(flagsMonitor)
+			}
+			flagsMonitor = NSEvent.addLocalMonitorForEvents(matching: .flagsChanged) { event in
```
sindresorhus, line 130: See `View#onModifierKeysChanged`

Gifski/CropOverlayView.swift
```diff
@@ -0,0 +1,838 @@
...
+				}
+			}
+		}
+		.onAppear {
+			/// Stop the entire window from dragging
+			/// when we drag on the the crop mask
+			/// I would love a pure SwitUI way to do
+			/// this, do you know of one?
+			let windowIsMovable = SSApp.swiftUIMainWindow?.isMovableByWindowBackground
+			SSApp.swiftUIMainWindow?.isMovableByWindowBackground = false
```
sindresorhus, line 122: Maybe: https://developer.apple.com/documentation/swiftui/windowdraggesture#

Gifski/EditScreen.swift
```diff
@@ -49,6 +55,45 @@ struct EditScreen: View {
 		.background(.ultraThickMaterial)
 		.navigationTitle(url.lastPathComponent)
 		.navigationDocument(url)
+		.toolbar {
+			ToolbarItem(placement: .automatic) {
```
sindresorhus, line 59: `automatic` is default. And you can put them all in a `ToolbarItemGroup`.

Gifski/EditScreen.swift
```diff
@@ -49,6 +55,45 @@ struct EditScreen: View {
...
+				Button(action: {
```
sindresorhus, line 60: Should be a `Toggle`.

sindresorhus (changes requested):

Gifski/Constants.swift
mmulet:
```suggestion
/// Rect for Cropping
/// The coordinates are as follows:
/// - The origin in the upper left hand corner
///  - x is in [0,1]
///  - y is in [0,1]
///  It's just like a texCoords in a UV texture
///  in common graphics subroutines.
typealias CropRect = CGRect

```
Good catch. We could keep the typealias just to denote that the bounds are expected to be between 0-1

sindresorhus: Sounds like you may want this type instead then: https://developer.apple.com/documentation/swiftui/unitpoint#

mmulet:
## changes

- [ ] CropRect -> a pair of UnitPoint s
- [ ] View#onModifierKeysChanged
- [ ] Try using windowDragGesture
- [ ] ToolbarItemGroup
- [ ] Toolbar button to toggle
- [ ] The crop button at the top-right should stay at the top right, even when activated.
Activating the cropper should not change the toolbar height or change the position of any other UI element.
- [ ] The aspect ratio thing should be a Picker, not a Menu. And it should show the selected aspect ratio. By default, it would be "None". It would also be useful if it had a "Custom..." entry at the bottom (with Divider between) that would let users input a custom ratio. It should also have an entry for the video aspect ratio and with a " (current)" after it. If the video aspect ratio is already one from the list, that enrtry can get " (current) instead.
- [ ] "Reset" button should be disabled when there is nothing to reset.
- [ ] The crop handles at the edges look a bit weird and not like in Markup. And they have the wrong color in light mode.
- [ ] I think we should add an entry for activating crop mode to the menu bar (SwiftUI.Commands) with a shortcut, so users can easily trigger it.
- [ ] Would be nice to show a popover from the crop button the first time it's enabled with the tips about holding Shift and Option.
- [ ] It would be nice if I could hold both Shift and Option to get both aspect ratio lock and resize from center.
- [ ]  When trying to drag the top-left corner to resize the crop rect, it's too easy accidentally moving it instead.
- [ ] Also, the resize triggers not just at the top-left, but also if you move the mouse further inside the rectangle. It should not do that.
- [ ] Having to manually toggle the trimmer is not very elegant. It would be better to find a way to move it to outside the video view. Probably with some view hacking, like done in TrimmingAVPlayer.hideTrimButtons.
- [ ] Refactor to use macOS15 apis

mmulet:
> The crop handles at the edges look a bit weird and not like in Markup.

I think I found the source of the weirdness:

Here is the border in Quick Look: 

Screenshot 2025-04-10 at 3 22 38 PM

And here is the handle in Gifski:

Screenshot 2025-04-10 at 3 27 21 PM

The border needs to extend past the handle, and the mustache has to be one pixel shorter, should be an easy fix.

mmulet:
Uploaded the latest changes, vids:


https://github.com/user-attachments/assets/899b2128-e31b-4eaf-ac1e-186bb89c3476


https://github.com/user-attachments/assets/47a50bf9-5009-4ec6-a792-3dc7fac02c8c


It's mostly what you asked for, but the biggest change is for the trimmer

I tried moving it out of the way of the video (below or above) but I ran into this issue:
Whenever the view was moved outside the container view it would lose all interaction, mouse events would not respond at all. So what I did instead is add a drag bar to the top of the trimmer during crop editing. Now, if it's in your way, you can just drag it up and down. Then, when you close the crop editor it goes back to normal.

Gifski/EditScreen.swift
```diff
@@ -101,6 +154,90 @@ struct EditScreen: View {
...
+		var x = a
+		var y = b
+		var remainder = 0
+		while y != 0 {
+			remainder = x % y
+			x = y
+			y = remainder
+		}
+		return x
+	}
```
sindresorhus, line 239: This should be in Utilities.

Probably as this instead:

```swift
func greatestCommonDivisor<T: BinaryInteger>(_ a: T, _ b: T) -> T {
		let result = a % b
		return result == 0 ? b : greatestCommonDivisor(b, result)
	}
```

Gifski/EditScreen.swift
```diff
@@ -101,6 +154,90 @@ struct EditScreen: View {
...
+			set: { (newAspectRatio: AspectRatio) in
+				switch newAspectRatio {
+				case .divider:
+					return
+				case .custom:
+					let (width, height) = currentCropRectAspectRatio
+
+					enterCustomAspectRatioWidth = width
+					enterCustomAspectRatioHeight = height
+					showEnterCustomAspectRatio = true
```
sindresorhus, line 177: Use `SSApp.runOnce`

Gifski/EditScreen.swift
```diff
@@ -101,6 +154,90 @@ struct EditScreen: View {
 		}
 	}
 
+	var aspectRatioPicker: some View {
```
sindresorhus, line 157: Would probably be better to move this to a separate view, so it can keep some state local. Or maybe even a separate view for the `ToolbarItemGroup` contents.

Gifski/CropOverlayView.swift
```diff
@@ -0,0 +1,827 @@
...
+//  Gifski
+//
+//  Created by Michael Mulet on 3/23/25.
+
+import SwiftUI
+import AVFoundation
+import AVKit
+
+fileprivate let cornerWidthHeight = 20.0
+
```
sindresorhus, line 12: The logic in this file is not very maintainable. Would be nice to have it cleaned up, simplified. There is a lot of repetition that could be DRY'd up. Should probably happen when the logic is finalized though.

Gifski/AppState.swift
```diff
@@ -10,6 +10,17 @@ final class AppState {
 	var navigationPath = [Route]()
 	var isFileImporterPresented = false
 
+	var outputCrop = false
+	var onEditScreen: Bool {
```
sindresorhus, line 14:
```suggestion
	var isOnEditScreen: Bool {
```

Gifski/AppState.swift
```diff
@@ -10,6 +10,17 @@ final class AppState {
 	var navigationPath = [Route]()
 	var isFileImporterPresented = false
 
+	var outputCrop = false
```
sindresorhus, line 13:
```suggestion
	var isCropActive = false
```

sindresorhus (changes requested):

sindresorhus:
- The corner handles are close, but not perfect. Zoom into it and compare to the markup one in Quick Look.
- It's sometimes a bit hard to hit the corner handle to be able to drag it. I feel like the hit target is too small. For example, if the mouse pointer is exactly at the top-left of the corner, it won't drag.
- When setting a "Custom" aspect ratio, I think it would be better to add that ratio to a list (right above "Custom") with `Divider()` around it, if it's not already one of the defined aspect ratios. That way it would show in the dropdown when it's not open, which is more useful than just showing "Custom".
- For the custom aspect ratio popover, limit the inputs to `100` and resize the textfields to make that fit.
	- The left text field should have its text aligned to the right
		- You can do this by adding alignment setting to IntTextFieldCocoa. It can be set in `setup()`, like `alignment = .right`.
		- The text in the textfield should be monospaceddigit so they don't move when changing them.
- "Enter custom aspect ratio" text can be dropped.
- Maybe the "Reset" button should be in the aspect ratio picker instead (at the bottom)? To make the UI more minimal.
- As commented in the other pull request, the code requires cleanup; code style, simplify, etc.
- It's clever to let the user drag the trimmer components up and down, I feel like it's still in the way when cropping. I wonder if there's a better solution to this that we haven't yet thought of.

---

Screen Shot 2025-04-19 at 03 03 23

Gifski/EditScreen.swift
mmulet: If we use run Once, won't width and height be set to the defaults (except on the very first time the app is launched)? If I'm understanding you correctly, what you want is to set the width and height to the last used values that user entered, defaulting to the current width/height ratio at first? In that case, we should probably do something like this:

```swift

struct Aspect {
  let width = 1
  let height = 1
 }
@State private var customAspectRatio: Aspect?

// in the Binding(...,set:
  if customAspectRatio == nil {
    let (width, height) = currentCropRectAspectRatio
    customAspectRatio = .init(width: width, height: height)
  }
```

mmulet:
# Crop
- [x] CommandGroup Button to Toggle
- [x] CommandGroup Indentation
- [x] Move gcd to Utilities/greatestCommonDivisor
- [x] Use SSApp.runOnce instead of customAspect ratio, [see discussion](https://github.com/sindresorhus/Gifski/pull/331#discussion_r2062678450)
- [x] Move aspectRatioPicker to it's own view
- [x] Simplify CropOverLayView
  - I cleaned a lot of it up, and I rewrote all the handlers for each of the drag modes to make them much simpler and more readable.
- [x] onEditScreen -> isOnEditScreen
- [x] outputCrop -> isCropActive
- [x] Perfect crop corner handles* see note
- [x] Enlarge crop corner handle hit box
  - hit box is bigger now around the 
- [x] "Custom" aspect ratio, " I think it would be better to add that ratio to a list (right above "Custom") with Divider() around it, if it's not already one of the defined aspect ratios. That way it would show in the dropdown when it's not open, which is more useful than just showing "Custom".
- [x] For the custom aspect ratio popover, limit the inputs to 100 and resize the textfields to make that fit.   The left text field should have its text aligned to the right
You can do this by adding alignment setting to IntTextFieldCocoa. It can be set in setup(), like alignment = .right.
The text in the textfield should be monospaceddigit so they don't move when changing them.
- [x] Drop "Enter custom aspect ratio" text
- [x] Move Reset button to the aspect ratio picker
- [x] General cleanup
- [x] Brainstorm other ideas for getting the trimmer out of the way

I haven't had any particularly great ideas for getting the trimmer out of they way, but this is what I've got:
- On entering crop mode, shrink the video so that it won't get in the way (could animate this so it would look nice)
- Add a minimize button to the trimmer itself, or a way to drag most, but not all of it offscreen
- Find some way to recognize user intent, if they want to use the trimmer or adjust the crop, and move the trimmer out of the way intelligently
- ||**not actually a good idea**|| Turn the trimmer off every other frame in a flashing manner so you have a 50% chance to do what you actually want to every time.
- Allow the trimmer to be pushed the the bottom so that only the drag handle is showing, so it's mostly out of the way.

----
Other things I thought of that needed to be changed:
- [x] Show crop tooltip if you turn on crop from the menu bar

\* The Perfect crop handles:
I've been wracking my brain about this, because I swear I did make these handles pixel perfect. I had been taking screen shots back and forth between quick look and and Gifksi, and they were the same. Finally I came to a realization: I've been doing all my dev work on my external monitor, not my MacBook Pro screen! I moved both windows to the laptop screen and now the difference was night and day! So, I've put in some code to detect high dpi screens and it mimics the how each handle looks in quick look (on high dpi and low dpi)

sindresorhus:
Cropping works much better now. Nice work!

The crop handles are still not perfect though. It's the transparency that is the problem.

Markup left, Gifski right:

Screen Shot 2025-05-01 at 21 28 56

Screen Shot 2025-05-01 at 21 32 01

If you zoom in, you can see that the Gifski one has different opacity for the different parts.

Gifski/Crop/EditCropDragGesture.swift
```diff
@@ -0,0 +1,65 @@
...
+//  EditCropDragGesture.swift
+//  Gifski
+//
+//  Created by Michael Mulet on 4/27/25.
+//
+
+import Foundation
+import SwiftUI
+
+struct EditCropDragGesture: ViewModifier {
```
sindresorhus, line 11: ChatGPT simplified/improved it to this, which seems to work:

```swift
import SwiftUI

private struct CropDragGestureModifier: ViewModifier {
	@Binding var isDragging: Bool
	@Binding var cropRect: CropRect
	let frame: CGRect
	let position: CropHandlePosition
	let dragMode: CropRect.DragMode

	@GestureState private var initialCropRect: CropRect?

	func body(content: Content) -> some View {
		let dragGesture = DragGesture()
			.updating($initialCropRect) { _, state, _ in
				if state == nil {
					state = cropRect
				}
			}
			.onChanged { drag in
				guard let initial = initialCropRect else {
					return
				}

				isDragging = true
				cropRect = initial.applyDragToCropRect(
					drag: drag,
					frame: frame,
					position: position,
					dragMode: dragMode
				)
			}
			.onEnded { _ in
				isDragging = false
			}

		content.highPriorityGesture(dragGesture)
	}
}

extension View {
	func cropDragGesture(
		isDragging: Binding<Bool>,
		cropRect: Binding<CropRect>,
		frame: CGRect,
		position: CropHandlePosition,
		dragMode: CropRect.DragMode
	) -> some View {
		modifier(CropDragGestureModifier(
			isDragging: isDragging,
			cropRect: cropRect,
			frame: frame,
			position: position,
			dragMode: dragMode
		))
	}
}
```

Gifski/Crop/CropToolBarItems.swift
```diff
@@ -0,0 +1,224 @@
...
+				}.frame(width: 90)
+			}
+			.padding()
+			.frame(width: 90)
+		}
+	}
+	private var currentCropRectAspectRatio: PickerAspectRatio {
+		let width = Int(outputCropRect.width * metadata.dimensions.width)
+		let height = Int(outputCropRect.height * metadata.dimensions.height)
+
```
sindresorhus, line 152: Some of the code here could be simplified with this utility:


```swift
extension CGSize {
	/**
	Returns the simplest integer aspect ratio (width, height) for the current size.
	
	Usage:
		let (widthRatio, heightRatio) = size.integerAspectRatio()
	*/
	func integerAspectRatio() -> (Int, Int) {
		let roundedWidth = Int(width.rounded())
		let roundedHeight = Int(height.rounded())
		let divisor = greatestCommonDivisor(roundedWidth, roundedHeight)
		let widthRatio = roundedWidth / divisor
		let heightRatio = roundedHeight / divisor
		return (widthRatio, heightRatio)
	}
}
```

Gifski/Crop/CropToolBarItems.swift
```diff
@@ -0,0 +1,224 @@
...
+					outputCropRect = .initialCropRect
+				case .custom:
+					customAspectRatio = customAspectRatio ?? currentCropRectAspectRatio
+					showEnterCustomAspectRatio = true
+				case .preset(let preset):
+					outputCropRect = preset.cropRect(dimensions: metadata.dimensions)
+				}
+			}
+		)) {
+			ForEach(PickerAspectRatio.presets, id: \.self) { aspectRatio in
```
sindresorhus, line 90: Wrap it in `Section` to avoid needing `Divider`.

Gifski/Crop/CropToolBarItems.swift
```diff
@@ -0,0 +1,224 @@
...
+				Text(aspectRatio.label).tag(AspectRatio.preset(aspectRatio))
+			}
+			Divider().tag(AspectRatio.divider)
+			if let customAspectRatio,
+			   !customAspectRatio.isAPreset()
+			{
+				Text(customAspectRatio.label).tag(AspectRatio.preset(customAspectRatio))
+				Divider().tag(AspectRatio.divider)
```
sindresorhus, line 98: When I click "Reset", this one does not go away from the menu. It should.

Gifski/Crop/CropRect.swift
```diff
@@ -0,0 +1,382 @@
...
+		let height = height * scale
+		return CropRect(
+			x: midX - width / 2,
+			y: midY - height / 2,
+			width: width,
+			height: height
+		)
+	}
+}
+extension UnitPoint {
```
sindresorhus, line 363: This is weird, did you mean

```swift
var horizontal: Double { x }
var vertical: Double { y }
```

?

sindresorhus (changes requested):

sindresorhus:
- In the aspect ratio dropdown, when I click "Reset", it should not be selected. It's an action. It should select the aspect ratio it resets to.
- If I click "Reset", then click "Custom", but change nothing and dismiss the popover, then click aspect ratio again, "Custom" is selected, even though the ratio is "16:9" and it could have selected "16:9" instead.
- If the crop fills the whole video, and you move the mouse from outside the window (top-left of the app) to the top-left of the crop rect, the mouse does not turn into a drag indicator, even though it's draggable.
- If I move the mouse to exactly on the bottom-left handle and then start moving very slowly to the center of the crop rect, it jumps a lot. It also doesn't start moving right away, which is probably related.
- I recommend looking over the [diff of all changes](https://github.com/sindresorhus/Gifski/pull/331/files), which makes it very clear when there are some leftovers or inconsistent code style.

sindresorhus:
The App Intent also needs to support cropping. I think it could be useful to support both exact position and centered aspect ratio crop. Something like this: https://gist.github.com/sindresorhus/6d44ba44c605f7bf39dfd1caa05a03fa (AI generated, but should work) Ideas welcome if I missed anything.

sindresorhus:
> Allow the trimmer to be pushed the the bottom so that only the drag handle is showing, so it's mostly out of the way.

I think this is the best solution.

Gifski/EditScreen.swift
mmulet: That makes much more sense.

Gifski/Crop/CropRect.swift
mmulet: That just comes from using `UnitPoint`s to express sizes in the [0-1] coordinate system instead of frame.min to frame.max coordinate system. Maybe it would be better to just make a struct called `UnitSize`?

Gifski/TrimmingAVPlayer.swift
```diff
@@ -42,9 +46,86 @@ final class TrimmingAVPlayerViewController: NSViewController {
...
+				height: cropRectHolder.playerSize.height
+			)
+		}
+	}
+
+	private var underTrimCropOverlayView: NSHostingView<UnderTrimCropOverlay>!
 
 	var playerView: TrimmingAVPlayerView { view as! TrimmingAVPlayerView }
 
+	fileprivate var showCropRectUnderTrim = false {
```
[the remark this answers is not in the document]
mmulet: I already did something very similar in the gif preview branch*, so it should be easy to add it here.

\* I took them out of the most recent commit because it no longer touches `TrimmingAVPlayer`

mmulet:
## Batched changes
- [x] App.swift Toggle("Crop", isOn: appState.isCropActiveBinding)
- [x] CropRect.swift comment cleanup
- [x] CropToolBarItems.swift monospacedSystemFont -> monospacedDigitSystemFont
- [x] CropToolBarItems.swift 23.0 -> 26.0
- [x] CropToolBarItems.swift for to -> .contains
- [x] CropHandlePosition.swift location remove returns
- [x] CropHandlePosition.swift remove Equatable
- [x] CropHandlePosition.swift comment change

## Changes from this commit
- [x] SSApp.runOnce for showCropTooltip
- [x] DRY up CropToolBarItems
- [x] initial: false
- [x] Popover to separate view
- [x] View for the Fields in the popover
- [x] isAPreset -> matchesPreset
- [x] cropSideWhite -> = Self.white.opacity(0.75)
- [x] IntTextField nil, for custom alignment
- [x] IntTextField nil for custom font
- [x] delete divisor tag ( used Section instead)
- [x] Comment currentCropAspectRatio
- [x] add aspectRatio to PickerAspectRatio 
- [x] should Crop in DimensionSettings -> gone
- [x] Clean leftovers
- [x] remove returns from line-line-case-enums
- [x] rename is a control to move crop vertically
- [x] SSApp.swiftUIMainWindow -> bindHostingWindowgi
- [x] UnTask -> @MainActor
- [x] ChatGPTfy
 - GPT-4o approved
- [x] remove view from trim (used old code from gif preview branch)
- [x] Show GIF generator error to user
- [x] EditScreen leftovers
- [x] EditCropDragGesture simplify 
- [x] use IntegerAspectRatio
- [x] Wrap ToolbArItems in A section to avoid needing Divider
- [x] delete old custom on reset
- [x] move to `UnitSize` where applicable
- [x] If I click "Reset", then click "Custom", but change nothing and dismiss the popover, then click aspect ratio again, "Custom" is selected, even though the ratio is "16:9" and it could have selected "16:9" instead.
 - Moved to Menu instead of Picker.
- [x] shrink external hitbox for corner handles
- [x] If I drag the trimmer using the new drag-handle all the way to the bottom and more and then move the mouse up again until it reaches the trimmer, it starts moving when the mouse is inside the trimmer, not over the drag handle. It should only move when over the drag handle again.
- [x] I recommend looking over the diff of all changes, which makes it very clear when there are some leftovers or inconsistent code style.
- [x] Allow the trimmer to be pushed the the bottom so that only the drag handle is showing, so it's mostly out of the way.
- [x] Crop border opacity.
 - I was laser focused on the corners, that I forgot to look at the border. I made it look like Quick Look.
- [x] If I move the mouse to exactly on the bottom-left handle and then start moving very slowly to the center of the crop rect, it jumps a lot. It also doesn't start moving right away, which is probably related.
 - This was because the dragGesture didn't recognize the drag. I shrank the miniumDistance (10.0 by default) for drag, fixed it.
 - [x] The App Intent also needs to support cropping. I think it could be useful to support both exact position and centered aspect ratio crop
  - See notes below about the App Intent.

# TODO
- [ ] If the crop fills the whole video, and you move the mouse from outside the window (top-left of the app) to the top-left of the crop rect, the mouse does not turn into a drag indicator, even though it's draggable.
 - I haven't been able to reproduce this. It seems to work like expected to me, what am I doing wrong? See video

https://github.com/user-attachments/assets/477e9e6a-69bc-4898-a729-f36ec60cf2f9

- [ ] Move showCropTooltip out of the appState, see [discussion](https://github.com/sindresorhus/Gifski/pull/331#discussion_r2070618222)


## Nots:
I found something you might notice later on:
If you click outside of a corner of the crop rect, the handle will jump to the mouse position. This is the same behavior as Quick Look.

-----
The app intent the the Shortcuts app is kind of buggy, as in the progress doesn't always show up (as the TODO says in [Intent.swift](Gifski/Intent.swift), and sometimes the export will produce a blank (see screenshot). I tested it on the main branch and it still has the same issue, so it's not caused by my changes. I don't know if this is a bug in Gifski or in Shortcuts.

IntentFailureOnOriginMain

It does work with the crop intent for both aspect ratio and exact:
Screenshot 2025-05-02 at 1 41 03 PM
Screenshot 2025-05-02 at 2 17 11 PM

I also added the presets:

Screenshot 2025-05-02 at 2 18 16 PM

But, the Shortcuts app remains buggy. Sometimes (it really is stochastic, I haven't found a root cause) it will fail to recognize the crop at all and will repeatable prompt for one. If I keep pressing the run button over again it will fix it itself, that's really strange. 


-------

Finally, I left in the behavior of the Custom Text defaulting to the last values you typed in (unless you reset).

Gifski/Intents.swift
```diff
@@ -1,6 +1,247 @@
...
+			"mp4",
+			"mov"
+		],
+		resultValueName: "Crop"
+	)
+
+	static var parameterSummary: some ParameterSummary {
+		Switch(\.$mode) {
+			Case(.exact) {
+				Summary("Create a crop with exact dimensions") {
```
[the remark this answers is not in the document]
sindresorhus: Also maybe feed ChatGPT the whole file and ask it to "Clean up, and make sure it uses AppIntents best practices".

Gifski/Intents.swift
```diff
@@ -1,6 +1,247 @@
 import AppIntents
 import AVFoundation
 
+enum CropEntity: AppEntity, Codable {
```
sindresorhus, line 4: I have never seen an AppEntity be an enum before. I guess it works, but Apple doesn't use it in any examples, so maybe that could be the cause of the weird issues you mentioned.

Alternative:

```swift
struct CropEntity: AppEntity, Codable, Equatable {
	static let typeDisplayRepresentation = TypeDisplayRepresentation(name: "Crop")

	var id: String {
		get throws {
			try JSONEncoder().encode(self).base64EncodedString()
		}
	}

	var displayRepresentation: DisplayRepresentation {
		DisplayRepresentation(title: description)
	}

	var mode: CropIntentMode
	var x: Int?
	var y: Int?
	var width: Int
	var height: Int

	var description: String {
		switch mode {
		case .exact:
			"\(width)x\(height) at (\(x ?? 0),\(y ?? 0))"
		case .aspectRatio:
			"\(width):\(height)"
		}
	}

	func cropRect(forDimensions dimensions: (Int, Int)) throws -> CropRect? {
		let (videoWidth, videoHeight) = dimensions

		switch mode {
		case .aspectRatio:
			return CropRect.from(aspectWidth: Double(width), aspectHeight: Double(height), forDimensions: CGSize(width: videoWidth, height: videoHeight))
		case .exact:
			guard let x, let y else { return nil }
			let topY = videoHeight - y - height
			let rect = CGRect(x: x, y: topY, width: width, height: height)
			let videoRect = CGRect(origin: .zero, size: CGSize(width: videoWidth, height: videoHeight))
			let intersection = videoRect.intersection(rect)

			guard intersection.width >= 1, intersection.height >= 1 else {
				throw CropOutOfBounds.cropOutOfBounds(entityRect: rect, videoRect: videoRect)
			}

			return CropRect(
				x: Double(intersection.minX) / Double(videoWidth),
				y: Double(intersection.minY) / Double(videoHeight),
				width: Double(intersection.width) / Double(videoWidth),
				height: Double(intersection.height) / Double(videoHeight)
			)
		}
	}
}
```

Gifski/Intents.swift
```diff
@@ -1,6 +1,247 @@
...
+	case noCrop
+	/**
+	 The exact coordinates in the image using a *bottom-left* origin (the rest of the app uses top-left origin)
+	 */
+	case exact(x: Int, y: Int, width: Int, height: Int)
+	case aspectRatio(width: Int, height: Int)
+
+	var id: String {
+		do {
+			return try JSONEncoder().encode(self).base64EncodedString()
```
sindresorhus, line 14: Ensure the output is stable by using `.sortedKeys` for this.

Gifski/Intents.swift
```diff
@@ -1,6 +1,247 @@
...
 
+enum CropEntity: AppEntity, Codable {
+	case noCrop
+	/**
+	 The exact coordinates in the image using a *bottom-left* origin (the rest of the app uses top-left origin)
+	 */
+	case exact(x: Int, y: Int, width: Int, height: Int)
+	case aspectRatio(width: Int, height: Int)
+
+	var id: String {
```
sindresorhus, line 12: I think this type should maybe rather be a TransientAppEntity. I think you can use the `optionsProvider` in the `crop` `@Parameter` to give it a default list of aspect ratios.

Gifski/Intents.swift
```diff
@@ -1,6 +1,247 @@
...
+}
+
+enum CropIntentMode: String, AppEnum, CaseIterable {
+	static var caseDisplayRepresentations: [Self: DisplayRepresentation] = [
+		.exact: DisplayRepresentation(stringLiteral: "Exact"),
+		.aspectRatio: DisplayRepresentation(stringLiteral: "Aspect Ratio")
+	]
+
+	case exact
+	case aspectRatio
```
sindresorhus, line 92: This should be first everywhere and the default, since users are more likely to want aspect ratio crop than exact.

sindresorhus:
This is looking really good now. Only some minor things left:

- After setting a custom ratio, and then changing it by resizing the crop rectangle, the custom old ratio is still shown in the menu.
- I just realized that when we crop, we don't show the cropped exact size anywhere. Some users may need to know the exact final GIF size after cropping.
- When the trimmer control is dragged down to be not shown, the pause button focus border still shows through (you get it if you enable "Keyboard navigation" in System Settings and relaunch the app): Screen Shot 2025-05-05 at 18 34 06

sindresorhus (changes requested):

Gifski/Intents.swift
```diff
@@ -1,6 +1,247 @@
...
+
+	static var typeDisplayRepresentation = TypeDisplayRepresentation(name: "Crop Mode")
+}
+
+enum CropOutOfBounds: LocalizedError, CustomNSError {
```
sindresorhus, line 97: I was not able to trigger this error no matter what invalid values I put into the crop action. How do I trigger it?

sindresorhus: Also, should not be an enum when there is only one case. Then it should be a struct. If you anticipate more cases, it should have a more general name.

Gifski/Intents.swift
```diff
@@ -1,6 +1,247 @@
...
+		]
+	}
+}
+
+struct CreateCropIntent: AppIntent {
+	static var title: LocalizedStringResource = "Create Crop for Gifski"
+
+	static let description = IntentDescription(
+		"""
+		Crop your Gifski GIF output
```
sindresorhus, line 142:
```suggestion
		Create a crop to pass into the “Convert Video to Animated GIF” action.
```

Gifski/Intents.swift
```diff
@@ -86,6 +327,13 @@ struct ConvertIntent: AppIntent, ProgressReportingIntent {
 	)
 	var dimensionsHeight: Int?
 
+	@Parameter(
+		title: "Crop",
+		description: "Optionally crop the video",
+		default: .noCrop
+	)
+	var crop: CropEntity
```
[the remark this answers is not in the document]
mmulet: I tried that, but the bugs in Shortcuts with an optional `CropEntity` were worse than with `.noCrop`. For example, with `.noCrop` if shortcuts fails to connect the crop it will then prompt the user for a selection. With the optional crop, it will just skip the crop altogether and export the full video without a crop (which means you have to wait 30 seconds or more for it to just fail, with `.noCrop` it fails instantly).

But, if the suggestions below fix the bugs in Shortcut app, then that won't be a problem and we can make this optional.

Gifski/Intents.swift
mmulet: How to trigger: for a 1920x1080 video, use exact mode and extern a position greater than 1920x1080 like `x=2000,y=2000`. 

I will test this again to make sure it works.

mmulet:
- [x] Remove the Custom Text defaulting to the last values you typed in  (that I left in)
- [x] Show the cropped size in pixels somewhere
- [x] Fix pause button focus when dragged fully down
- [x] Move showCropTooltip out of the appState
- [x] cleanup Parameter summary
- [x] Make CropEntity Optional if it is free of bugs
- [x] structure the old trimmer offsets in TrimmerDragViews
- [x] `CropEntity` from enum -> struct
- [x] Entity id use `.sortedKeys`
- [ ] ❌ AppEntity -> TransientAppEntity
 - I could not get it TransientAppEntity to work in shortcuts, even buggier than when it was an enum.
- [ ] ❌ use `optionsProvider` in `@Parameter` instead of suggestions in the Query
 - Since it's not TransientAppEntity, it's forced to have a query, so no need for the `optionsProvider`
- [x] Default to `aspectRatio` instead of `exact`
- [x] Test and if necessary fix `CropOutOfBounds` Error
- [x] Move `CropOutOfBounds` error to a struct

## Other things I thought of:
- [x] Now that we show the video size, I added the ability to enter in a custom video size underneath the custom aspect ratio.
- [x] I changed how the aspect ratio selection changed
 
#### Here is how it used to be:

When changing the aspect ratio and the crop is not centered (you have dragged it or moved the handles), it defaults to a full screen aspect ratio. As in it re-centers it to the center of the video.

#### Here is how it is now:

When changing the aspect ratio and the crop is not centered (you have dragged it or moved the handles) it now keeps the center position of the crop the same and adjusts the crop the fit the aspect ratio.
Then, if you centered and the crop is full screen, it will revert to its old behavior.



## Notes

The intent is working reliably now. `AppEntity` on an enum was just cursed. As mentioned above, I did try to use `TransientAppEntity` with an `optionsProvider`, but I had absolutely no luck, it was buggy but in a different way. (The options would all have a checkmark next to them,
but absolutely none of them would work, it would be default, every time). Also passing in the Crop variable from the `CropIntent`, also didn't work, it was an all around failure.
Screenshot 2025-05-06 at 1 08 00 AM

sindresorhus:
> When changing the aspect ratio and the crop is not centered (you have dragged it or moved the handles) it now keeps the center position of the crop the same and adjusts the crop the fit the aspect ratio.
Then, if you centered and the crop is full screen, it will revert to its old behavior.

Much better 🙏

sindresorhus:
- When changing pixel size in the "aspect ratio" popover, the popover moves since the dropdown item in the toolbar changes width. This does not look very good. A trick I use for this is to use the size of the select box when it opened, so the popover never moves: https://gist.github.com/sindresorhus/3883cab306c2228b9133f308e40cf411

- I don't think the pixel size should be shown in the "Presets" list in the menu. The pixel sizes there don't mean anything. The user can resize anyway. I like how it's shown in the dropdown in the toolbar though.

- In the "custom" popover, if I enter `30x30` for the pixel size. It doesn't do anything. And i'm unable to use arrow-keys to change anything.

- In the "custom" popover, if I enter `100x100` for the pixel size, then I focus the second input, and press arrow-up, it changes to `1001` in the first input.

- In the "custom" popover, if I enter a custom pixel size, the aspect ratio does not reflect the current size, which may be confusing to the user. So when custom pixel size is active, maybe we should give less opacity to the aspect ratio shown in the "custom" popover, or some other way indicate that the shown aspect ratio is not the active one.

- The pixel size inputs are able to make the crop rect go outside of the bounds (notice right side of crop rect): 
Screen Shot 2025-05-06 at 15 15 18


- The "custom" popover is able to make the crop rect smaller than what's allowed with just manual resizing. I don't think it should:
Screen Shot 2025-05-06 at 15 23 42




---

- Intents: Maybe we should add a "is preview" parameter to the "convert" action which would generate just a single frame. So users could quickly debug their crop settings, etc. Just a frame from the middle of the video.

Gifski/Intents.swift
```diff
@@ -1,6 +1,245 @@
...
+	]
+	case aspectRatio
+	case exact
+	static var typeDisplayRepresentation = TypeDisplayRepresentation(name: "Crop Mode")
+}
+struct CropOutOfBounds: LocalizedError, CustomNSError {
+	let enteredRect: CGRect
+	let videoRect: CGRect
+	var errorDescription: String? {
+		"Crop rectangle is out of bounds! It's \(enteredRect.width)x\(enteredRect.height) at (\(enteredRect.x),\(enteredRect.y)), but it needs to fit inside \(videoRect.width)x\(videoRect.height) for this video."
```
sindresorhus, line 115: Convert the values to `Int` so they are displayed without `.0` since they are always whole numbers.

Gifski/Intents.swift
```diff
@@ -1,6 +1,245 @@
...
+enum CropIntentMode: String, AppEnum, CaseIterable, Codable, Hashable {
+	static var caseDisplayRepresentations: [Self: DisplayRepresentation] = [
+		.aspectRatio: DisplayRepresentation(title: "a fixed aspect ratio"),
+		.exact: DisplayRepresentation(title: "exact dimensions")
+	]
+	case aspectRatio
+	case exact
+	static var typeDisplayRepresentation = TypeDisplayRepresentation(name: "Crop Mode")
+}
+struct CropOutOfBounds: LocalizedError, CustomNSError {
```
sindresorhus, line 111: Instead of all this. I believe you can just conform to `CustomLocalizedStringResourceConvertible` to have any error be displayed in Shortcuts. No need for `CustomNSError`.

mmulet:
- [x] Use select box size trick, to keep popover from moving \*
  - see note
 - [x] Remove size from Menu, but keep in toolbar
 - [x] Custom pixel size field fix arrow keys, (when les than 30 pixels is too small)
 - [x] up arrow key moving to 1001 when switching fields with tab
 - [x] change opacity of aspect ratio when changing pixel field and vice versa
 - [x] fix pixel size input allowing out of bounds
 - [x] stop aspect ratio fields form letting the crop be too small on any axis
 - [x] Add is preview to output just a single frame
 - [x] For the CroppedOutOfBound error convert to Int
 - [x] Use `CustomLocalizedStringResourceConvertible` instead of `CustomNSError`

 ## Fixed 
 - [x] Added the dimensions parameter to crop drag rep because the frame size is not always the same as the video size on hi-dpi screens. (nor is the difference the same as `displayScale`)

## Note

\* The code provided doesn't really work  (it must be a change in SwiftUI because ChatGPT keeps telling me to do it your way), despite being in an overlay the popover is still anchored to the menu view. I colored the overlay white so you can see it, and as you can see the overlay is the correct size, but the popover is still moving.


https://github.com/user-attachments/assets/9bc2f9e3-d4e3-4dcb-8419-c588c4557d39

I created a workaround with plain ZStack, the only downside is that if you start go to say `1:1 - 100x100`

https://github.com/user-attachments/assets/eceed7e2-f30f-4201-abf0-3a7c8e1fcf2b

Gifski/Intents.swift
```diff
@@ -86,29 +316,44 @@ struct ConvertIntent: AppIntent, ProgressReportingIntent {
...
+		description: "Optionally crop the video.",
+	)
+	var crop: CropEntity?
+
+	@Parameter(
+		title: "Output Mode",
+		description: "Output an animated GIF or a single frame preview of the GIF.",
+		default: .animatedGIF
+	)
+	var outputMode: ConvertOutputMode
```
sindresorhus, line 330: Since it's just two cases, would be better to make it just a bool. Like `isPreview`.

Gifski/Intents.swift
```diff
@@ -86,29 +316,44 @@ struct ConvertIntent: AppIntent, ProgressReportingIntent {
...
+
 	// TODO: Dimensions setting. Percentage or width/height.
 
 	static var parameterSummary: some ParameterSummary {
 		Switch(\.$dimensionsType) {
 			Case(.pixels) {
-				Summary("Convert \(\.$video) to animated GIF") {
+				Summary("Convert \(\.$video) to \(\.$outputMode)") {
```
sindresorhus, line 337: It's not important enough to put in the summary. It can just be the last parameter after `$crop`.

sindresorhus:
If I write 10 in the custom aspect ratio popover for the size width field, it now shows a validation popover, but it would be even nicer if it just reset to the smallest size. Because right now, if a user accidentally writes 10, they are stuck on it if they try the arrow keys. So it could reset when either changing focus or using arrow keys.

sindresorhus:
If I change the height with arrow keys and it hits the bottom, the text field does not stop, but the size in the rect stops and in the dropdown. They should be consistent about the limit. Maybe a solution here is to move the rect up to make space so it's able to expand in height.

mmulet:
FYI: I'm going to be a bit busy for the next week or two, but I will get back to this issue afterwards.

sindresorhus:
🎉🎉🎉🎉🎉🎉🎉

mmulet:
> No worries at all. I really appreciate all your hard work on perfecting this. 🙏
> 
> I'm going to merge this now, but it would be great if you could address the remaining issues in another pull request when you are able to.

For sure. Thanks for your work too, you've been really clear, direct, and easy to work with.

PR #334, opened 2025-05-19
Crop Video part 2.

mmulet:
## Crop Video part 2.
- [x] The custom aspect ratio now lets me use the arrow-up key even though the actual rect does not change. It should probably stop when the crop rect cannot be changed anymore.
 - Now it automatically adjust the `IntTextField.minMax` range based on the current aspect ratio bounds.
- [x] Remove `ConvertOutputMode` because it is not used anymore
- [x] If I write 10 in the custom aspect ratio popover for the size width field, it now shows a validation popover, but it would be even nicer if it just reset to the smallest size. Because right now, if a user accidentally writes 10, they are stuck on it if they try the arrow keys. So it could reset when either changing focus or using arrow keys.
- [x] If I change the height with arrow keys and it hits the bottom, the text field does not stop, but the size in the rect stops and in the dropdown. They should be consistent about the limit. Maybe a solution here is to move the rect up to make space so it's able to expand in height.

### DIFF
--- a/Gifski/Crop/CropToolBarItems.swift
+++ b/Gifski/Crop/CropToolBarItems.swift
@@ -37,11 +37,6 @@ struct CropToolbarItems: View {
 	]
 }
 
-/**
-The range of valid numbers for the aspect ratio.
-*/
-private let aspectRatioNumberRange = 1...99
-
 private enum CustomFieldType {
 	case pixel
 	case aspect
@@ -143,7 +138,7 @@ private struct AspectRatioPicker: View {
 
 		customAspectRatio = PickerAspectRatio.closestAspectRatio(
 			for: cropSizeRightNow,
-			within: aspectRatioNumberRange
+			within: CropRect.defaultAspectRatioBounds
 		)
 
 		customPixelSize = cropSizeRightNow
@@ -189,14 +184,26 @@ private struct CustomAspectRatioView: View {
 				CustomAspectField(
 					customAspectRatio: $customAspectRatio,
 					modifiedCustomField: $modifiedCustomField,
-					side: \.width
+					side: \.width,
+					aspectRatioNumberRange: cropRect.aspectRatioBoundsForSide(
+						aspectWidth: Double(customAspectRatio?.width ?? 1),
+						aspectHeight: Double(customAspectRatio?.height ?? 1),
+						forDimensions: dimensions,
+						side: \.width
+					)
 				)
 				Text(":")
 					.foregroundStyle(.secondary)
 				CustomAspectField(
 					customAspectRatio: $customAspectRatio,
 					modifiedCustomField: $modifiedCustomField,
-					side: \.height
+					side: \.height,
+					aspectRatioNumberRange: cropRect.aspectRatioBoundsForSide(
+						aspectWidth: Double(customAspectRatio?.width ?? 1),
+						aspectHeight: Double(customAspectRatio?.height ?? 1),
+						forDimensions: dimensions,
+						side: \.height
+					)
 				)
 			}
 			.frame(width: 90)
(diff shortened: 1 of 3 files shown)


sindresorhus:
> So it could reset when either changing focus or using arrow keys.

It does not change when using arrow keys, only when switching focus to the other text input.

mmulet:
- [x] Much improved `CustomPixelField` by adding a `onInvalid` callback to `IntTextField`. Works for tab switching, arrow keys, and entering invalid values manually.

## Aspect Ratio Question

Validating the aspect ratio one side at a time is frustrating. For example:
(Math made up here, so the numbers are not accurate, but the idea is the same)
- say I want the aspect ratio 15 : 2, which for example let's say it's big enough to be valid.
- but my current aspect ratio is 34 : 1,
- let's say 15 : 1 is not valid, but 15 : 2 is.
- I go to enter 15 : 2, and I start with the width, and I enter 15 but the UI doesn't let me do that (it autocorrect s "to say 16")

So I can't enter 15 : 2 because the intermediate state of 15 : 1 is not valid. I think it would be better to validate the aspect ratio as a whole, and not one side at a time.



I think we could could do this by instead of having two text fields, we could have a single text field with a ratio in it (:). This way the user can enter the ratio as a whole, and not one side at a time.

Screenshot 2025-05-22 at 1 09 35 PM


That or we add an `apply` button below the text fields, and when the user presses it, we validate the aspect ratio as a whole. This way the user can enter whatever they want, and then press apply to see if it's valid or not.

Screenshot 2025-05-22 at 1 07 59 PM


What do you think?

Gifski/Crop/CropToolBarItems.swift
```diff
@@ -242,43 +242,48 @@ private struct CustomPixelField: View {
...
+		.onChange(of: warningCount){
+//			// no-op. Having the `warningCount` in the view hiearchy causes SwiftUI to refresh the `IntTextField` whenever an invalid value is entered even if we have already set the pixel size to `Self.minValue`. Can't use `.id()` modifier because it will close the popover.
+		}
 		.frame(width: 42.0)
 		.popover2(isPresented: $showWarning) {
 			VStack {
-				Text("Value is too small")
-				Text("\(value) < \(Self.minValue)")
+				Text("Value must be:")
+				Text("\(minMax.lowerBound) ≤ Value ≤ \(minMax.upperBound)")
```
sindresorhus, line 284: This is not very user-friendly. Can just say:

> `"Value must be in the range \(minMax.lowerBound) to \(minMax.upperBound)"`

sindresorhus:
Both proposed solutions feel like a UX downgrade. A single `:` field is less intuitive and harder to work with (no arrow keys to change value), and an Apply button adds unnecessary friction. The case where the aspect ratio is temporarily invalid is rare, and I don’t think the UI should be optimized around that edge case.

A better solution is to defer validation until editing ends (on blur), and maybe debounce validation during input. This allows users to freely enter values, even if intermediate states are invalid, without interruption.

sindresorhus (changes requested):

mmulet:
- [x] Popover message change
- [x] Allows users to freely enter values, even if intermediate states are invalid

> A better solution is to defer validation until editing ends (on blur), and maybe debounce validation during input. This allows users to freely enter values, even if intermediate states are invalid, without interruption.

I think this is a great solution! I implemented it such that:
- the `cropRect` itself is never in an invalid state (too small).
- But the text fields for the aspectRatio can be invalid temporarily. 
- After a debounce, the text fields switch to reflect the actual aspect ratio.

https://github.com/user-attachments/assets/4db1a6d2-2307-473f-8ef9-267178a41924

If you watch the video below you can see it in action:
1. (25:1) is invalid (20:1 is as high as the width can go)
2. But the user wants 25:5
3. They set the the width to 25
4. then they set the height to 5

At all times, the actual crop rect is valid, but the user can still edit the text fields.

## A few questions
- I picked the debounce time arbitrarily, is it too fast or too slow?
- It reduces the aspect ratio (dividing by least common multiple) after the debounce (`25:5` -> `5:1`). Is this too surprising, or is it fine?

sindresorhus:
> It reduces the aspect ratio (dividing by least common multiple) after the debounce (25:5 -> 5:1). Is this too surprising, or is it fine?

That's too surprising. We can do that after the user has closed the popover.

mmulet:
> > It reduces the aspect ratio (dividing by least common multiple) after the debounce (25:5 -> 5:1). Is this too surprising, or is it fine?
> 
> That's too surprising. We can do that after the user has closed the popover.

- [x] done.

sindresorhus:
Great 🙏

PR #339, opened 2025-06-27
implemented export modified video issue #337

### LINKED ISSUE #337: Add ability to save modified original video
Now that we have cropping, it could be useful to also be able to save the modified video (trimming and cropping), so Gifski could also be used to cropping video. I'm thinking a "Export as Video…" in the "File" menu.

mmulet:
Fixes #337 

- [x] save the modified video (trimming and cropping), so Gifski could also be used to cropping video. I'm thinking a "Export as Video…" in the "File" menu.
- [x] Export original audio

I implemented export of the original video, let me know what you think of the UI. Meanwhile, I will work on inserting the original audio track into the export.


mmulet:
- [x] Added sound export support (works with trimming and speed changes)
- [x] Fix: a bug where the video preview would flash black when focus changed between export windows.
- [x] Fix: Can have multiple export windows open at the same time now.
- [x] Fix: Now, it only deleted the temp video when the window closes, not on save (wasn't really a bug, I just don't know why I did it that way).

 Bug: I found a bug where if you change the speed of the video, `TrimmingAVPlayer.timeRangeDidChange` will be called with the full range of the video instead of the trimmed range. 
 So, if the time range is not full, and you change the speed, then the export time range will export the entire video instead of trimmed range. 
 I don't think this a bug from this pull request, it seems to be a bug in `TrimmingAVPlayer` itself. Is there a reason why it should be like this?

sindresorhus:
The video export is unlikely to take a long time, since most people use it for GIFs. So I think we can just show a sheet with a system spinner and cancel button if the video is longer than 20 seconds. And then afterwards, just immediately show the save dialog. One less click. And we don't really need the preview as the user can already preview in the main Gifski UI.

Gifski/App.swift
```diff
@@ -28,6 +28,12 @@ struct AppMain: App {
 				.keyboardShortcut("o")
 				.disabled(appState.isConverting)
 			}
+			CommandGroup(replacing: .importExport) {
+				Button("Export as Video…") {
+					appState.onExportAsVideo?()
+				}
+				.disabled(appState.onExportAsVideo == nil)
```
sindresorhus, line 36: Command+E shortcut

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,309 @@
...
+						progressContinuation.yield(progress.fractionCompleted)
+					default:
+						continue
+					}
+				}
+			}
+			defer {
+				progressTask.cancel()
+			}
+			try await exportSession.export(to: outputURL, as: .mov)
```
sindresorhus: Let's export as `.mp4`

Gifski/Utilities.swift
```diff
@@ -5420,8 +5430,8 @@ extension AVPlayerView {
 
 		Task {
 			/**
-			In about 20% of debug sessions, `beginTrimming` will crash because `canBeginTrimming` is false. We have seen multiple cases where this guard catches into the else statement and the trimming controls work just fine: in each and every case where `canBeginTrimming` was false, this function gets called again with a value of true.
-			*/
+			 In about 20% of my debug sessions, `beginTrimming` will crash because canBeginTrimming is false, so I added this check. I've seen multiple cases where this guard catches into the else statement and the trimming controls work just fine: in each and every case where canBeginTrimming was false, this function gets called again with a value of true.
+			 */
```
sindresorhus: Don't make unrelated changes.

Gifski/Utilities.swift
```diff
@@ -6560,3 +6578,62 @@ extension CompositePreviewFragmentUniforms: Equatable {
 		lhs.gridSize == rhs.gridSize
 	}
 }
+
+extension CGAffineTransform {
+	init(scaledBy size: CGSize) {
+		self = Self(scaleX: size.width, y: size.height)
+	}
+
+	func translatedBy(point: CGPoint) -> CGAffineTransform {
```
sindresorhus:
```suggestion
	func translated(by: CGPoint) -> Self {
```

Gifski/Utilities.swift
```diff
@@ -6560,3 +6578,62 @@ extension CompositePreviewFragmentUniforms: Equatable {
...
+		translatedBy(x: point.x, y: point.y)
+	}
+}
+
+
+
+extension Sequence {
```
sindresorhus: I would prefer using this:

```swift
extension Sequence where Element: Sendable {
	func concurrentMap<T: Sendable>(
		withPriority priority: TaskPriority? = nil,
		_ transform: @Sendable (Element) async -> T
	) async -> [T] {
		await withoutActuallyEscaping(transform) { escapingTransform in
			await withTaskGroup(of: (offset: Int, value: T).self) { group -> [T] in
				for (offset, element) in enumerated() {
					group.addTask(priority: priority) {
						await (offset, escapingTransform(element))
					}
				}

				var result = [(offset: Int, value: T)]()
				result.reserveCapacity(count)

				while let next = await group.next() {
					result.append(next)
				}

				return result
					.sorted { $0.offset < $1.offset }
					.map(\.value)
			}
		}
	}
}
```

Gifski/Utilities.swift
```diff
@@ -2793,6 +2797,12 @@ extension CMTimeRange {
 	}
 }
 
+extension ClosedRange where Bound == Double {
```
sindresorhus:
```suggestion
extension ClosedRange<Double> {
```

sindresorhus:
You also need to add MP4 to NSExportableTypes in Info.plist.

sindresorhus (changes requested):

mmulet:
- [x] The video export is unlikely to take a long time, since most people use it for GIFs. So I think we can just show a sheet with a system spinner and cancel button if the video is longer than 20 seconds. And then afterwards, just immediately show the save dialog. One less click. And we don't really need the preview as the user can already preview in the main Gifski UI.
- [x] Add MP4 to NSExportableTypes in Info.plist.
- [x] Command+E shortcut to export modified video.
- [x] .mov -> .mp4
- [x] restore canBeginTrimming comment in Utilities.swift
 - Sorry, this was a mistake, shouldn't have been in the commit.
- [x] `translatedBy(point:` -> `translated(by:`
- [x] `taskGroupMap` -> `concurrentMap`
- [x] `ClosedRange where Bound == Double` -> `ClosedRange<Double>`

Gifski/EditScreen.swift
```diff
@@ -34,14 +37,17 @@ struct EditScreen: View {
 
 private struct _EditScreen: View {
 	@Environment(AppState.self) private var appState
+	@Environment(\.openWindow) private var openWindow
```
sindresorhus: Unused

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,323 @@
...
+	@Environment(\.dismiss) private var dismiss
+	let input: Self.Input
+	@State private var state = ConvertState.generating(progress: 0)
+
+	var body: some View {
+		VStack(spacing: 20) {
+			switch state {
+			case let .empty(error: error):
+				if let error {
+					Text(error)
```
sindresorhus: Make sure you have visually checked this case. From experience, it could make the sheet huge if no max limits are set and the text is long.

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,323 @@
...
+				}
+				Button("Cancel") {
+					dismiss()
+				}
+			case .generating, .finished:
+				ProgressView()
+				if input.gifDuration >= 20 {
```
sindresorhus: I meant not show the sheet at all if a short video clip. But I don't think it's worth the complication, so just drop it.

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,323 @@
...
+					aspectRatio: aspectRatio
+				)
+			} catch {
+				if Task.isCancelled || error.isCancelled {
+					return
+				}
+				self.state = .empty(error: error.localizedDescription)
+			}
+		}
+		.fileExporter(
```
sindresorhus: This should not be attached to the sheet, because we should dismiss the sheet when showing the file exporter, so it's not momentarily visible after saving.

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,323 @@
...
+		switch state {
+		case let .finished(_, url, _, _):
+			url
+		default:
+			nil
+		}
+	}
+
+	private enum ConvertState: Equatable, Sendable {
+		case empty(error: String?)
```
sindresorhus: Should be called `idle` and error should be a separate case.

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,323 @@
...
+		case finished(filename: String, url: URL, player: AVPlayer, aspectRatio: Double)
+	}
+
+	private static func exportModifiedVideo(input: Input) -> ProgressableTask<Double, (URL, Double)> {
+		ProgressableTask { progressContinuation in
+			let exportComposition = try await ExportComposition(input: input)
+
+			let outputURL = FileManager.default.temporaryDirectory.appendingPathComponent( "\(UUID().uuidString).mp4")
+
+			try? FileManager.default.removeItem(at: outputURL)
```
sindresorhus: I don't understand this. The file is using a UUID, so it practically cannot exist already.

Gifski/VideoValidator.swift
```diff
@@ -147,13 +147,22 @@ enum VideoValidator {
 			)
 		}
 
+		let audioTracks = (try await asset.loadTracks(withMediaType: .audio))
+
+		let audioAssets = try await audioTracks.concurrentMap {
```
sindresorhus: A problem with doing it this way is that audio tracks don't always start at duration zero.

Gifski/EditScreen.swift
```diff
@@ -200,10 +221,19 @@ private struct _EditScreen: View {
 	private func setSpeed() async {
 		do {
 			// We could have set the `rate` of the player instead of modifying the asset, but it's just easier to modify the asset as then it matches what we want to generate. Otherwise, we would have to translate trimming ranges to the correct speed, etc.
+			let outputSpeed = Defaults[.outputSpeed]
+			async let changedSpeedAssetResult = try await asset.firstVideoTrack?.extractToNewAssetAndChangeSpeed(to: outputSpeed)
+			async let changedSpeedAudioAssetsResult = try await audioAssets.concurrentMap {
+				try await $0.firstAudioTrack?.extractToNewAssetAndChangeSpeed(to: outputSpeed)
+			}
```
sindresorhus: Why do this every time the speed is changed? Can just be done when the user actually wants to export the video.

Gifski/ExportModifiedVideo.swift
sindresorhus: Actually, would be better to dismiss the sheet and show the error using a `.alert(error:)`.

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,323 @@
...
+			url
+		default:
+			nil
+		}
+	}
+
+	private enum ConvertState: Equatable, Sendable {
+		case empty(error: String?)
+		case generating(progress: Double)
+		case finished(filename: String, url: URL, player: AVPlayer, aspectRatio: Double)
```
sindresorhus:
```suggestion
		case finished(filename: String, url: URL, aspectRatio: Double)
```

Make sure there are no leftovers.

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,323 @@
...
+						dismiss()
+					}
+				}
+			}
+		}
+		.padding()
+		.task(priority: .medium) {
+			do {
+				let task = Self.exportModifiedVideo(input: input)
+				self.state = .generating(progress: 0)
```
sindresorhus: It's safer to set state first.

```suggestion
				self.state = .generating(progress: 0)
				let task = Self.exportModifiedVideo(input: input)
```

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,323 @@
...
+			videoComposition.instructions = [instruction]
+			return videoComposition
+		}
+		enum Error: Swift.Error {
+			case noVideoTrack
+			case noAudioTrack
+			case unableToAddCompositionTrack
+
+
+			var errorDescription: String? {
```
sindresorhus:
```suggestion
		}

		enum Error: Swift.Error {
			case noVideoTrack
			case noAudioTrack
			case unableToAddCompositionTrack

			var errorDescription: String? {
```

example cleanup

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,323 @@
...
+					"The audio asset does not contain an audio track."
+				case .unableToAddCompositionTrack:
+					"Failed to add a composition track to the video."
+				}
+			}
+		}
+	}
+
+	enum Error: Swift.Error {
+		case unableToCreateExportSession
```
sindresorhus:
```suggestion
		case unableToCreateExportSession

```

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,323 @@
...
+				self.state = .empty(error: error.localizedDescription)
+			}
+		}
+		.fileExporter(
+			isPresented: .init(get: {
+				if case .finished = state {
+					return true
+				}
+				return false
+			}, set: { _ in }),
```
sindresorhus: The setter is empty, meaning when the file export sheet is dismissed (e.g., user cancels), no state is updated.

sindresorhus (changes requested):

sindresorhus:
The audio preservation is complicating this a lot, when it's not directly needed for Gifski. Maybe we should just drop it? We could show an alert once, the first time the user tries to export a video with audio tracks.

sindresorhus:
Also try to clean up and simplify the code more.

mmulet:
> The audio preservation is complicating this a lot, when it's not directly needed for Gifski. Maybe we should just drop it? We could show an alert once, the first time the user tries to export a video with audio tracks.

I agree. 



> Also try to clean up and simplify the code more.

Yeah, I'll take out the audio and simplify the PR a bunch.

mmulet:
- [x] Remove audio
- [x] Remove unused code
- [x] Simplify code
 - Simplified it quite a bit.
- [x] Show error with `Alert` if the export fails instead of in the sheet.
- [x] Check error for really long text
 - Moved to `appState.error`
- [x] Drop cancellation
- [x] Move the `fileExporter` modifier out of the sheet
- [x] Remove unnecessary `FileManager.removeItem`
- [x] The setter is empty, meaning when the file export sheet is dismissed (e.g., user cancels), no state is updated.

mmulet:
- [x] warning on export if audio track is present.

mmulet:
- [x] Fix: If another alert (like bounce warning) occurs when you activate `Save Modified Video`, the filexporter won't show and the state will be stuck on `.exported`. This fixes that by reassigning the state to force a swiftUI draw and bring up the file exporter.

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,208 @@
...
+					return
+				}
+				appState.onExportAsVideo?()
+			}
+		)
+	}
+
+	enum State {
+		case idle
+		case audioWarning
```
sindresorhus: This is an unrelated independent state and should not be here.

Gifski/GIFGenerator.swift
```diff
@@ -380,6 +380,92 @@ extension GIFGenerator.Conversion {
...
+				return trackSize
+			}
+			return dimensions / trackSize
+		}
+	}
+
+	/**
+	- Returns: Dimensions of the first video track
+	 */
+	var trackSize: CGSize? {
```
sindresorhus:
```suggestion
	var trackDimensions: CGSize? {
```

Gifski/GIFGenerator.swift
```diff
@@ -380,6 +380,92 @@ extension GIFGenerator.Conversion {
...
+			guard let dimensions = dimensionsAsSize else {
+				return trackSize
+			}
+			return dimensions / trackSize
+		}
+	}
+
+	/**
+	- Returns: Dimensions of the first video track
+	 */
```
sindresorhus:
```suggestion
	*/
```

same in many places.

Look over the diff for nitpick like this: https://github.com/sindresorhus/Gifski/pull/339/files

Gifski/Utilities.swift
```diff
@@ -1364,6 +1364,7 @@ extension AVAsset {
 		let duration: Duration
 		let frameRate: Double
 		let fileSize: Int
+		var originalVideoHasAudio: Bool
```
sindresorhus:
```suggestion
		var hasAudio: Bool
```

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,208 @@
...
+					try? url.setAppAsItemCreator()
+				} catch {
+					appState.error = error
+				}
+			}
+			.fileDialogCustomizationID("export")
+			.fileDialogMessage("Choose where to save the video")
+			.fileDialogConfirmationLabel("Save")
+			.alert2(
+				"Export video Limitation",
```
sindresorhus:
```suggestion
				"Export Video Limitation",
```

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,208 @@
...
+				switch state {
+				case .audioWarning:
+					true
+				default:
+					false
+				}
+			},
+			set: {
+				guard !$0,
+					  case .audioWarning = state else{
```
sindresorhus:
```suggestion
					  case .audioWarning = state else {
```

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,208 @@
...
+			}
+		)
+	}
+
+	enum State {
+		case idle
+		case audioWarning
+		case exporting(Task<Void, Never>)
+		case exported(URL)
+
```
sindresorhus: Add a `isExporting` and `isExported` getter and use it in isProgressSheetPresented and isFileExporterPresented

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,208 @@
...
+	conversion: GIFGenerator.Conversion
+) async throws -> AVMutableVideoComposition {
+	let videoComposition = AVMutableVideoComposition()
+
+	let cropRectInPixels = try await conversion.cropRectInPixels
+	videoComposition.renderSize = cropRectInPixels.size
+	videoComposition.frameDuration = try await compositionVideoTrack.load(.minFrameDuration)
+
+	let instruction = AVMutableVideoCompositionInstruction()
+	// The instruction time range must be greater than or equal to the video and there is no penality for making it longer, so add 1.0 second to the duration just to be safe
```
sindresorhus:
```suggestion
	// The instruction time range must be greater than or equal to the video and there is no penality for making it longer, so add 1.0 second to the duration just to be safe
```

sindresorhus: Use AI to check for typos, bugs, and potential improvements for all the code in this pull request.

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,208 @@
...
+		try await conversion.exportModifiedVideoTimeRange,
+		of: try await conversion.firstVideoTrack,
+		at: .zero
+	)
+	return (composition, compositionTrack)
+}
+
+/**
+Create an  `AVMutableVideoComposition` that wll scale, translate, and crop the `compositionVideoTrack`
+ */
```
sindresorhus:
```suggestion
Create an  `AVMutableVideoComposition` that will scale, translate, and crop the `compositionVideoTrack`.
*/
```

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,208 @@
...
+				appState.onExportAsVideo?()
+			}
+		)
+	}
+
+	enum State {
+		case idle
+		case audioWarning
+		case exporting(Task<Void, Never>)
+		case exported(URL)
```
sindresorhus:
```suggestion
		case finished(URL)
```

Gifski/GIFGenerator.swift
```diff
@@ -380,6 +380,92 @@ extension GIFGenerator.Conversion {
...
+	var trackSize: CGSize? {
+		get async throws {
+			try await asset.firstVideoTrack?.dimensions
+		}
+	}
+
+	var dimensionsAsSize: CGSize? {
```
sindresorhus:
```suggestion
	var dimensionsAsCGSize: CGSize? {
```

sindresorhus (changes requested):

mmulet:
- [x] `ExportModifiedVideoView.State` -> `ExportModifiedVideoState`
- [x] `GifGenerator.trackSize` -> `GifGenerator.trackDimensions`
- [x] Style fixes like `\s*/` -> `*/`, `else {` -> `else \n {`
- [x] `AVAsset.VideoMetadata.originalVideoHasAudio` -> `AVAsset.VideoMetadata.hasAudio`
- [x] `Export video Limitation` -> `Export Video Limitation`
- [x] `ExportModifiedVideoState.isExporting` getters for isProgressSheetPresented, etc
- [x] Use AI to check for typos, bugs, and potential improvements for all the code in this pull request.
- [x] `exported(..)` -> `finished(...)`
- [x] `GifGenerator.dimensionsAsSize` -> `GifGenerator.dimensionsAsCGSize`

sindresorhus:
https://github.com/sindresorhus/Gifski/pull/339#discussion_r2217805519 is not done.

mmulet:
My bad! I must have missed somethings. I will look at it tomorrow.

mmulet:
- [x] Now it only shows the progress when the video is over 20 seconds long.

What else am I missing? It seems to work to me.

sindresorhus:
That's not what the link points to. It's:

> This is an unrelated independent state and should not be here.

mmulet:
> That's not what the link points to. It's:
> 
> > This is an unrelated independent state and should not be here.

Ah, I see now. I went and moved the ExportVideoState to its own file.

sindresorhus:
No, that's not it. The comment is on the `case audioWarning` line. And I meant that that case is not related to the state and should be a independent boolean state in the view itself, not part of this enum.

mmulet:
> No, that's not it. The comment is on the `case audioWarning` line. And I meant that that case is not related to the state and should be a independent boolean state in the view itself, not part of this enum.

Oh, that makes a lot more sense. Sorry about the confusion. I will fix it now.

mmulet:
- [x] Move audioWarning out of ExportModifiedVideoState

Gifski/ExportModifiedVideoState.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+
```
sindresorhus: Leftover

Gifski/GIFGenerator.swift
```diff
@@ -366,20 +366,117 @@ extension GIFGenerator.Conversion {
...
+	var scale: CGSize {
+		get async throws {
+			guard let trackDimensions = try await trackDimensions else {
+				return .one
+			}
+			guard trackDimensions > 0 else {
+				throw Error.invalidDimensions
+			}
+			guard let dimensions = dimensionsAsCGSize else {
+				return trackDimensions
```
sindresorhus: Incorrect return value.

Gifski/EditScreen.swift
```diff
@@ -175,6 +196,44 @@ private struct _EditScreen: View {
...
+					if Task.isCancelled || error.isCancelled {
+						return
+					}
+					await MainActor.run {
+						exportModifiedVideoState = .idle
+						appState.error = error
+					}
+				}
+			},
+			videoIsOverTwentySeconds: conversionSettings.gifDuration(assetTimeRange: modifiedAssetTimeRange).toTimeInterval > 20
```
sindresorhus:
```suggestion
			videoIsOverTwentySeconds: conversionSettings.gifDuration(assetTimeRange: modifiedAssetTimeRange) > .seconds(20)
```

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,232 @@
...
+				}
+				try? FileManager.default.removeItem(at: url)
+				state = .idle
+			}
+		)
+	}
+
+
+	enum Error: Swift.Error {
+		case unableToCreateExportAtHighestQuality
```
sindresorhus: This case is never thrown.

Gifski/ExportModifiedVideo.swift
```diff
@@ -0,0 +1,232 @@
...
+
+/**
+Create an `AVMutableVideoComposition` that will scale, translate, and crop the `compositionVideoTrack`.
+*/
+private func createVideoComposition(
+	compositionVideoTrack: AVMutableCompositionTrack,
+	conversion: GIFGenerator.Conversion
+) async throws -> AVMutableVideoComposition {
+	let videoComposition = AVMutableVideoComposition()
+
```
sindresorhus, line 214: You may have to use `preferredTransform` here to correctly handle videos with metadata that they are rotated.

sindresorhus:
If I trim a video, and then change the speed, the saved video does not seem to respect the trim.

The speed also seems to be additive, so if I drag it to faster multiple times, it just gets faster.

sindresorhus (changes requested):

mmulet:
- [x] Fix: If I trim a video, and then change the speed, the saved video does not seem to respect the trim.
 - This was a bug in `AVTrimmingPlayerView`, not from this pr, but I fixed it anyways.
- [x] Remove leftovers exportModifiedVideoState.swift file
- [x] GifGenerator return value fix
- [x] .toTimeInterval > 20 -> _ > .secondons(20)
- [x] remove leftover case
- [x] Use preferredTransform in `ExportModifiedVideo` to correctly handle videos with metadata that they are rotated.

## Support for preferredTransform
preferredTransform turned out to be a can of worms, because almost nothing else worked with it! I had to make changes to preview, crop, intents, and normal gif export to support it.

## Couldn't reproduce
- [ ] Fix: The speed also seems to be additive, so if I drag it to faster multiple times, it just gets faster.
 - I could not reproduce this bug. I've attached a video of me trying to change the speed a lot and each time it exports correctly?

mmulet:
exports just fine:

https://github.com/user-attachments/assets/1cc11df1-441c-4387-87c7-46436c1de993

and I don't see the problem when scrubbing the speed while playing (although it does look like sometimes it does not change the play speed, which is a separate issue that I should look into*):



https://github.com/user-attachments/assets/b9f2bb26-c9ef-4927-bab2-f39a7c014798

\* I did look into it a bit. It's hard to diagnose because 

1. It doesn't always happen. It works most of the time, and I'm still having trouble triggering it reliably.
2. When it does happen, neither `setSpeed` or or the on receive get an update. 
```swift
.onReceive(Defaults.publisher(.outputSpeed, options: []).removeDuplicates().debounce(for: .seconds(0.4), scheduler: DispatchQueue.main)) { _ in
			Task {
				await setSpeed()
			}
		}
```

I'm out of time today to figure out what the problem is.

sindresorhus:
> Fix: The speed also seems to be additive, so if I drag it to faster multiple times, it just gets faster.

One way to reproduce each time is to set speed to 4x, then relaunch Gifski and then drag to 1x. It then doesn't play as 1x.

However, in the latest commits, I noticed the speed slider doesn't have any effect at all.

Gifski/Components/TrimmingAVPlayer.swift
```diff
@@ -284,6 +283,31 @@ final class TrimmingAVPlayerViewController: NSViewController {
...
+	func onNewDurationRange(durationRange newItemDurationRange: ClosedRange<Double>?) {
+		guard let newItemDurationRange else {
+			currentItemDurationRange = nil
+			return
+		}
+		defer {
+			currentItemDurationRange = newItemDurationRange
+		}
+		guard let timeRange,
+			  let currentItemDurationRange else {
```
sindresorhus: Code style. Applies everywhere.
```suggestion
		guard
			let timeRange,
			let currentItemDurationRange
		else {
```

Gifski/Components/TrimmingAVPlayer.swift
```diff
@@ -284,6 +283,31 @@ final class TrimmingAVPlayerViewController: NSViewController {
...
+			timeRangeDidChange?(newItemDurationRange)
+			return
+		}
+		// convert timeRange from oldItemDurationRange to new duration range
```
sindresorhus:
```suggestion
		// Convert `timeRange` from `oldItemDurationRange` to new duration range
```

sindresorhus:
* Crop API typos.
  `unormalziedCropFor` → `unnormalizedCropRect`, `prefferedSize` → `preferredSize`, `origninalSize` → `originalSize`.

* Wrong math for crop transform.
  `CGRect(origin:.zero,size:preferredSize).applying(trackPreferredTransform.inverted()).size` to derive “original size” is bogus for 90° rotations (you get the bounding box, not the pre-rotated size). Don’t transform a size. Instead, unnormalize in **original** pixel space, then apply the preferred transform to the **rect**:

  ```swift
  let norm = crop ?? .initialCropRect
  let base = norm.unnormalize(forDimensions: preferredSize) // original pixel space
  let rectInPreferred = trackPreferredTransform.map { base.applying($0) } ?? base
  return rectInPreferred
  ```

  Replace your `unormalziedCropFor(...)` body with the above logic.

* `croppedImage` guard + logic mismatch.
  You early-return when `crop == nil`, but `unormalziedCropFor` internally falls back to `.initialCropRect` anyway. Pick one rule: either (a) return the image when `crop == nil` **and** never use `.initialCropRect` internally, or (b) always crop using `.initialCropRect` when `crop == nil`. Right now it’s inconsistent. I’d do (a).

* Time-range handoff bug on speed change.
  In `onNewDurationRange`, when there’s no previous `timeRange/currentItemDurationRange`, you only call `timeRangeDidChange?`, but you don’t update `self.timeRange`. Set it so downstream code sees the correct range:

  ```swift
  guard let timeRange, let currentItemDurationRange else {
  	self.timeRange = newItemDurationRange
  	timeRangeDidChange?(newItemDurationRange)
  	return
  }
  ```

* You gate the spinner with `gifDuration(...)`, which can double with bounce, while MP4 export ignores bounce. Base it on the **non-bounce** segment length (the modified asset/timeRange), not the GIF duration.

* **Export session nit:**
  Set `exportSession.shouldOptimizeForNetworkUse = true` after creating the session. It’s harmless and improves quick-start playback.

---

And as always, try to find a way to simplify. There are a lot of changes here.

mmulet:
I will take a look at these suggestions and make the necessary changes this weekend.

Gifski/Components/TrimmingAVPlayer.swift
mmulet: Sorry about this one. I was a bit confused because the alignment did look good on both, xcode and vscode (see below). But, I turned on whitespace and it looks like it was a interplay between tabs and spaces that the Github file viewer doesn't like. I'll switch everything over to pure tabs.

Screenshot 2025-10-04 at 12 26 51 PM
Screenshot 2025-10-04 at 12 29 07 PM

mmulet:
## Done
- [x] convert time range... -> Convert `time range`...
- [x] Crop api typos: unormalziedCropFor → unnormalizedCropRect, prefferedSize → preferredSize, origninalSize → originalSize.
- [ ] Wrong math for crop transform.
  - WontFix: This fix is "bogus", 1. it doesn't work. 2. The code I wrote does work.
- [x] croppedImage guard + logic mismatch
- [x] Time-range handoff bug on speed change
- [x] Decide whether or not to show spinner based on video length without bounce
- [x] Add exportSession.shouldOptimizeForNetworkUse = true

mmulet:
- [x] Add icon to Export modified video menu item
- [x] Style fixes 
- [x] Simplify code
- [x] Fix speed slider
 - The problem was with the combine code. So instead of using combine, I used the debouncer from utilities.swift and implemented deduping manually.
- [ ] Reproduce `The speed also seems to be additive, so if I drag it to faster multiple times, it just gets faster.`
 - After fixing the speed slider, I did as you said: set the speed to 4x, then quit Gifski, then re-open it and set the speed to 1x. It worked correctly, so I couldn't reproduce the bug. I attached a video, and as you can see, it's all working fine.

https://github.com/user-attachments/assets/6fd2b751-563c-4b29-a6e3-e3b71f0d7bfd

sindresorhus:
Thank you and sorry about the huge delay. Happy new year!

mmulet:
No worries. Happy new year to you too!

Issue #95, opened 2019-06-24
After newest libgifski update the generator needs few seconds to start and the progress is laggy


sunshinejr:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2480%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/gifski-app/issues/95)

# The bounty here now applies to https://github.com/sindresorhus/Gifski/issues/180 instead.

---

This is the video I reproduced it with.

[new_gifski_transitions.mov.zip](https://github.com/sindresorhus/Gifski/files/3321317/new_gifski_transitions.mov.zip)

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $80.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($80.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/gifski-app/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

sindresorhus:
I've made a branch with logging: https://github.com/sindresorhus/Gifski/tree/logging-for-95 I've wrapped the `generateCGImagesAsynchronously` callback in a DispatchQueue, so libgifski won't block it.

Here's the log output:

```
generate 0 2019-06-24 14:51:53 +0000
generate 1 2019-06-24 14:51:53 +0000
generate 2 2019-06-24 14:51:53 +0000
generate 3 2019-06-24 14:51:54 +0000
generate 4 2019-06-24 14:51:54 +0000
generate 5 2019-06-24 14:51:54 +0000
generate 6 2019-06-24 14:51:54 +0000
generate 7 2019-06-24 14:51:54 +0000
generate 8 2019-06-24 14:51:54 +0000
generate 9 2019-06-24 14:51:54 +0000
generate 10 2019-06-24 14:51:54 +0000
generate 11 2019-06-24 14:51:54 +0000
generate 12 2019-06-24 14:51:54 +0000
generate 13 2019-06-24 14:51:54 +0000
generate 14 2019-06-24 14:51:55 +0000
[...]

PR #340, opened 2025-06-27
Quick action tip

mmulet:
Added the quick action tip along with a guide on how to create the shortcut.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#95: After newest libgifski update the generator needs few seconds to start and the progress is laggy](https://oss.issuehunt.io/repos/119822304/issues/95)
---
</details>


sindresorhus:
I meant just a short tip. We don't really need to provide this much handholding and I also don't want to maintain this. The exact UI may change in later macOS versions. Instead, just a short paragraph tip and we can provide an example shortcut.

mmulet:
> I meant just a short tip. We don't really need to provide this much handholding and I also don't want to maintain this. The exact UI may change in later macOS versions. Instead, just a short paragraph tip and we can provide an example shortcut.

Sure, updated with just a short paragraph and just the example screenshot. Or if you'd rather, instead of the screenshot, I can just include a link to the exported .shortcut

PR #345, opened 2026-02-04
Show warning when frame count is too low

### LINKED ISSUE #212: Show warning if the number of frames are too low
Gifski requires at least 2 frames in the resulting GIF, but users might accidentally trim or choose low enough FPS for the resulting GIF to contain less than 2 frames. We should show a warning in the editor's view in that case.

Right now, we just show an error when the conversion starts:

Screen Shot 2020-10-24 at 16 53 31

william-laverty:
## Summary
Adds a warning in the editor view when the current trim and FPS settings would result in fewer than 2 frames, which is the minimum required for a valid GIF.

## Changes
- Added `estimatedFrameCount` computed property to calculate frames based on current duration and FPS
- Added `hasLowFrameCount` property to check if frame count < 2
- Added `lowFrameCountWarning` view that appears in the bottom bar
- The warning replaces the file size estimate when active (to avoid clutter)

## Screenshots
The warning appears with an orange triangle icon and displays the current frame count, helping users understand why their settings won't work before attempting to convert.

## Testing
1. Open a video in Gifski
2. Either:
   - Trim the video to a very short duration, OR
   - Set FPS very low
3. When frame count drops below 2, the warning should appear
4. When frame count is 2 or more, the normal file size estimate should show

Closes #212


sindresorhus:
Thanks, but I'm not accepting fully AI-generated contributions.

PR #346, opened 2026-02-04
Show warning when frame count would be too low

### LINKED ISSUE #212: Show warning if the number of frames are too low

william-laverty:
## Summary

Adds a proactive warning in the editor view when current settings (trim range and FPS) would result in fewer than 2 frames in the output GIF.

## Problem

Currently, users only see an error after attempting to convert:

error

This is frustrating because:
1. Users don't know what settings to adjust
2. They've already waited for the conversion to start

## Solution

Show a warning banner in the editor that appears immediately when settings would result in < 2 frames:

### Changes

1. **Added `estimatedFrameCount` computed property** - Calculates expected frames based on `timeRange.length * frameRate`

2. **Added `hasLowFrameCount` computed property** - Returns `true` when `estimatedFrameCount < 2`

3. **Added warning banner** - Yellow warning banner appears in the bottom bar showing:
   - Warning icon
   - Message explaining the issue and suggesting fixes (increase FPS or extend trim range)
   - Dismiss button (warning reappears when settings change)

4. **Disabled Convert button** - When frame count is too low, the Convert button is disabled to prevent users from starting a doomed conversion

5. **Auto-reset on settings change** - Warning dismiss state resets when FPS or trim range changes, so users see updated feedback

## Implementation Notes

- Uses existing SwiftUI patterns from the codebase
- Warning matches the app's visual style
- Minimal code addition (~50 lines)
- No external dependencies

Closes #212


sindresorhus:
Thanks, but I'm not accepting fully AI-generated contributions.

PR #348, opened 2026-05-07
Use crop size for export dimensions

VoxelAgentSimon:
## Summary

The export dimensions controls now use the cropped video area as their reference size.

Previously, the width and height fields were based on the full source video dimensions even after cropping. That made it hard to understand the final exported size, since users had to mentally calculate the cropped output dimensions themselves.

## Changes

- Use the active crop rectangle to determine the maximum export dimensions.
- Preserve the current export scale when the crop size changes.
- Update pixel and percent presets to be based on the cropped area size.
- Add tests for preserving the selected scale when the reference dimensions change.

## Testing

- Built and ran the app locally.
- Verified that after cropping, 100% export size matches the cropped area size instead of the original full video size.
- Verified that existing width/height and percent controls still update correctly.


sindresorhus:
Thanks for contributing, but this is fixed as a larger refactor in https://github.com/sindresorhus/Gifski/commit/7f873856e2acd8b52e6681dee3aec31e6cab23e4

PR #349, opened 2026-05-08
Update "Estimated Size" when cropping

VoxelAgentSimon:
### Summary

Fixes the estimated file size not updating when changing the screen-area crop.

### Changes

Added an outputCropRect change observer in EditScreen.
Also refreshes preview settings after crop changes, matching the other settings observers.

### Testing

Built a local Debug app and confirmed working.


sindresorhus:
Thanks for contributing, but this is fixed as a larger refactor in https://github.com/sindresorhus/Gifski/commit/7f873856e2acd8b52e6681dee3aec31e6cab23e4

